PR #79, opened 2022-03-02
feat: support fetchMostRecentBeforeStart time series data request setting

boweihan:
## Overview
`fetchMostRecentBeforeStart` is needed for the line chart to display leading lines and for the status timeline to display pre-viewport status. This PR adds support for this time series data request setting.

Example of leading line:
Screen Shot 2022-03-03 at 1 26 22 PM

The current cache behaviour needs to be adjusted so that cache intervals for `fetchMostRecentBeforeStart` requests don't start at `Date(0,0,0)` _unless_ there was no data returned. Core cache logic was added for this case to start at the data point returned.

## Legal

### DIFF
--- a/packages/core/src/__mocks__/data-source.ts
+++ b/packages/core/src/__mocks__/data-source.ts
@@ -3,10 +3,9 @@ import {
   DataSourceRequest,
   DataStream,
   RequestInformationAndRange,
-  SiteWiseDataStreamQuery
-} from "../data-module/types";
-import { toDataStreamId } from "../common/dataStreamId";
-import { toId } from "@iot-app-kit/source-iotsitewise";
+  SiteWiseDataStreamQuery,
+} from '../data-module/types';
+import { toDataStreamId } from '../common/dataStreamId';
 
 // A simple mock data source, which will always immediately return a successful response of your choosing.
 export const createMockSiteWiseDataSource = (
@@ -19,24 +18,34 @@ export const createMockSiteWiseDataSource = (
   } = { dataStreams: [], onRequestData: () => {} }
 ): DataSource<SiteWiseDataStreamQuery> => ({
   name: 'site-wise',
-  initiateRequest: jest.fn(({ query, request, onSuccess = () => {} }: DataSourceRequest<SiteWiseDataStreamQuery>, requestInformations: RequestInformationAndRange[]) => {
-    query.assets.forEach(({ assetId, properties }) =>
-      properties.forEach(({ propertyId }) => {
-        const correspondingRequestInfo = requestInformations.find(({ id }) => `${ assetId }---${propertyId}` === id);
-        if (correspondingRequestInfo) {
-          onRequestData({ assetId, propertyId, request });
-          onSuccess(dataStreams, 'fetchFromStartToEnd', correspondingRequestInfo.start, correspondingRequestInfo.end);
-        }
-      })
-    );
-  }),
+  initiateRequest: jest.fn(
+    (
+      { query, request, onSuccess = () => {} }: DataSourceRequest<SiteWiseDataStreamQuery>,
+      requestInformations: RequestInformationAndRange[]
+    ) => {
+      query.assets.forEach(({ assetId, properties }) =>
+        properties.forEach(({ propertyId }) => {
+          const correspondingRequestInfo = requestInformations.find(({ id }) => `${assetId}---${propertyId}` === id);
+          if (correspondingRequestInfo) {
+            onRequestData({ assetId, propertyId, request });
+            onSuccess(
+              dataStreams,
+              correspondingRequestInfo,
+              correspondingRequestInfo.start,
+              correspondingRequestInfo.end
+            );
+          }
+        })
+      );
+    }
+  ),
   getRequestsFromQuery: ({ query }) =>
     query.assets
       .map(({ assetId, properties }) =>
         properties.map(({ propertyId, refId }) => ({
           id: toDataStreamId({ assetId, propertyId }),
           refId,
-          resolution: '0'
+          resolution: '0',
         }))
       )
       .flat(),
--- a/packages/core/src/data-module/IotAppKitDataModule.ts
+++ b/packages/core/src/data-module/IotAppKitDataModule.ts
@@ -16,7 +16,7 @@ import DataSourceStore from './data-source-store/dataSourceStore';
 import { DataCache } from './data-cache/dataCacheWrapped';
 import { TimeSeriesDataRequest } from './data-cache/requestTypes';
 import { requestRange } from './data-cache/requestRange';
-import { getDateRangesToRequest } from './data-cache/caching/caching';
+import { getRequestInformations } from './data-cache/caching/caching';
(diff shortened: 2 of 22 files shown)


diehbria:
we'll also want to provide this option as default to on for the components where it makes sense, if they aren't all ready.

Are we 100% confident all the caching behavior works as appropriately for this scenario?

packages/core/src/data-module/IotAppKitDataModule.ts
```diff
@@ -88,22 +88,39 @@ export class IotAppKitDataModule implements DataModule {
...
       })
       .flatMap(({ dateRanges, request }) =>
-        dateRanges.map(([rangeStart, rangeEnd]) => ({ start: rangeStart, end: rangeEnd, ...request }))
+        dateRanges.map(([rangeStart, rangeEnd]) => ({
+          start: rangeStart,
+          end: rangeEnd,
+
+          // determine if the most recent point before start is already cached
+          // to avoid redundant fetchMostRecentBeforeStart
+          leadingPointInCache: checkCacheForRecentPoint({
```
diehbria, line 114: I'm not convinced we need to introduce this new concept and carry it through out. We already have a function that returns the date ranges the request settings to request, what if we instead moved all the caching logic in one place, `getDateRangesToRequest`?

diehbria (changes requested):

packages/core/src/data-module/IotAppKitDataModule.ts
boweihan: The trouble with fetching leading points is that for a potentially large date range, we only want to fetch a single point (whereas currently, all data within these date ranges are fetched). Request settings don't have the ability to differentiate between different _types_ of requests so this new request setting introduces that capability.

packages/core/src/__mocks__/data-source.ts
```diff
@@ -19,24 +18,34 @@ export const createMockSiteWiseDataSource = (
...
-    );
-  }),
+  initiateRequest: jest.fn(
+    (
+      { query, request, onSuccess = () => {} }: DataSourceRequest<SiteWiseDataStreamQuery>,
+      requestInformations: RequestInformationAndRange[]
+    ) => {
+      query.assets.forEach(({ assetId, properties }) =>
+        properties.forEach(({ propertyId }) => {
+          const correspondingRequestInfo = requestInformations.find(({ id }) => `${assetId}---${propertyId}` === id);
```
diehbria, line 28: well want to use use the `toId` method (though i see this was existing code)

packages/core/src/data-module/data-cache/caching/caching.spec.ts
```diff
@@ -369,6 +370,142 @@ describe('getDateRangesToRequest', () => {
...
+          },
+        },
+        resolution: '0',
+        dataStreamId: STREAM_ID,
+        start: START_DATE,
+        end: END_DATE,
+        cacheSettings: DEFAULT_CACHE_SETTINGS,
+      })
+    ).toEqual([]);
+  });
```
diehbria, line 406: nit: line breaks help readability to avoid 400 line blocks of text

packages/core/src/data-module/data-cache/caching/caching.spec.ts
```diff
@@ -369,6 +370,142 @@ describe('getDateRangesToRequest', () => {
...
+        end: END_DATE,
+        cacheSettings: DEFAULT_CACHE_SETTINGS,
+      })
+    ).toEqual([
+      expect.objectContaining({
+        start: new Date(0, 0, 0),
+        end: START_DATE,
+        id: STREAM_ID,
+        resolution: '0',
+        fetchMostRecentBeforeStart: true,
```
diehbria, line 497: I'm confused, why would we want to fetch the most recent before year 0?

packages/core/src/data-module/data-cache/caching/caching.ts
```diff
@@ -104,6 +106,79 @@ export const getDateRangesToRequest = ({
...
+  }
+
+  // convert date ranges to request information, applying fetchMostRecentBeforeEnd to the last range
+  const requestInformations: RequestInformationAndRange[] = dateRanges.map(([rangeStart, rangeEnd], index) => ({
+    start: rangeStart,
+    end: rangeEnd,
+    id: dataStreamId,
+    resolution,
+    fetchMostRecentBeforeEnd: request.settings?.fetchMostRecentBeforeEnd && index === dateRanges.length - 1,
+    fetchFromStartToEnd: request.settings?.fetchFromStartToEnd,
```
diehbria, line 156: how do we capture the scenarios where `start` to `end` is cached by the most recent before end isn't?

diehbria (changes requested):

diehbria (changes requested):
main change is to handle the case where `[start, end]` is cached but we need a request on the segment `(-infinity, start]`

packages/core/src/data-module/data-cache/caching/caching.spec.ts
boweihan: Yeah this approach builds the correct slice to request prior to passing it to the source, but that's actually a different paradigm than what is done for `beforeEnd` so i'll change it to pass the viewport in and have the source request the leading point.

packages/core/src/data-module/data-cache/caching/caching.ts
boweihan: This case is missing, good catch!

packages/core/src/__mocks__/data-source.ts
boweihan: I'm going to leave this for now, this mock is in core, and `toId` is in source-iotsitewise so it'll be a circular dependency to import it. In the future we'll probably want to move this mock out of core.

packages/core/src/data-module/data-cache/caching/caching.ts
```diff
@@ -104,6 +106,94 @@ export const getDateRangesToRequest = ({
...
+    cacheSettings,
+  });
+
+  // convert date ranges to request information
+  const requestInformations: RequestInformationAndRange[] = dateRanges.map(([rangeStart, rangeEnd], index) => ({
+    start: rangeStart,
+    end: rangeEnd,
+    id: dataStreamId,
+    resolution,
+    fetchFromStartToEnd: request.settings?.fetchFromStartToEnd,
```
diehbria, line 151: with the current logic, if `fetchFromStartToEnd` is false, then none of these request informations does anything - we should make sure we don't create `requestInformations` that 'fetch nothing'

packages/core/src/data-module/data-cache/caching/caching.ts
```diff
@@ -104,6 +106,100 @@ export const getDateRangesToRequest = ({
...
+    dataStreamId,
+    start,
+    end,
+    resolution: parseDuration(resolution),
+    cacheSettings,
+  });
+
+  let requestInformations: RequestInformationAndRange[] = [];
+
+  let fetchFromStartToEnd = request.settings?.fetchFromStartToEnd;
```
diehbria, line 147: nit: can use `const`

diehbria (approved):

PR #102, opened 2022-05-13
batching for historical property data requests

boweihan:
## Overview
Batch requests by frame for historical property data

Screen Shot 2022-05-16 at 3 37 28 PM

Implementation notes:
- batches are sliced by maxResults, mirroring the [API behaviour](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_BatchGetAssetPropertyValueHistory.html#iotsitewise-BatchGetAssetPropertyValueHistory-request-entries).
- success and error handling should remain the same everywhere except in the client layer.
- batch timeframe is not finalized, but currently per browser frame.

## Legal
This project is available under the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0.html).

PR #114, opened 2022-05-24
batching for aggregated data requests

boweihan:
## Overview
Batch requests by frame for aggregated property data.

Screen Shot 2022-05-24 at 4 30 47 PM

Implementation re-uses elements from historical batching - https://github.com/awslabs/iot-app-kit/pull/102.

## Legal


.gitignore
```diff
@@ -38,6 +38,7 @@ __diff_output__
 # Cypress screenshots
 **/cypress/screenshots
 **/cypress/videos
+**/cypress/snapshots/All Specs
```
square-li, line 41: Do we need to escape the space character in the path? Like `".../All\ Spacs"`

boweihan: I believe this works, at least it works on my mac.

https://git-scm.com/docs/gitignore

gitignores are line-based and it looks like only trailing spaces need to be escaped.

square-li: I am more worried about Windows. I am going to verify it. But everything else looks good!

square-li: I just verified that it works for windows as well.

square-li (approved):

PR #126, opened 2022-06-13
Mocked out reducer for move action Feature dashboard

sydopy:
## Overview
First stab at creating reducer function to manage actions for undo/redo button. The function creates an inverse action to a move action. The test file, as of now, only checks that it returns an empty array when given an empty dashboard. A move action is defined in the action.ts file.

## Legal


packages/dashboard/src/dashboard-actions/dashboardReducer.ts
```diff
@@ -0,0 +1,25 @@
+import { Reducer } from 'redux';
+import { DashboardConfiguration } from '../types';
+import { getMovedDashboardConfiguration } from './move';
+import { dashboardAction } from './actions';
```
diehbria, line 4: convention to capitalize all types

packages/dashboard/src/dashboard-actions/dashboardReducer.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { dashboardReducer } from './dashboardReducer';
+import { onMoveAction } from './actions';
+import { Position } from '../types';
+
+describe('dashboardReducer', () => {
```
diehbria, line 5: were already in the `dashboardReducer` scope since were in the `dashboardReducer` test suite, there's no need for a describe block that repeats that

packages/dashboard/src/dashboard-actions/dashboardReducer.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  it('returns empty dashboard configuration when provided an empty dashboard', () => {
+    const x: number = 7;
```
diehbria, line 7: you can omit these explicit type declarations, they add verbosity without providing any value (as well as not matching the existing code base standards)

diehbria (changes requested):

diehbria:
also check out the alerts your getting and see if you can resole them first: https://lgtm.com/projects/g/awslabs/iot-app-kit/rev/pr-5bf721e9413260426bfbd64ffb4c0cd3f10484eb

PR #128, opened 2022-06-17
expose batchDuration as a configurable setting on sitewise source

boweihan:
## Overview
The ideal batch duration (timeframe in which requests are batched) observed by the sitewise source is highly dependent on factors such as dashboard configuration, widget configuration, browser performance, and latency. Batching by a single browser tick is fine as a default behaviour but needs to be customizable for the use-case.

This change exposes batchDuration as a setting on the sitewise source (and thus on app kit initialize). batchDuration accepts a numeric value representing the batching window in ms.

Documentation to follow.

## Legal


packages/source-iotsitewise/src/initialize.ts
```diff
@@ -27,7 +27,10 @@ type IoTAppKitInitInputs =
       registerDataSources?: boolean;
       awsCredentials: Credentials | AWSCredentialsProvider<Credentials>;
       awsRegion: string;
-    };
+    }
+) & {
```
NorbertNader, line 31: Why intersection type?

NorbertNader (approved):

PR #129, opened 2022-06-17
Add new Table component

square-li:
## Overview
- Add new Table component 
using AWS-UI Table component plus useCollectionHook to support property filtering & sorting.

- add helpers
iconUtilis and tests - copied from Synchro-charts with modifications.
spinner and tests - copied from Synchro-charts with modifications.
tableHelpers and tests - get default configs for table component and hooks.

TODO:
- add tests of Table component for
thresholds
alarms
- WIP: add the wrapper for the new component in @iot-app-kit/component

## Tests

## Legal

### DIFF
--- a/packages/table/src/table/index.tsx
+++ b/packages/table/src/table/index.tsx
@@ -1,12 +1,12 @@
 import React, { FunctionComponent } from 'react';
 import { PropertyFilter, Table as AWSUITable } from '@awsui/components-react';
 import { useCollection } from '@awsui/collection-hooks';
-import { TableItem, TableProps } from '../utils';
+import { TableProps } from '../utils';
 import { defaultI18nStrings, getDefaultColumnDefinitions } from '../utils/tableHelpers';
 
-export const Table: FunctionComponent<TableProps<TableItem>> = (props) => {
+export const Table: FunctionComponent<TableProps> = (props) => {
   const { items, useCollectionOption = { sorting: {} }, columnDefinitions } = props;
-  const { collectionProps, propertyFilterProps } = useCollection<TableItem>(items, useCollectionOption);
+  const { collectionProps, propertyFilterProps } = useCollection(items, useCollectionOption);
   return (
     <AWSUITable
       className="iot-table"
--- a/packages/table/src/table/index.tsx
+++ b/packages/table/src/table/index.tsx
@@ -9,7 +9,6 @@ export const Table: FunctionComponent<TableProps> = (props) => {
   const { collectionProps, propertyFilterProps } = useCollection(items, useCollectionOption);
   return (
     <AWSUITable
-      className="iot-table"
       {...props}
       {...collectionProps}
       columnDefinitions={getDefaultColumnDefinitions(columnDefinitions)}
--- a/packages/table/src/table/index.tsx
+++ b/packages/table/src/table/index.tsx
@@ -1,0 +1,19 @@
+import React, { FunctionComponent } from 'react';
+import { PropertyFilter, Table as AWSUITable } from '@awsui/components-react';
+import { useCollection } from '@awsui/collection-hooks';
+import { TableItem, TableProps } from '../utils';
+import { defaultI18nStrings, getDefaultColumnDefinitions } from '../utils/tableHelpers';
+
+export const Table: FunctionComponent<TableProps<TableItem>> = (props) => {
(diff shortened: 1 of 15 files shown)


packages/table/src/utils/iconUtils.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { ReactElement, SVGAttributes } from 'react';
+import { StatusIcon } from '@synchro-charts/core';
```
boweihan, line 2: This is just a type right, is it worth it to have a dependency on synchro-charts for this?

packages/table/src/utils/spinner.spec.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
boweihan, line 1: are these tests also copied from synchro-charts?

NorbertNader (approved):

packages/table/src/utils/spinner.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { createRoot, Root } from 'react-dom/client';
+import { act } from 'react-dom/test-utils';
+import { LoadingSpinner } from './spinner';
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
NorbertNader, line 7: Any way to avoid this ts ignore?

packages/table/src/utils/iconUtils.tsx
```diff
@@ -0,0 +1,113 @@
...
+export const getIcons: (
+  name: StatusIcon,
+  color?: string,
+  size?: number
+) => ReactElement<SVGAttributes<unknown>> | undefined = (name: StatusIcon, color?: string, size?: number) => {
+  if (icons[name]) {
+    return icons[name](color, size);
+  }
+  /* eslint-disable-next-line no-console */
+  console.warn(`Invalid status icon requested: ${name}`);
```
NorbertNader, line 111: Can we avoid this console warn and maybe just throw an error instead?

NorbertNader (approved):

packages/table/src/utils/spinner.spec.tsx
square-li: Yes. These tests also copied from sychro-charts with some modifications to make it work with React.

packages/table/src/table/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { useCollection } from '@awsui/collection-hooks';
+import { TableProps } from '../utils';
+import { defaultI18nStrings, getDefaultColumnDefinitions } from '../utils/tableHelpers';
+
+export const Table: FunctionComponent<TableProps> = (props) => {
+  const { items, useCollectionOption = { sorting: {} }, columnDefinitions } = props;
+  const { collectionProps, propertyFilterProps } = useCollection(items, useCollectionOption);
+  return (
+    <AWSUITable
+      className="iot-table"
```
diehbria, line 12: does this class do anything? className will simply be ignored, `class` is the correct keyword. The className is a react concept, and peoples IDEs will sometimes automatically convert to `className` when you use JSX/TSX. But in general don't add classes that aren't necessary

diehbria (changes requested):

packages/table/src/table/index.tsx
square-li: Agreed. I was trying reuse sc-table's styles but abandoned that idea later. Removing it.

packages/table/src/utils/spinner.spec.tsx
square-li: Removed it in the newer commit.

packages/table/src/utils/iconUtils.tsx
square-li: It is an enum as 
```
export declare enum StatusIcon {
    ERROR = "error",
    ACTIVE = "active",
    NORMAL = "normal",
    ACKNOWLEDGED = "acknowledged",
    SNOOZED = "snoozed",
    DISABLED = "disabled",
    LATCHED = "latched"
}
```
 So we would need to keep it and so it can be easier to manage constant values.

packages/table/src/utils/iconUtils.tsx
square-li: No. Throwing an error in rendering could break the rendering process leading to a blank screen, if the error is not properly handled.

diehbria (approved):

PR #132, opened 2022-06-23
AudioAlertPlayer and  AudioAlert

janezhang10:
This PR is for adding AudioAlertPlayer and AudioAlert classes into @iot-app-kit/core.


packages/core/src/common/audioAlerts.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+  }
+
+  public play(audioSrc?: string): Howl | undefined {
+    if (this.isMute) {
+      this.audioSrc = audioSrc;
+      return;
+    }
+
+    console.log('try to play, curr severity: ', this.severity);
```
NorbertNader, line 162: missed console.log

packages/core/src/common/audioAlerts.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+    const { audioAlert1, audioAlert2 } = initializeAudioAlertEnvironment();
+    audioAlert1.getAudioAlertPlayer().stop();
+    const player1 = audioAlert1.play();
+    expect(player1).not.toBe(undefined);
+    const player2 = audioAlert2.play();
+    expect(player2).toBe(undefined);
+    expect(player2).not.toBe(player1);
+  });
+
+  // it("sets the volume", () => {
```
NorbertNader, line 165: Why are these commented?

packages/core/src/common/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const minimumSeverity = 3;
+export const testSound = 'https://s3-us-west-2.amazonaws.com/s.cdpn.io/858/outfoxing.mp3';
```
dpportet, line 2: Are these sounds used anywhere in this commit? If not we should leave it out for now.

packages/core/src/common/audioAlerts.ts
```diff
@@ -0,0 +1,191 @@
+import { Howl } from 'howler';
+import { minimumSeverity, testAudioSrc } from './constants';
+
+export class AudioAlertPlayer {
+  private isMute = false;
+  private isPlay = false;
+  private maxVolume = 1.0;
+  private currentlyPlayingSeverity: number = minimumSeverity;
+  private soundID: number | undefined = undefined;
```
dpportet, line 9: I don't think you need the initialization part here, since an uninitialized variable will be given the value `undefined` anyway

packages/core/src/common/audioAlert.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+  });
+
+  it("player doesn't plays when called on by a muted audio alerts", () => {
+    audioAlert = new AudioAlert({
+      audioAlertPlayer: audioAlertPlayer,
+      isMuted: true,
+    });
+    expect(audioAlert.isMuted()).toBeTrue();
+    audioAlert.play();
+    expect(audioAlert.audioAlertPlayer.isPlaying()).toBeFalse();
```
boweihan, line 48: Do we need to expose the player implementation for isPlaying() in contrast to how you're hiding it with play()?

packages/core/src/common/audioAlert.ts
```diff
@@ -0,0 +1,97 @@
+import { Howl } from 'howler';
+import { AudioAlertPlayer } from './audioAlertPlayer';
+import { mostSevere, leastSevere } from './constants';
+
+export class AudioAlert {
+  private _audioAlertPlayer: AudioAlertPlayer;
+  private _isMute: boolean;
+  private _volume;
```
boweihan, line 8: Looks like you might have missed some explicit types here

packages/core/src/common/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const leastSevere = 3;
+export const mostSevere = 1;
+
+// change when sounds are determined
+export const defaultAudioSrc = [
+  'https://audio-alert-sounds.s3.amazonaws.com/Cute-Chime-2.mp3',
+  'https://audio-alert-sounds.s3.amazonaws.com/new-message-3.ogg',
```
boweihan, line 7: Why the difference in file type?

packages/core/src/common/audioAlertPlayer.ts
```diff
@@ -0,0 +1,112 @@
...
+      this._player?.mute(false, this._soundID);
+    }
+    this._isMuted = false;
+  }
+
+  /* plays the incoming alert if no other alert is playing, or if the incoming alert has
+     a higher severity than the currently playing one. Returns true if incoming alert is
+     played, false otherwise */
+  public play = ({ severity, volume, audioSrc }: { severity: number; volume: number; audioSrc?: string }): boolean => {
+    severity = Math.max(Math.min(severity, leastSevere), mostSevere);
```
boweihan, line 38: might be worth extracting these severity / volume validations into a separate utility.

packages/core/src/common/audioAlertPlayer.ts
```diff
@@ -0,0 +1,112 @@
+import { Howl } from 'howler';
+import { leastSevere, mostSevere, defaultAudioSrc } from './constants';
+
+export class AudioAlertPlayer {
```
boweihan, line 4: The java in me wants you to make the interface for this player explicit :)

packages/core/src/common/audioAlert.ts
```diff
@@ -0,0 +1,97 @@
+import { Howl } from 'howler';
+import { AudioAlertPlayer } from './audioAlertPlayer';
+import { mostSevere, leastSevere } from './constants';
+
+export class AudioAlert {
+  private _audioAlertPlayer: AudioAlertPlayer;
```
NorbertNader, line 6: Maybe could be simplified to something like

```

type AudioAlertConfig {
  audioAlertPlayer: AudioAlertPlayer;
  isMute: boolean;
  volume;
  severity;
  player: Howl | undefined;
  soundID: number | undefined;
  audioSrc: string | undefined;
}


export class AudioAlert {
  private config: AudioAlertConfig;

  constructor(config: AudioAlertConfig) {
    this.config = config;
  }
```

NorbertNader (approved):

packages/core/src/common/audioAlertPlayer.ts
```diff
@@ -0,0 +1,112 @@
...
+          // updates AudioAlertPlayer when sound ends
+          this.stop();
+        },
+      });
+
+      this._soundID = newPlayer.play();
+      this._player = newPlayer;
+      this._severity = severity;
+      this._isPlaying = true;
+      if (this.isMuted()) {
```
boweihan, line 59: Does it make sense to retain the "muted" state across different instances of alerts? Also I notice you call play() prior to muting - will this play the sound, or will it always mute first due to the order of execution of synchronous code / promise queue?

boweihan (commented):
Looking great so far! Some minor structural comments and a question about muting.

packages/core/src/common/audioAlert.spec.ts
janezhang10: Nice catch, I forgot to refactor this test when I changed `play()`

packages/core/src/common/constants.ts
janezhang10: I just found a few royalty free sound samples online and uploaded it to my own S3 bucket to play with. These sounds aren't defined yet and I will have to review with Mahti, Courtney, and Tracy to finalize what the default sounds will be.
They're both audio file types, OGG is just has better sound quality, a smaller file size, and higher bitrate.

packages/core/src/common/audioAlertPlayer.ts
janezhang10: It seems like in `IotBlackPearlApp` classes aren't explicit; `eslint` also doesn't let me do that :)

boweihan: How about defining an interface and having this class implement it?

boweihan (approved):

packages/core/src/common/types.ts
```diff
@@ -39,3 +40,46 @@ export type DataModuleSession = {
...
+  audioAlertPlayer: AudioPlayer;
+  isMuted: boolean;
+  volume: number;
+  severity: number;
+  player: Howl | undefined;
+  soundID: number | undefined;
+  audioSrc: string | undefined;
+};
+
+export interface AudioAlertNode {
```
dpportet, line 76: Is there any specific reason to call this a `Node`? If not, I think it would be more clear and simple to call this `AudioAlert`

dpportet (approved):

PR #138, opened 2022-06-29
Dashboard delete widget action

diehbria:
## Overview
Implement the ability to delete the widgets that are selected.

Also minor cleanup of unused code on `iot-dashboard`.

## Legal

### DIFF
--- a/packages/dashboard/src/components/iot-dashboard/iot-dashboard.tsx
+++ b/packages/dashboard/src/components/iot-dashboard/iot-dashboard.tsx
@@ -7,6 +7,7 @@ import { getMovedDashboardConfiguration } from '../../dashboard-actions/move';
 import { getSelectionBox } from './getSelectionBox';
 import { DASHBOARD_CONTAINER_ID, getDashboardPosition } from './getDashboardPosition';
 import { trimWidgetPosition } from './trimWidgetPosition';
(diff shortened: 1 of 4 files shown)


packages/dashboard/src/components/iot-dashboard/iot-dashboard.tsx
```diff
@@ -307,6 +317,14 @@ export class IotDashboard {
     this.onGestureEnd();
   }
 
+  @Listen('keydown')
+  onKeyDown({ key }: KeyboardEvent) {
```
jmbuss, line 321: Is there a button to do this also? I know figjam doesn't have this, but they have a button that shows on hover to duplicate. I think the keyboard shortcuts are great for people doing a lot of dashboard work, but a first time user might not think to do this.

diehbria: I'm unsure. I agree on the discoverability. As we start fleshing out the UX of the dashboard more we can explore that option

jmbuss: Ya, doesn't change the underlying method. Just a thought to build in! lgtm otherwise.

jmbuss (approved):

PR #150, opened 2022-07-08
sync changes with private repo for common files

sheilaXu:
## Overview
sync changes with private repo for common files to have less conflicts

## Tests

## Legal


packages/related-table/package.json
```diff
@@ -36,8 +36,8 @@
     "@awsui/collection-hooks": "^1.0.0",
     "@awsui/components-react": "^3.0.0",
     "@awsui/design-tokens": "^3.0.0",
-    "react": ">=17.0.2",
-    "react-dom": ">=17.0.2",
+    "react": "^17.0.2",
```
diehbria, line 39: why the change in range for react versions?

sheilaXu: the current version locks react and react-dom to v18, but our new package installs a v17 which then caused multi react error. our new package was using v16, it can build with v17 but it could take a while to migrate itself and its dependencies to v18. the react-components packages also uses v17. so if there isn't a hard requirement to support v18, can we target v17?

square-li (approved):

PR #154, opened 2022-07-09
Add a global audioAlertPlayer

janezhang10:
Make a global `audioAlertPlayer`; change names and tests for other classes accordingly.

We want to have a global `AudioPlayer` that is exported from core to provide the convenience factor of setting it up. We   want to have a global `AudioPlayer` that controls global volume and can mute/unmute, but this `AudioPlayer` shouldn't have to be a prop passed into the components. There's no need or use case for having multiple `AudioPlayer`s, since the main idea was to only have one overarching `AudioPlayer` that "manages" the `AudioAlert`s by making sure that no more than one `AudioAlert` plays at a time, hence instantiating a global `AudioPlayer` exported by core makes sense.

If a user doesn't want audio alerts in their components, there will be a boolean flag passed in to signify whether they want audio alerts or not. Default will be false, AKA audio alerts are turned off. This is because audio alerts is a feature that users may or may not be aware of, and we don't want developers or users to unexpectedly hear a sound and be startled.


packages/core/src/common/audioAlertPlayer.ts
```diff
@@ -1,115 +1,3 @@
...
-  private calculateMaxVolume(maxVolume: number): number {
-    return Math.max(Math.min(maxVolume, 1.0), 0.0);
-  }
-
-  private isMoreSevere(newSeverity: number): boolean {
-    // Lower number corresponds to higher severity
-    return this.config.severity ? newSeverity < this.config.severity : false;
-  }
-}
+export const audioAlertPlayer = new AudioPlayer();
```
boweihan, line 3: What's the purpose of making the audio player global? Would be nice if that was in the PR description :).

As far as I can see, making it global in this way makes it convenient to use and guarantees that it's a singleton across an entire application build - but doesn't this make it harder to pass in different audio players? What if a user doesn't want to use audio, can we avoid instantiating a player in that case?

janezhang10: Added the description, sorry about that -- will keep this in mind for future PR's!

janezhang10: It makes sense to have the player instantiated in either case, because an `AudioPlayer` is not bulky. Otherwise, users would have to instantiate a player themselves, instead of simply setting a boolean flag to true/false. This complicates the set up for app kit, which is not what we want.

NorbertNader (approved):

packages/core/src/common/audioAlertPlayer.ts
NorbertNader, line 3: nit: could just do this in `audioAlert.ts`

PR #158, opened 2022-07-12
Integrate Audio Alert into Threshold - Helper Functions

janezhang10:
## Summary
`audioAlertManager` contains helper functions to determine whether an audio alert should be played or not when given annotations and datastreams.

## More in detail
`isLiveData`: Audio alerts will only play if newly pushed points within view of the viewport breaches a threshold. This function determines whether the viewport is in Live Mode or if the end of the viewport is past the current date -- both of these cases will result in users seeing newly pushed points on their visualizations, hence audio alerts will play. 

`initializeAudioAlerts`: Creates a map from threshold or threshold id to an Audio Alert instance. This allows for each threshold to be customizable with its own volume and audio source. Thresholds that are already in the map should not get a new `AudioAlert` linked to it.

`playThresholdAudioAlert`: Checks if users can see newly pushed point and plays an audio alert if a threshold is breached.

## Legal

### DIFF
--- a/packages/core/src/common/audioAlertManager.ts
+++ b/packages/core/src/common/audioAlertManager.ts
@@ -15,7 +15,7 @@ import { getDataPoints } from './getDataPoints';
 import { isNumberDataStream } from './predicates';
 
 // returns true if viewport is in Live Mode or if viewport.end is past the current time and date
-const isLiveData = (viewport: MinimalViewPortConfig): boolean => {
+export const isLiveData = (viewport: MinimalViewPortConfig): boolean => {
   // duration in viewport if in live mode
   if ('duration' in viewport) {
     return true;
--- a/packages/core/src/common/audioAlertManager.ts
+++ b/packages/core/src/common/audioAlertManager.ts
@@ -9,7 +9,6 @@ import {
 import { DataStream } from '../data-module/types';
 
 import { AudioAlert } from './audioAlert';
-import { AudioAlertPlayer } from './audioAlertPlayer';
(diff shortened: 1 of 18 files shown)


packages/core/src/common/audioAlertManager.ts
```diff
@@ -0,0 +1,89 @@
...
+import { isNumberDataStream } from './predicates';
+
+// returns true if viewport is in Live Mode or if viewport.end is past the current time and date
+export const isLiveData = (viewport: MinimalViewPortConfig): boolean => {
+  // duration in viewport if in live mode
+  if ('duration' in viewport) {
+    return true;
+  } else {
+    // if the viewport is past the current time then we expect audio alerts to behave as if we're in live mode
+    const endTimestamp = typeof viewport.end === 'string' ? new Date(viewport.end).getTime() : viewport.end.getTime();
```
boweihan, line 24: can we use the helper `viewportEndDate` in `viewport.ts` here instead?

packages/core/src/common/audioAlertManager.ts
```diff
@@ -0,0 +1,89 @@
...
+    return endTimestamp + liveDataTimeBuffer > Date.now();
```
boweihan, line 25: nit: would be nice to have a small comment explaining why we choose 1.5 seconds specifically.

boweihan (commented):
Thanks for the nicely detailed PR description!

boweihan (approved):

PR #162, opened 2022-07-14
Basic resizable panes for dashboard

mpopv:
Screen Shot 2022-07-14 at 12 39 15 PM

## Overview
This PR adds a new component, `<resizable-panes/>`. It consists of a left, center, and right pane, with handles between them that dynamically change the widths of the panes when dragged left or right, similar to a library like [Split.js](https://split.js.org/).

This is a basic first step, and needs to be followed up by several additional changes adding enhancements and covering anticipated edge cases, detailed below.

## Next steps
- We will need to expand the mouse event listeners here to touch event listeners as well. [This is coming shortly in a follow-up PR.]
- We will need to save the state of the pane widths in sessionStorage so they can persist across page refreshes. [This is coming shortly in a follow-up PR.]
- We need to add a window resize listener that recalculates the pane widths when the window is resized. (Or potentially not if we figure out a way to automate this with grid values?) [This is coming shortly in a follow-up PR.]
- We have basic unit tests here that check that the panes are successfully rendering their content. I attempted, in vain, to also implement full functionality unit tests that dispatch mouse events to drag the handles and check that the associated style value has updated successfully, but it looks like Stencil's testing framework only allows a limited set of mock events, and those don't seem to include MouseEvents beyond `click`, so we will likely need to implement this in an e2e test instead.
- I haven't yet added any logic to enforce minimum/maximum pane width or prevent collision bugs. In manual testing, it seems to 'just work' as is, but my guess is we will run into weird problems with panes overlapping, becoming undraggable, etc. unless we explicitly seek to prevent this, and/or add a recovery mechanism like a 'reset panes to default' button.
- The styles involved here are meant to be temporary: the drag handles are rather ugly, I have inserted some demonstration content in the left/right panes that should be replaced by the correct components, and I didn't put too much time into making sure the pane width calculations were perfect, so there might be a better way to achieve them with different grid values.

## Tests

## Legal

### DIFF
--- a/packages/dashboard/src/components.d.ts
+++ b/packages/dashboard/src/components.d.ts
@@ -86,6 +86,8 @@ export namespace Components {
          */
         "width": number;
     }
+    interface IotResizablePanes {
+    }
     interface IotSelectionBox {
         "cellSize": number;
         "height": number;
@@ -126,6 +128,12 @@ declare global {
         prototype: HTMLIotDashboardWrapperElement;
         new (): HTMLIotDashboardWrapperElement;
     };
+    interface HTMLIotResizablePanesElement extends Components.IotResizablePanes, HTMLStencilElement {
+    }
+    var HTMLIotResizablePanesElement: {
+        prototype: HTMLIotResizablePanesElement;
+        new (): HTMLIotResizablePanesElement;
+    };
     interface HTMLIotSelectionBoxElement extends Components.IotSelectionBox, HTMLStencilElement {
     }
     var HTMLIotSelectionBoxElement: {
@@ -149,6 +157,7 @@ declare global {
         "iot-dashboard-dynamic-widget": HTMLIotDashboardDynamicWidgetElement;
         "iot-dashboard-widget": HTMLIotDashboardWidgetElement;
         "iot-dashboard-wrapper": HTMLIotDashboardWrapperElement;
+        "iot-resizable-panes": HTMLIotResizablePanesElement;
         "iot-selection-box": HTMLIotSelectionBoxElement;
         "iot-selection-box-anchor": HTMLIotSelectionBoxAnchorElement;
         "testing-ground": HTMLTestingGroundElement;
@@ -232,6 +241,8 @@ declare namespace LocalJSX {
          */
         "width"?: number;
     }
+    interface IotResizablePanes {
+    }
     interface IotSelectionBox {
         "cellSize"?: number;
         "height"?: number;
@@ -251,6 +262,7 @@ declare namespace LocalJSX {
         "iot-dashboard-dynamic-widget": IotDashboardDynamicWidget;
         "iot-dashboard-widget": IotDashboardWidget;
         "iot-dashboard-wrapper": IotDashboardWrapper;
+        "iot-resizable-panes": IotResizablePanes;
         "iot-selection-box": IotSelectionBox;
         "iot-selection-box-anchor": IotSelectionBoxAnchor;
         "testing-ground": TestingGround;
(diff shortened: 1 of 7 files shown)


packages/dashboard/src/components/iot-dashboard/iot-selection-box/iot-selection-box-anchor.css
```diff
@@ -49,26 +49,26 @@
   cursor: ns-resize;
   height: var(--selection-border-touch-width);
   width: 100%;
-  top: calc(-1/2 * var(--selection-border-touch-width));
+  top: calc(-1 / 2 * var(--selection-border-touch-width));
```
mpopv, line 52: Linter forced me to do this for some reason

packages/dashboard/src/components.d.ts
```diff
@@ -130,13 +139,16 @@ declare namespace LocalJSX {
...
+    interface ResizablePanes {
+    }
     interface TestingGround {
     }
     interface IntrinsicElements {
         "iot-dashboard": IotDashboard;
         "iot-dashboard-widget": IotDashboardWidget;
         "iot-selection-box": IotSelectionBox;
         "iot-selection-box-anchor": IotSelectionBoxAnchor;
+        "resizable-panes": ResizablePanes;
```
diehbria, line 151: best practice with webcomponents to have everything name spaced, since they are globally registered and collide (i.e. add the `iot-` prefix to all the components)

packages/dashboard/src/components/resizable-panes/resizable-panes.tsx
```diff
@@ -0,0 +1,108 @@
...
+  /** Last seen mouse x position during a drag, in px from screen left side */
+  @State() lastSeenAtX: number | null = null;
+
+  /** Total x distance moved during a drag, in px */
+  @State() movedX: number | null = null;
+
+  /** Current widths of the three panes, in percentage of screen width */
+  @State() leftPaneWidth: number = window.innerWidth * LEFT_WIDTH_PERCENT;
+  @State() centerPaneWidth: number = window.innerWidth * CENTER_WIDTH_PERCENT;
+  @State() rightPaneWidth: number = window.innerWidth * RIGHT_WIDTH_PERCENT;
```
diehbria, line 25: will this still work if the dashboard isn't utilized as a full page widget?

diehbria (changes requested):

packages/dashboard/src/components/resizable-panes/resizable-panes.tsx
mpopv: No it won't - great point. I pushed a revised version that determines these by measuring the parent element instead. 

As a bonus, in the process I replaced a bunch of JS with CSS grid, so the component is simpler now.

There's only one downside: to measure the actual width of the parent element, the component needs to render, measure, and then re-render. Stencil complains about this with console warnings about unnecessary re-renders.

packages/dashboard/src/components.d.ts
mpopv: Done!

packages/dashboard/src/components/iot-resizable-panes/iot-resizable-panes.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    <div slot="center">
+      <p>${centerSlotContent}</p>
+    </div>
+    <div slot="right">
+      <p>${rightSlotContent}</p>
+    </div>
+  </iot-resizable-panes>
+`;
+
+describe('iot-resizable-panes', () => {
```
diehbria, line 24: nit: we can remove this describe block, already in context of `iot-resizeable-panes` file

diehbria (approved):

packages/dashboard/src/components/iot-resizable-panes/iot-resizable-panes.spec.ts
mpopv: Will do in follow-up PR

PR #164, opened 2022-07-15
feat: Integrate Audio Alert with Thresholds on Component Level

janezhang10:
## Overview
Add audio alert support with thresholds by calling helper functions from `@iot-app-kit/core` and adding props that need to be passed in.

Changes: 
- New prop: `audioAlertsEnabled`: boolean flag as to whether audio alerts are enabled or not.
- remove initialization of audio alerts
- add `audioAlert` as a property to `Threshold`
- add jest tests that watches `Howl` calls
- use react dev properties instead of a manual `localDev`
----------------
Also a slight modification with `AudioAlertManager`, where I remove the check for using number data streams. This is because the check is done in `breachedThreshold` as well, so this check would have been extraneous. Also moved the `isLiveData` check earlier in the function to account for any time lag that might happen while looping through `dataStreams`.

## Legal


packages/components/src/components/iot-bar-chart/iot-bar-chart.tsx
```diff
@@ -34,6 +35,10 @@ export class IotBarChart {
 
   @Prop() styleSettings: StyleSettingsMap | undefined;
 
+  @Prop() audioAlertsEnabled = false;
```
boweihan, line 38: personal pref: "enableAudioAlerts"

boweihan (changes requested):

packages/components/src/components/iot-bar-chart/iot-bar-chart.tsx
janezhang10: "enableAudioAlerts" follows the same pattern as `enableXScroll` in Synchro Charts, I'll change this :+1:

boweihan (approved):

NorbertNader (approved):

packages/core/src/common/audioAlertManager.spec.ts
```diff
@@ -58,45 +21,55 @@ describe('isLiveData', () => {
...
+    expect(threshold1.audioAlert.volume).toBe(0.5);
   });
 });
 
 describe('playThresholdAudioAlert', () => {
+  let threshold1: Threshold;
+
   beforeEach(() => {
     audioAlertPlayer.stop();
+    threshold1 = {
```
diehbria, line 57: why is this necessary?

Issue #153, opened 2022-07-08
Data incorrectly pushed to series-connector


janezhang10:
**Describe the bug**
Data is being pushed upon every Live tick to `iot-time-series-connector` when in Live mode. The `prevDataCache` for `hasRequestedInformationChanged()` is sometimes undefined, which causes `hasRequestedInformationChanged` to always return true.

**To Reproduce**
1. Have a working testing-ground with working assetIds
2. Produce a time series chart with duration '1m' (preferably line chart so that the data is aggregated compared to scatter)
3. Have some console logs in `hasRequestedInformationChanged` that details `prevDataStreamStore`
4. Run the App Kit locally and inspect the page
5. See that `prevDataStreamStore` is occasionally undefined in console

**Expected behavior**
Only new incoming data should be pushed to `iot-time-series-connector` in Live mode. `hasRequestedInformationChanged` should only return true if there's new incoming data that needs to be rendered, and `prevDataCache` should only be undefined upon initialization of `DataCache`.

**Screenshots**
example of `prevDataStreamStore` being undefined while `currDataStreamStore` is defined after letting App Kit run locally for a while.
<img width="487" alt="Screen Shot 2022-07-08 at 3 59 35 PM" src="https://user-images.githubusercontent.com/40257222/178080002-3f0cff93-2a6d-45d7-bd62-dc3c232d5b79.png">
<img width="1067" alt="Screen Shot 2022-07-08 at 4 00 38 PM" src="https://user-images.githubusercontent.com/40257222/178080078-6406075d-3935-4a4f-a70a-24eaa9dd6d28.png">

**Desktop (please complete the following information):**
 - OS: macOS Monterey
 - Browser Firefox
 - Version 91.10.0esr

**Additional context**
The frequency of `prevDataCache` being undefined depends on the duration of the viewport. The smaller the duration (e.g. 1 minute), the more often `prevDataCache` is undefined (about every other tick).

NorbertNader:
This issue no longer exists in latest version.

PR #179, opened 2022-07-26
feat: audio alert for thresholds

janezhang10:
## Summary
Add an audio player feature in @iot-app-kit/core to allow developers to play sounds in App Kit. Also add an audio alert feature that plays an audible alert whenever a threshold is breached by a visible new data point.

**Note**: This PR is complete and ready for review but it should not be opened or merged in until the following bugs are fixed and merged into `main` to prevent incorrect behaviors of audio alerts:

## Breaking Blockers

- https://github.com/awslabs/synchro-charts/issues/150
points constantly rerender and cause audio alerts to play over and over again even though there's no new points. Make sure to update Synchro Chart package version when this is fixed.
- https://github.com/awslabs/iot-app-kit/issues/153
points constantly rerender and cause audio alerts to play over and over again even though there's no new points.

## Non-breaking blocker
- https://github.com/awslabs/synchro-charts/issues/153
This is an issue where thresholds aren't properly calculated as breaching, however audio alerts are meant to reflect what's displayed on Synchro Charts. If a threshold isn't displayed as being breached then no audio alerts are played, which follows the correct behavior as audio alerts. When this issue is fixed, audio alerts will follow along with the correct behavior as well without any necessary additional changes.

## Legal


packages/components/src/components/iot-time-series-connector/iot-time-series-connector.spec.ts
```diff
@@ -337,3 +341,78 @@ it('when assignDefaultColors is true, provides a default color', async () => {
     viewport,
   });
 });
+
+it('plays audio alert if audio alerts are enabled and the breached threshold has an AudioAlert', async () => {
+  audioAlertPlayer.stop();
```
diehbria, line 346: it looks like this may be a good line to move into a afterEach/beforeEach hook

packages/core/src/common/audioAlertManager.ts
```diff
@@ -0,0 +1,63 @@
...
+  const thresholds = getThresholds(annotations);
+  if (thresholds.length === 0) {
+    return;
+  }
+  // audio alerts are only applied to live data
+  dataStreams.forEach((dataStream: DataStream) => {
+    const allVisiblePoints = getVisibleData(getDataPoints(dataStream, dataStream.resolution), viewport);
+    if (allVisiblePoints.length != 0) {
+      const latestPoint = allVisiblePoints[allVisiblePoints.length - 1];
+      const breachedThresh: Threshold | undefined = breachedThreshold({
```
diehbria, line 46: shouldn't need to declare the type here, we can just let the types flow, as breachedThreshold should already correctly specify the types

packages/core/src/common/audioAlertManager.ts
```diff
@@ -0,0 +1,63 @@
...
+        value: latestPoint.y,
+        date: new Date(latestPoint.x),
+        thresholds: thresholds,
+        dataStreams: [],
+        dataStream: dataStream as SynchroChartsDataStream,
+      });
+      if (breachedThresh && breachedThresh.audioAlert) {
+        const severity = calculateSeverity(breachedThresh.severity ?? leastSevere);
+        audioAlertPlayer.play({
+          severity: severity,
```
diehbria, line 56: prefer short hand notation (though ideally this should be enforced by a linter)

packages/core/src/common/audioPlayer.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    audioAlertPlayer.stop();
+
+    expect(howlMock.stop).toHaveBeenCalledOnce();
+    expect(audioAlertPlayer.isPlaying()).toBeFalse();
+  });
+});
+
+describe('setMaxVolume', () => {
+  const audioAlertPlayer = new AudioPlayer();
```
diehbria, line 71: better to just declare this in each test, instead of having one instance shared

packages/core/src/common/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+// change when sounds are determined
+export const defaultAudioSrc = [
+  'https://audio-alert-sounds.s3.amazonaws.com/Cute-Chime-2.mp3',
+  'https://audio-alert-sounds.s3.amazonaws.com/new-message-3.ogg',
+  'https://audio-alert-sounds.s3.amazonaws.com/new-message-5.ogg',
+];
+
+// 1500 ms is approximately the amount of lag time between the end viewport of a live viewport and real time
+// This number was accidentally discovered in the IoT App Kit when Synchro Charts had this issue: https://github.com/awslabs/synchro-charts/issues/150
+export const liveDataTimeBuffer = 1500;
```
diehbria, line 13: i think it's a good practice to explicitly state the units within the variable name itself. i.e. liveDataTimeBufferMs or what have you

packages/core/src/common/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+export const mostSevere = 1;
+
+// change when sounds are determined
+export const defaultAudioSrc = [
+  'https://audio-alert-sounds.s3.amazonaws.com/Cute-Chime-2.mp3',
+  'https://audio-alert-sounds.s3.amazonaws.com/new-message-3.ogg',
+  'https://audio-alert-sounds.s3.amazonaws.com/new-message-5.ogg',
+];
+
+// 1500 ms is approximately the amount of lag time between the end viewport of a live viewport and real time
```
diehbria, line 11: can you elaborate?

packages/core/src/common/audioPlayer.ts
```diff
@@ -0,0 +1,112 @@
...
+  return Math.max(Math.min(maxVolume, 1.0), 0.0);
+};
+export class AudioPlayer implements AudioPlayerInterface {
+  config: AudioPlayerConfig;
+  private soundID: number | undefined;
+  private player: Howl | undefined;
+  private severity: number | undefined;
+
+  constructor() {
+    this.config = {
```
diehbria, line 19: we don't need to do this in a constructor, we can just say `config: AudioPlayerConfig = {...}` (preferred)

packages/core/src/common/audioPlayer.ts
```diff
@@ -0,0 +1,112 @@
...
+  public isPlaying(): boolean {
+    return this.config.isPlaying;
+  }
+
+  public isMuted(): boolean {
+    return this.config.isMuted;
+  }
+
+  public mute(): void {
+    if (this.player && this.soundID) {
```
diehbria, line 35: what if soundID is 0? Or a negative number?

packages/core/src/common/audioPlayer.ts
```diff
@@ -0,0 +1,112 @@
...
+      this.stop();
+    }
+
+    volume = Math.max(Math.min(volume, 1.0), 0.0);
+    if (!this.isPlaying()) {
+      const newPlayer = new Howl({
+        src: [audioSrc ?? defaultAudioSrc[severity - 1]],
+        volume: volume * this.config.maxVolume,
+        mute: this.isMuted(),
+        html5: process.env.NODE_ENV === 'development', //Web Audio API doesn't work on local development due to cross origin
```
diehbria, line 64: space after `//`

diehbria (changes requested):

packages/core/src/common/audioPlayer.ts
janezhang10: negative numbers are truthy values in typescript, but will add the case that `soundID` is 0

packages/core/src/common/audioAlertManager.ts
```diff
@@ -0,0 +1,63 @@
...
+    return;
+  }
+  const thresholds = getThresholds(annotations);
+  if (thresholds.length === 0) {
+    return;
+  }
+  // audio alerts are only applied to live data
+  dataStreams.forEach((dataStream: DataStream) => {
+    const allVisiblePoints = getVisibleData(getDataPoints(dataStream, dataStream.resolution), viewport);
+    if (allVisiblePoints.length != 0) {
```
jmbuss, line 44: use triple `!==` for comparisons. Prefer to return early so you have less code nested in a block. 
```
if (allVisiblePoints.length === 0) return;
// rest of the code.
```

packages/core/src/common/audioAlertManager.ts
```diff
@@ -0,0 +1,63 @@
...
+      const latestPoint = allVisiblePoints[allVisiblePoints.length - 1];
+      const breachedThresh = breachedThreshold({
+        value: latestPoint.y,
+        date: new Date(latestPoint.x),
+        thresholds: thresholds,
+        dataStreams: [],
+        dataStream: dataStream as SynchroChartsDataStream,
+      }) as Threshold;
+      if (breachedThresh && breachedThresh.audioAlert) {
+        const severity = calculateSeverity(breachedThresh.severity ?? leastSevere);
```
jmbuss, line 54: should we make the severity optional since the implementation applies a default?

jmbuss (changes requested):

packages/core/src/common/audioAlertManager.ts
janezhang10: We could but the code would end up looking about the same either way, since I have to call on `calculateSeverity` to make sure that the severity given to the threshold is at most 1 and at least 3. i think the code for this part is alright.

packages/core/src/common/audioAlertManager.ts
```diff
@@ -0,0 +1,65 @@
...
+}): Map<string, AudioAlert> | undefined => {
+  if (!isLiveData(viewport)) {
+    return;
+  }
+  const thresholds = getThresholds(annotations);
+  if (thresholds.length === 0) {
+    return;
+  }
+
+  // Audio alerts are only applied to live data
```
diehbria, line 42: i'm confused as to how this only applies to live data? it looks like it applies to any data within the current viewport

diehbria (changes requested):

diehbria:
is this safe to merge in with the outstanding bugs you have outlined in the overview?

janezhang10:
> is this safe to merge in with the outstanding bugs you have outlined in the overview?

No it is not, the bugs must be fixed beforehand otherwise the behavior of audio alerts would not be as expected.

packages/core/src/common/audioAlertManager.ts
janezhang10: I should move this comment to earlier in the code where I call on `isLiveData`.

tracfren:
What is the plan to merge this PR in?

TheEvilDev:

PR #187, opened 2022-08-01
Persist resizable pane position in sessionStorage and prevent collisions on window resize

mpopv:
## Overview
This PR adds two followup changes from #162, the resizable pane PR:

1. It persists pane position in sessionStorage, so the page can be refreshed and they will appear to not move at all -- it 'just works'.

2. It adds a window resize listener that will, if the resize would cause the panes to collide with one another, scales them down proportionally, as if they are being 'squished together'. This is necessary to prevent situations where one pane's handle is not grabbable due to panes overlapping.

## Tests
https://github.com/awslabs/iot-app-kit/runs/7616444204?check_suite_focus=true

## Legal

### DIFF
--- a/packages/dashboard/src/components/iot-resizable-panes/iot-resizable-panes.css
+++ b/packages/dashboard/src/components/iot-resizable-panes/iot-resizable-panes.css
@@ -2,7 +2,7 @@
   display: grid;
   width: 100%;
   max-width: 100%;
-  grid-template-columns: min-content 10px auto 10px min-content;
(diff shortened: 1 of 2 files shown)


diehbria (approved):
looks good! I think we could add an integration test that actually resizes them - but this can be merged in without that

mpopv:
> looks good! I think we could add an integration test that actually resizes them - but this can be merged in without that

Yeah, that's coming next; was hoping to include that in this PR but didn't finish by Friday

PR #190, opened 2022-08-05
fix: add edge case check and fix documentation

janezhang10:
## Overview
Add an edge case about what if `soundID` is 0, remove the constructor in `AudioAlertPlayer`, fix documentation style, and change name from `AudioPlayerInterface` to `AudioPlayer`.

## Legal

### DIFF
--- a/packages/core/src/common/audioAlertPlayer.ts
+++ b/packages/core/src/common/audioAlertPlayer.ts
@@ -1,6 +1,6 @@
 import { Howl } from 'howler';
 import { leastSevere, mostSevere, defaultAudioSrc } from './constants';
-import { AudioPlayerInterface, AudioPlayerConfig } from './types';
+import { AudioPlayer, AudioPlayerConfig } from './types';
 
 export const calculateSeverity = (severity: number): number => {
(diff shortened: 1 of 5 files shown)


packages/core/src/common/audioAlertPlayer.ts
```diff
@@ -32,7 +28,7 @@ export class AudioPlayer implements AudioPlayerInterface {
   }
 
   public mute(): void {
-    if (this.player && this.soundID) {
+    if (this.player && (this.soundID || this.soundID == 0)) {
```
dpportet, line 31: Can we add a test which validates the soundId = 0 case?

janezhang10: Not sure if this is possible since `soundID` is a private parameter that is assigned to a value that Howler returns

dpportet: Ok, making the variable public probably isn't worth it just for this test since the extra public variable could be confusing for customers, so it should be ok without a test.

dpportet (approved):

PR #192, opened 2022-08-09
chore(dashboard): add env plugin for local development, cleanup dynam…

jmbuss:
## Overview
Moves @iot-app-kit/source-iotsitewise to devDependencies (only used in testing grounds)
Cleanup dynamic-widget implementation.
adds env plugin, can now define an env file with credentials and assets

## Tests

## Legal


diehbria (approved):

mpopv:
Is the .env file gitignored? Don't see any gitignore changes in this change and don't see .env in the existing gitignore file.

Probably should add that if not; playing with fire otherwise.

tjuranek:
> Is the .env file gitignored? Don't see any gitignore changes in this change and don't see .env in the existing gitignore file.
> 
> Probably should add that if not; playing with fire otherwise.

If we don't want to commit the `.env` file, we could commit an `.env.example` file so we at least know the expected keys?

jmbuss:
> > Is the .env file gitignored? Don't see any gitignore changes in this change and don't see .env in the existing gitignore file.
> > Probably should add that if not; playing with fire otherwise.
> 
> If we don't want to commit the `.env` file, we could commit an `.env.example` file so we at least know the expected keys? You could also add ` && cp .env.example .env` to the `bootstrap` script too if you decide to add one.

Great idea! will add

PR #193, opened 2022-08-09
fix(dashboard):Remove previous position from internal dashboard

sydopy:
## Overview
Move managing the previous widget position into dashboard state instead of the internal dashboard

## Tests

## Legal

### DIFF
--- a/packages/dashboard/src/dashboard-actions/dashboardReducer.ts
+++ b/packages/dashboard/src/dashboard-actions/dashboardReducer.ts
@@ -21,6 +21,7 @@ const initialState: DashboardState = {
   cellSize: 10,
   undoQueue: [],
   redoQueue: [],
+  previousPosition: undefined,
 };
 
 export const dashboardReducer: Reducer<DashboardState, DashboardAction> = (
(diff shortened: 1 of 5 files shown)


packages/dashboard/src/dashboard-actions/dashboardReducer.ts
```diff
@@ -34,10 +35,11 @@ export const dashboardReducer: Reducer<DashboardState, DashboardAction> = (
         return {
           ...state,
           selectedWidgetIds: action.payload.widgetIds,
+          previousPosition: action.payload.position,
           intermediateDashboardConfiguration: move({
             dashboardConfiguration: state.intermediateDashboardConfiguration || state.dashboardConfiguration,
             position: action.payload.position,
-            previousPosition: action.payload.prevPosition,
+            previousPosition: state.previousPosition || action.payload.position,
```
diehbria, line 42: if there is no previous position, shouldn't it be `undefined`?

sydopy: if it is left undefined, it kind of jumps all over the place when you try to move it. this is because of how the move() is written. i'm not sure if it's a bug, but if previous position doesn't exist, the delta returned is the current position rather than the delta being 0. So the way to fix that bug is either include action.payload.position, or i can edit your code. i just wasn't sure if you wanted me to touch move

diehbria (approved):

PR #152, opened 2022-07-08
Time point annotations

raymfang:
## Overview
Added time-point annotation functionality to the dashboard. Time-point annotations can be created, edited, and deleted, filtered by visibility, and have label/color specified for each widget on the dashboard.

## Legal
This project is available under the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0.html).

PR #198, opened 2022-08-10
feat(dashboard): add time-point annotation functionality

raymfang:
## Overview
Added time-point annotation functionality to the dashboard. Time-point annotations can be created, edited, and deleted, filtered by visibility, and have label/color specified for each widget on the dashboard.

NOTE: Opening this PR as a continuation of the original PR for time-point annotations (https://github.com/awslabs/iot-app-kit/pull/152), which was automatically merged by Github (a '0 commits merged') after a force-push to the PR's branch. Comments from this original PR have been addressed and added to the code.

Screen Shot 2022-08-12 at 9 45 31 AM

## Tests

## Legal

### DIFF
--- a/packages/dashboard/src/components/iot-annotations/annotations.ts
+++ b/packages/dashboard/src/components/iot-annotations/annotations.ts
@@ -1,0 +1,71 @@
+import { XAnnotation } from '@synchro-charts/core';
+import { DashboardConfiguration } from '../../types';
+
+export const addXAnnotation = ({
+  dashboardConfiguration,
+  widgetId,
+  annotation,
+}: {
+  dashboardConfiguration: DashboardConfiguration;
+  widgetId: string;
+  annotation: XAnnotation;
+}) => {
+  return dashboardConfiguration.widgets.map((widget) => {
+    if (widget.id === widgetId) {
+      const currAnnotations = widget.annotations || {};
+      const currXAnnotations = currAnnotations.x || [];
+      return { ...widget, annotations: { ...currAnnotations, x: [...currXAnnotations, annotation] } };
+    }
+    return widget;
+  });
+};
+
+export const deleteXAnnotation = ({
+  dashboardConfiguration,
+  widgetId,
+  annotationIdToDelete,
+}: {
+  dashboardConfiguration: DashboardConfiguration;
+  widgetId: string;
+  annotationIdToDelete: string;
+}) => {
+  return dashboardConfiguration.widgets.map((widget) => {
+    if (widget.id !== widgetId) {
+      return widget;
+    }
+    if (widget.annotations && widget.annotations.x) {
+      widget.annotations = {
+        ...widget.annotations,
+        x: widget.annotations.x.filter((annotation) => annotation.id != annotationIdToDelete),
+      };
+    }
+    return {
+      ...widget,
+      annotations: widget.annotations,
+    };
+  });
+};
+
+export const editXAnnotation = ({
+  dashboardConfiguration,
+  widgetId,
+  newAnnotation,
+}: {
+  dashboardConfiguration: DashboardConfiguration;
+  widgetId: string;
+  newAnnotation: XAnnotation;
+}) => {
+  return dashboardConfiguration.widgets.map((widget) => {
+    if (widget.id !== widgetId) {
+      return widget;
+    }
+    if (widget.annotations && widget.annotations.x) {
+      widget.annotations.x = widget.annotations.x.filter((annotation) => annotation.id != newAnnotation.id);
+      widget.annotations = { ...widget.annotations, x: [...widget.annotations.x, newAnnotation] };
+    }
+    return {
+      ...widget,
+      annotations: widget.annotations,
+    };
+  });
+};
--- a/packages/dashboard/src/components/iot-annotations/iot-annotations.tsx
+++ b/packages/dashboard/src/components/iot-annotations/iot-annotations.tsx
@@ -1,0 +1,260 @@
+import { Component, State, Prop, h, Watch } from '@stencil/core';
+import { DashboardState, DashboardStore } from '../../types';
+import { onUpdateAction } from '../../dashboard-actions/actions';
+import { XAnnotation } from '@synchro-charts/core';
+import { addXAnnotation, editXAnnotation, deleteXAnnotation } from './annotations';
+import { viewportEndDate, viewportStartDate } from '@iot-app-kit/core';
+
+const DEFAULT_LIST_VISIBLE_ANNOTATIONS = false;
+const DEFAULT_DATE = new Date(0);
+
+@Component({
+  tag: 'iot-annotations',
+})
+export class IotAnnotations {
+  /** Holds all necessary information about dashboard */
+  @Prop() state: DashboardState;
+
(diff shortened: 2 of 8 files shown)


diehbria:
not going to review this until the checks are passing

packages/dashboard/src/index.html
```diff
@@ -10,9 +10,9 @@
     <script type="module" src="/build/iot-app-kit-dashboard.esm.js"></script>
     
 
-    <script type="module" src="http://localhost:3334/build/iot-app-kit-components.esm.js"></script>
-    <link rel="stylesheet" href="http://localhost:3334/build/iot-app-kit-components.css">
-    <script nomodule src="http://localhost:3334/build/iot-app-kit-components.js"></script>
+    <script type="module" src="http://localhost:3333/build/iot-app-kit-components.esm.js"></script>
```
mpopv, line 13: Is this necessary for some reason? If not, can we remove it?

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
```diff
@@ -1,13 +1,104 @@
...
   tag: 'testing-ground',
   styleUrl: 'testing-ground.css',
 })
 export class TestingGround {
   @State() dashboardConfiguration: DashboardConfiguration = dashboardConfig;
+  @State() selectedWidgetId = this.dashboardConfiguration.widgets[0].id;
+  @State() onlyListVisibleAnnotations = DEFAULT_LIST_VISIBLE_ANNOTATIONS;
+
+  onListVisibleAnnotations = (e: Event) => {
+    this.onlyListVisibleAnnotations = (e as any).target.checked;
```
mpopv, line 20: Need to remove all `(e as any)` in added lines before merge

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
```diff
@@ -20,6 +111,79 @@ export class TestingGround {
...
+                <input type="submit" value="Add" id="addAnnotation" onSubmit={this.addAnnotation}></input>
+              </form>
+            </div>
+            <br />
+            <br />
+            <div>
+              <form onSubmit={this.deleteAnnotation}>
+                <label htmlFor="deleteAnnotation"> Delete annotation:</label>
+                <select id="deleteAnnotation">
+                  {this.onlyListVisibleAnnotations
```
mpopv, line 148: This is a lot of logic to be stuffing into a single JSX expression - Can we move this up at least to above the render function's return, and make it a little clearer by refactoring to successive const declarations rather than one big nested ternary?

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
```diff
@@ -20,6 +111,79 @@ export class TestingGround {
...
+            <br />
+            <div>
+              <form onSubmit={this.editAnnotation}>
+                <label htmlFor="editAnnotation"> Edit annotation - </label>
+                <label htmlFor="addLabel"> Label:</label>
+                <input type="text" id="addLabel"></input>
+                <input type="color" value="#FF0000"></input>
+                <br />
+                <select id="editAnnotation">
+                  {this.onlyListVisibleAnnotations
```
mpopv, line 172: Same here. We are doing a lot of stuff here that's duplicative of the above ternary as well, so better to abstract these out above the JSX.

mpopv (changes requested):

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
mpopv: Also if you look at why checks are failing, it's because ESLint is finding too many warnings like `Warning:   111:44  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any`, so I think removing these will results in checks passing.

packages/dashboard/src/index.html
raymfang: From testing with these changes, I don't think these lines are necessary. I've removed them.

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
raymfang: Removed all cases of any!

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
raymfang: Refactored; moved filter and find logic above the render function

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
raymfang: Refactored; moved filter and find logic above the render function

mpopv (approved):

mpopv (commented):
Thanks for making those changes. Great work Raymond!

mpopv (approved):

packages/dashboard/src/components/iot-annotations/annotations.ts
```diff
@@ -0,0 +1,71 @@
...
+export const addXAnnotation = ({
+  dashboardConfiguration,
+  widgetId,
+  annotation,
+}: {
+  dashboardConfiguration: DashboardConfiguration;
+  widgetId: string;
+  annotation: XAnnotation;
+}) => {
+  return dashboardConfiguration.widgets.map((widget) => {
```
diehbria, line 13: rather than making an action per field, for this field we can just utilize 

```https://github.com/awslabs/iot-app-kit/blob/feature-dashboard/packages/dashboard/src/dashboard-actions/actions.ts#L132-L135```

and skip the action boilerplate

**Comment by diehbria** [comment]
can you add a screenshot of what this looks like?

**Review by diehbria — CHANGES_REQUESTED** [review]

**Comment by diehbria** [comment]
This requires some major architectural  changes - changes, we should almost _never_ alter a prop passed into a component. Changes to the dashboard need to be done thru the reducer - there is an `UPDATE` action you will emit, which will then update the dashboard configuration which flows back down

**Review by diehbria — CHANGES_REQUESTED** [review]
This requires some major architectural  changes - we should almost _never_ alter a prop passed into a component. Changes to the dashboard need to be done thru the reducer - there is an `UPDATE` action you will emit, which will then update the dashboard configuration which flows back down.

In general, for both components and 'just functions' inputs should not be mutated (prop being considered an input into a component)

File: packages/dashboard/src/components/iot-annotations/iot-annotations.tsx [hunk 6]
```diff
@@ -0,0 +1,260 @@
...
+
+const DEFAULT_LIST_VISIBLE_ANNOTATIONS = false;
+const DEFAULT_DATE = new Date(0);
+
+@Component({
+  tag: 'iot-annotations',
+})
+export class IotAnnotations {
+  /** Holds all necessary information about dashboard */
+  @Prop() state: DashboardState;
```
diehbria, line 16: there should be no `dashboard state` stored within IotAnnotations

diehbria (changes requested):
still not quite getting the overall architecture right, can meet up and explain in more detail

TheEvilDev:

PR #201, opened 2022-08-10
fix(dashboard): Fix bug with undo move/resize

sydopy:
## Overview
In a prior PR, I removed the 'widgetId' fields from the move and resize actions. This became a significant problem when trying to undo/redo these actions. The undo/redo would succeed when the widget was still selected, however it would fail if they were not selected. I added back the widgetId field so they can be undoable even when not selected.

## Tests

## Legal

### DIFF
--- a/packages/dashboard/src/dashboard-actions/redo.spec.ts
+++ b/packages/dashboard/src/dashboard-actions/redo.spec.ts
@@ -16,6 +16,7 @@ import { reverseMove } from './reverse-actions/reverseMove';
 import { resize } from './resize';
 import { createWidget } from './createWidget';
 import { deleteWidgets } from './delete';
+import { dashboardReducer } from './dashboardReducer';
 
 const state: DashboardState = {
   dashboardConfiguration: {
@@ -31,6 +32,7 @@ const state: DashboardState = {
   undoQueue: [],
   redoQueue: [],
   intermediateDashboardConfiguration: undefined,
+  previousPosition: undefined,
 };
 
 describe('MOVE', () => {
@@ -38,6 +40,7 @@ describe('MOVE', () => {
     position: { x: 10, y: 10 },
     prevPosition: { x: 20, y: 20 },
     isActionComplete: true,
+    widgetIds: [MOCK_KPI_WIDGET.id],
   });
 
   it('returns no change when provided empty dashboard', () => {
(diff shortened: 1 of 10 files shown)


diehbria:
is there a missing test that should be present that would have failed due to this bug?

packages/dashboard/src/dashboard-actions/redo.spec.ts
```diff
@@ -228,3 +231,56 @@ describe('UPDATE', () => {
...
+  it('redoes a resize action and create action', () => {
+    let redoState = dashboardReducer(
+      { ...state },
+      onResizeAction({
+        isActionComplete: true,
+        anchor: 'bottom-right',
+        changeInPosition: { x: 20, y: 20 },
+      })
+    );
+    redoState = dashboardReducer({ ...redoState }, onCreateAction({ widgets: [MOCK_LINE_CHART_WIDGET] }));
```
diehbria, line 271: rather than altering the reference, just make a new reference - variables are cheap! avoid mutations

packages/dashboard/src/dashboard-actions/redo.spec.ts
```diff
@@ -228,3 +231,56 @@ describe('UPDATE', () => {
...
+        dashboardState: redo({
+          dashboardState: { ...state },
+          dashboardAction: onCreateAction({ widgets: [MOCK_LINE_CHART_WIDGET] }),
+        }),
+      }).dashboardConfiguration.widgets
+    ).toEqual(
+      dashboardReducer({ ...redoState }, onCreateAction({ widgets: [MOCK_LINE_CHART_WIDGET] })).dashboardConfiguration
+        .widgets
+    );
+  });
```
diehbria, line 261: line breaks

packages/dashboard/src/dashboard-actions/reverse-actions/reverseMove.ts
```diff
@@ -1,13 +1,14 @@
 import { MoveAction } from '../../dashboard-actions/actions';
 
 export const reverseMove = (moveAction: MoveAction): MoveAction => {
-  if (typeof moveAction.payload.prevPosition != 'undefined') {
+  if (typeof moveAction.payload.prevPosition != 'undefined' && typeof moveAction.payload.widgetIds != 'undefined') {
```
diehbria, line 4: why not just say `moveAction.payload.prevPosition != null` instead of `typeof`? more idiomatic way or doing this

diehbria (changes requested):

packages/dashboard/src/dashboard-actions/reverse-actions/reverseMove.ts
sydopy: i tried that and i got an error of sorts, but now it's fixed

diehbria (approved):

PR #202, opened 2022-08-11
chore(dashboard): rename dashboard components and lift store up one level

mpopv:
_This PR has been revised to rename components and move functionality from testing-ground to the renamed iot-dashboard. Edits from the revision are italic or strikethrough below._

The store needs to be accessible by iot-annotations and other children of ~~testing-ground~~ _the new iot-dashboard_, so we need to lift it up from iot-dashboard into ~~testing-ground~~ _the new iot-dashboard_.

Dashboard widget add/delete/move/copy/paste and undo/redo all work as expected.

~~Some changes were required in the iot-dashboard integration tests; namely, we need to switch to using testing-ground and slightly modify the test results to account for the extra areas around iot-dashboard that are in testing-ground. If this requires modifications I would prefer to add them in a follow-up PR so that we can unblock work relying on this change.~~

I haven't looked too deeply into patterns to consume Redux state in Stencil, but my guess is that prop drilling will get annoying fast. Here we only need to drill 1 level, so I'd like to punt on going deep into the Redux weeds on this, but I'd imagine something like a Provider or [mapStateToProps](https://eliteionic.com/tutorials/state-management-with-redux-in-ionic-stenciljs-loading-data/#consuming-state-in-your-components) will be necessary soon.

## Overview
1. Renames iot-dashboard to iot-dashboard-grid, then adds a new iot-dashboard component with much of the contents of testing-ground.
2. Moves store up from iot-dashboard-grid to iot-dashboard so it can be accessed by other children of iot-dashboard.

## Next steps
1. Update time point annotations PR to hook into the now accessible store, and merge.
2. Simplify how we are using Redux (see my double comment below).
3. Move handle trim widget position into the reducer (see Brian's comment below). 

## Legal


diehbria (changes requested):
I think we should go ahead and take what you have in `testing-ground` and call it `iot-dashboard`, take the current `iot-dashboard` and call it `iot-dashboard-grid`, and take the durrect `iot-dashboard-internal` and call it `iot-dashboard-grid-internal`

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
```diff
@@ -9,6 +17,42 @@ import { dashboardConfig } from '../mocks';
...
+    redoQueue: [],
+    previousPosition: undefined,
+  };
+
+  /**
+   * Callback that is fired every time the dashboard configuration has been altered.
+   *
+   * When a widget is moved, resized, deleted, appended, or altered, then this method is called
+   */
+  onDashboardConfigurationChange = (config: DashboardConfiguration) => {
```
diehbria, line 40: with this 'new iteration' of testing ground, we should instead take a property `onDashboardConfigurationChange` and simply pass down to the child component that expects it

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
```diff
@@ -9,6 +17,42 @@ import { dashboardConfig } from '../mocks';
...
+  };
+
+  store: DashboardStore;
+
+  componentWillLoad() {
+    this.state.dashboardConfiguration = this.dashboardConfiguration;
+    this.store = createStore(dashboardReducer, this.state);
+    this.store.subscribe(() => {
+      this.state = this.store.getState();
+      this.state.dashboardConfiguration.widgets = this.state.dashboardConfiguration.widgets.map(trimWidgetPosition);
```
diehbria, line 51: unrelated to this your commit, but we should handle the trim widget position instead within the reducer (i don't expect you to fix this in this PR, more of a general note on the codebase)

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
mpopv: I'm a little confused by this request. Can you expand on this a bit? What exactly would this accomplish, and how does this replace internal parts of iot-dashboard-grid / iot-dashboard-grid-internal / iot-annotations dispatching to the store?

Right now (despite the descriptive comment above the method, which is not accurate) this is only being called on initial load of the component, and updates are made by this.store.dispatch() calls. I have not changed anything about this, merely moved it up one component level.

mpopv: I'm not sure what you're asking here exactly.

onDashboardConfigurationChange is a function that we only call inside store.subscribe that merely copies the updated Redux state's dashboardConfiguration into this Stencil component's local dashboardConfiguration state. We only want to call it within store.subscribe, assuming we always want the dashboardConfiguration we are passing to the UI to reflect what it is in Redux.

Are you saying child components of testing-ground should each be doing store.subscribe independently, rather than testing-ground being subscribed to Redux and passing state down? In that case, each child should have its own version of onDashboardConfigurationChange that updates its own internal copy of the Redux state; it doesn't make sense to have them both update their parent's copy of the Redux state when the Redux state changes - we'd be updating it twice on each change.

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
mpopv: Agreed, it's odd to have it here.

mpopv:
> I think we should go ahead and take what you have in `testing-ground` and call it `iot-dashboard`, take the current `iot-dashboard` and call it `iot-dashboard-grid`, and take the durrect `iot-dashboard-internal` and call it `iot-dashboard-grid-internal`

I strongly agree with this, but after trying to do this on this branch, I think it should be a follow-up task. It's a very large change that touches many things and would make the specific, minimally unblocking change I am trying to accomplish here (allowing components outside iot-dashboard to access the Redux store) very difficult to review due to the sheer size of the diff.

Would that be alright?

packages/dashboard/src/testing/testing-ground/testing-ground.tsx
mpopv: In general, what's happening with state in iot-dashboard currently needs a rethink. We are storing dashboardConfiguration in two different places (the dashboardConfiguration prop and state.dashboardConfiguration), and we are overwriting the dashboardConfiguration prop (!) with the Redux state's dashboardConfiguration when Redux updates.

This is one reason I really don't think we should get too deep into the weeds on trying to fix state on this branch, beyond minimally moving up what needs to move up -- it needs a larger rethink.

diehbria:
failing tests, will hold off on reviewing until tests pass

PR #209, opened 2022-08-15
feat(context-menu): add context menu to dashboard

jmbuss:
## Overview
Adds a right click context menu to the dashboard. The contents of which are defined by `getContextMenuConfiguration` in `packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx`. The positioning of the menu to handled through the popperjs library. This ensures that the popper will stay on screen no matter where you right click. 

This also introduces a message overrides configuration for the dashboard. Right now this includes the text for the context menu and fallback messaging for a dashboard widget that has failed to load.

## Tests

## Legal


packages/dashboard/src/dashboard-actions/dashboardReducer.ts
```diff
@@ -30,6 +32,22 @@ export const dashboardReducer: Reducer<DashboardState, DashboardAction> = (
...
   let actionToRevert: DashboardAction | undefined = onPasteAction();
   switch (action.type) {
+    case 'BRING_TO_FRONT':
+      return {
+        ...state,
+        dashboardConfiguration: bringToFront({
+          dashboardConfiguration: state.dashboardConfiguration,
+          widgetIds: action.payload.widgets.map((w) => w.id),
+        }),
+      };
```
diehbria, line 42: line breaks between cases please

packages/dashboard/src/integration/iot-dashboard-context-menu.spec.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    dashboardConfiguration: {
+      viewport: { duration: '5m' },
+      widgets: [createWidget()],
+    },
+  });
+
+  // Select the grid area of the dashboard.
+  cy.get('iot-dashboard-internal').children().first().rightclick();
+
+  cy.get('iot-context-menu').should('be.visible');
```
diehbria, line 27: i think were going to want at least one screenshot of it to catch weird style/build regressions

packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
```diff
@@ -419,6 +558,21 @@ export class IotDashboardInternal {
             }}
           ></div>
         )}
+
+        {this.contextMenuOpen && this.contextMenuPosition && (
+          <iot-context-menu {...this.contextMenuPosition}>
+            {this.getContextMenuConfiguration().map((section) => (
```
diehbria, line 564: is there someway we could encapsulate all of this, including the construction of the configuration object into it's own component? i.e. `iot-dashboard-context-menu`

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
```diff
@@ -0,0 +1,47 @@
...
+  invalidTagHeader: string;
+  invalidTagSubheader: string;
+};
+
+export type DashboardMessages = {
+  widgets: WidgetsMessages;
+  contextMenu: ContextMenuMessages;
+};
+
+export type DashboardMessageOverrides = {
```
diehbria, line 19: should this also be a partial?

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
```diff
@@ -0,0 +1,47 @@
...
+  contextMenu: {
+    copyText: 'Copy',
+    pasteText: 'Paste',
+    deleteText: 'Delete',
+    bringToFrontText: 'Bring to Front',
+    sendToBackText: 'Send to Back',
+  },
+};
+
+export const IsMacLike = /(Mac|iPhone|iPod|iPad)/i.test(navigator.platform);
```
diehbria, line 38: first letter capitalized should be reserved for class names and typescript types (i.e. rename to `isMacLike`)

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const mod = IsMacLike ? '⌘' : 'Ctrl';
+export const KeyboardShortcuts = {
```
diehbria, line 41: see comment about about capitalization

diehbria (changes requested):

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
jmbuss: I added the `DashboardMessages` type so that the default messages wouldn't be partial. It made extracting the message easier since we can just do something like -

`Object.assign(
      this.messageOverrides?.contextMenu ?? {},
      DefaultDashboardMessages.contextMenu
    );`
    
    Otherwise we'd have to add another check on the default even though we know it's defined.

packages/dashboard/src/integration/iot-dashboard-context-menu.spec.component.ts
jmbuss: 👍 great point. added.

packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
jmbuss: Ya, I was going back and forth on creating a component to house all that. I think you're right, will make it more maintainable.

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
jmbuss: 👍

packages/dashboard/src/dashboard-actions/dashboardReducer.ts
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-context-menu-option.tsx
```diff
@@ -0,0 +1,27 @@
...
+  tag: 'iot-context-menu-option',
+  styleUrl: 'iot-context-menu-option.css',
+  shadow: false,
+})
+export class IotContextMenuOption {
+  @Prop() disabled: boolean;
+  @Prop() onClick?: (event: MouseEvent) => void;
+
+  render() {
+    const handleClick = (event: MouseEvent) => {
```
diehbria, line 13: should be a method on the class rather than creating a new handler on every render

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-dashboard-context-menu.tsx
```diff
@@ -0,0 +1,159 @@
...
+      DefaultDashboardMessages.contextMenu
+    );
+
+    const copyAction = withClose(() => this.copyWidgets());
+    const pasteAction = withClose(() => this.pasteWidgets(this.contextMenuPosition));
+    const deleteAction = withClose(() => this.deleteWidgets());
+    const bringToFrontAction = withClose(() => this.bringToFront());
+    const sendToBackAction = withClose(() => this.sendToBack());
+
+    const copyMenuOption = {
```
diehbria, line 104: can we remove all this from the class? We typically want to separate out the configuration from the implementation. i.e. we can just make a `contextMenuOptions.ts` file and import them in.

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-dashboard-context-menu.tsx
```diff
@@ -0,0 +1,159 @@
...
+    ];
+  };
+
+  render() {
+    if (this.contextMenuOpen && this.contextMenuPosition) {
+      return (
+        <iot-context-menu {...this.contextMenuPosition}>
+          {this.getContextMenuConfiguration().map((section) => (
+            <iot-context-menu-section>
+              {section.map(({ text, hotkey, action, disabled }) => (
```
diehbria, line 146: I think we should add a key here, just in case `section` in the future could ever change based on other conditions. i.e. the list needs to update.

check out https://stenciljs.com/docs/templating-jsx section "Loops"

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
diehbria: this forces a user of the dashboard to specify each 'section' even if they don't want to, i.e. i must write:

```
<iot-dashboard messageOverrides={{ widgets: {}, contextMenu: { copyText: 'custom msg' } }} ... />
```

This would also make additional base keys of the message overrides a breaking change.

diehbria (changes requested):
see comments

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
jmbuss: 👍 I see your point, updated using RecursivePartial per our discussion!

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-context-menu-option.tsx
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-dashboard-context-menu.tsx
jmbuss: Ya good catch, updated

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-dashboard-context-menu.tsx
jmbuss: I see, yes moved the structuring of these options to it's own module.

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-dashboard-context-menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+    });
+  };
+
+  render() {
+    if (this.contextMenuOpen && this.contextMenuPosition) {
+      return (
+        <iot-context-menu {...this.contextMenuPosition}>
+          {this.getContextMenuConfiguration().map(({ id, options }) => (
+            <iot-context-menu-section key={id}>
+              {options.map(({ id, text, hotkey, action, disabled }) => (
```
diehbria, line 103: has shadow variable `id` (should be caught by eslint, but that's another story). should use a variable alias i.e. `id: optionId`

packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
```diff
@@ -419,6 +456,17 @@ export class IotDashboardInternal {
             }}
           ></div>
         )}
+
+        <iot-dashboard-context-menu
+          messageOverrides={this.messageOverrides}
+          copyWidgets={() => this.onCopy()}
```
diehbria, line 462: is there a reason not to just pass in the funcdtions? i.e.

```copyWidgets={this.onCopy}```

rather than all these wrappers.

Also this made me think - we should probably just keep the naming consistent. i.e. `onDelete` rather than turning `onDelete => deleteWidgets`

File: packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts [hunk 12]
```diff
@@ -0,0 +1,42 @@
+export type ContextMenuMessages = {
```
diehbria, line 1: get rid of the `iot-dashboard-messages` folder - we aren't naming a bunch of folders `iot-NAME` because we want the iot prefix everywhere - it's the pattern for declaring a component, if there's many files and subcomponents, to put it all in a folder that matches the name of the component. In this case I think `messages` should just be moved out of the `components` folder hierarchy completely. The entire package is regarding dashboard, i think it could just go in `src/messages.ts`

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
```diff
@@ -0,0 +1,42 @@
...
+  contextMenu: {
+    copyText: 'Copy',
+    pasteText: 'Paste',
+    deleteText: 'Delete',
+    bringToFrontText: 'Bring to Front',
+    sendToBackText: 'Send to Back',
+  },
+};
+
+export const isMacLike = /(Mac|iPhone|iPod|iPad)/i.test(navigator.platform);
```
diehbria, line 33: this is something a bit seperate from the `messages`, maybe just throw this in some sort of 'keyboard' util file

diehbria (changes requested):

packages/dashboard/src/components/iot-dashboard/iot-context-menu/contextMenuOptions.ts
```diff
@@ -0,0 +1,113 @@
...
+  return {
+    id: 'bringToFront',
+    text: messages.bringToFrontText,
+    hotkey: keyboardShortcuts.bringToFront,
+    disabled: !state.hasSelectedWidgets,
+    action: actions.bringToFrontAction,
+  };
+};
+
+const createSendToBackAction: ActionCreator = (
```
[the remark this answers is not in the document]
jmbuss: A context menu option has functions (what I called action) and messaging which are passed down from the dashboard component. It's not possible to have a static context menu configuration. The purpose of this module is to define the context menu configuration and provide a way to create an instance of it inside the component.

packages/dashboard/cypress/support/commands.js
```diff
@@ -25,3 +25,10 @@
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
 
 import '@4tw/cypress-drag-drop';
+import { addMatchImageSnapshotCommand } from 'cypress-image-snapshot/command';
+
+addMatchImageSnapshotCommand({
+  failureThreshold: 0.025,
+  failureThresholdType: 'percent',
+  customDiffConfig: { threshold: 0.2 },
```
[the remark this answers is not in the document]
jmbuss: We can change this, but I was using the configuration defined in the components package as the precedent here.

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-context-menu-option.css
```diff
@@ -0,0 +1,19 @@
+.iot-context-menu-option {
+  color: var(--selection-bg-color);
+  background-color: var(--colors-dark-grey);
+  font-weight: 400;
+  font-size: 12px;
```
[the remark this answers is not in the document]
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-context-menu-option.css
```diff
@@ -0,0 +1,19 @@
...
+  padding: 0.5rem 1rem;
```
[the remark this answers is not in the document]
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-context-menu-section.css
```diff
@@ -0,0 +1,8 @@
+.iot-context-menu-section {
+  padding: 0.5rem 0;
+  border-bottom: 0.5px solid var(--colors-grey);
```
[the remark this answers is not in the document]
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-dashboard-context-menu.tsx
jmbuss: yup good catch. not sure how the linter didn't catch that

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-dashboard-messages/messages.ts
jmbuss: 👍

packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
jmbuss: ya I suppose we can keep the naming the same. will update and just pass down the function

packages/dashboard/src/components/iot-dashboard/iot-context-menu/iot-context-menu.css
```diff
@@ -0,0 +1,12 @@
+@import "../../variables.css";
+
+.iot-context-menu-placement {
+  position: absolute;
+}
+
+.iot-context-menu {
+  min-width: 200px;
+  border-radius: 3px;
```
diehbria, line 9: there should be a border radius variable in `variables.css`

packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
```diff
@@ -419,6 +456,19 @@ export class IotDashboardInternal {
             }}
           ></div>
         )}
+
+        <iot-dashboard-context-menu
+          messageOverrides={this.messageOverrides}
+          actions={{
+            onCopy: this.onCopy.bind(this),
+            onPaste: this.onPaste.bind(this),
```
diehbria, line 464: skip the binding and just define the methods as arrow functions so `this` is always lexical

packages/dashboard/src/testing/renderContextMenu.tsx
```diff
@@ -0,0 +1,38 @@
...
+          <iot-dashboard-context-menu
+            messageOverrides={DefaultDashboardMessages}
+            actions={{
+              onCopy: noop(),
+              onPaste: noop(),
+              onDelete: noop(),
+              onSendToBack: noop(),
+              onBringToFront: noop(),
+            }}
+            hasCopiedWidgets={true}
```
diehbria, line 31: prefer shorthand i.e.

`hasCopiedWidgets` is the same as `hasCopiedWIdgets={true}`

diehbria (changes requested):

diehbria (approved):

PR #210, opened 2022-08-19
fix(dashboard): fix selection bug 

jmbuss:
## Overview
This PR addresses 3 bugs.
1. The selection box not disappearing on completing the selection action. This was because the gesture was not reactive.
2. Clicking on an element selects all elements that overlap that pixel (not the top most element). To fix this we can sort by the z index and select the highest one in the selection.
3. Paste at position with a selection pastes all the widgets at that position (not the selection at that position). To fix this we can offset the copy group by the difference of paste position.

## Tests

## Legal

### DIFF
--- a/packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
+++ b/packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
@@ -1,4 +1,6 @@
-import { Component, h, Listen, Prop } from '@stencil/core';
+import { Component, h, Listen, Prop, State } from '@stencil/core';
+import sortBy from 'lodash/sortBy';
+import last from 'lodash/last';
 import { Position, Rect, DashboardConfiguration, OnResize, Anchor, Widget, MouseClick } from '../../types';
 import {
   MoveActionInput,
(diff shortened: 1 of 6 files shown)


diehbria:
it seems like theres some missing tests since there were no tests failing prior for the previous behavior but none added

diehbria (changes requested):
add relevant tests

packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
```diff
@@ -125,12 +127,19 @@ export class IotDashboardInternal {
...
   };
+
   onSendToBack = () => {
     this.sendToBack({
       widgets: this.dashboardConfiguration.widgets.filter(({ id }) => this.selectedWidgetIds.includes(id)),
     });
   };
 
+  onClearSelection = () => {
+    this.selectWidgets({
```
boweihan, line 138: Are we planning to have integration tests for this type of functionality?

jmbuss: ahh good catch, I missed those tests. will add.

diehbria (approved):

PR #211, opened 2022-08-23
Add messageOverrrides for table.

square-li:
## Overview
User can override propertyFiltering labels and loading texts by providing messageOverrrides object to the table.

## Tests

## Legal

### DIFF
--- a/packages/components/src/components/iot-table/iot-table.tsx
+++ b/packages/components/src/components/iot-table/iot-table.tsx
@@ -1,5 +1,5 @@
 import { Component, Prop, h, State, Listen, Watch } from '@stencil/core';
-import { Annotations, MessageOverrides, TableColumn, Trend, getThresholds } from '@synchro-charts/core';
+import { Annotations, TableColumn, Trend, getThresholds } from '@synchro-charts/core';
 import {
   StyleSettingsMap,
   TimeSeriesDataRequestSettings,
@@ -11,7 +11,15 @@ import {
   ProviderWithViewport,
 } from '@iot-app-kit/core';
 import { v4 as uuidv4 } from 'uuid';
-import { createTableItems, Item, TableProps } from '@iot-app-kit/table';
+import {
+  createTableItems,
+  Item,
+  TableProps,
+  DefaultTableMessages,
+  TableMessages,
+  RecursivePartial,
+} from '@iot-app-kit/table';
+import merge from 'lodash/merge';
 
 @Component({
   tag: 'iot-table',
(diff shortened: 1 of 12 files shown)


diehbria:
check out how @jmbuss did messageOverides within https://github.com/awslabs/iot-app-kit/pull/209/files

We don't want a bunch of hard coded strings scattered around - instead have a `MessageOverrides` with all the fields set, that deep merges into it. We want this to be consistent across our components

diehbria (changes requested):
check out how @jmbuss did messageOverides within https://github.com/awslabs/iot-app-kit/pull/209/files

We don't want a bunch of hard coded strings scattered around - instead have a MessageOverrides with all the fields set, that deep merges into it. We want this to be consistent across our components

square-li:
Moved all hard coded strings to one single file `packages/table/src/utils/messages.ts`.  Re-requesting for review.

square-li:
Moved all messageOverride together at `packages/components/src/components/iot-table/iot-table.tsx` and deep merge at the entry with default messages strings.

Screen Shot 2022-08-25 at 10 28 13 AM

packages/table/src/utils/messages.ts
```diff
@@ -0,0 +1,56 @@
...
+export type TableCellMessages = {
+  loading: string;
+};
+
+export type TableMessages = {
+  propertyFilter: PropertyFilterMessages;
+  tableCell: TableCellMessages;
+};
+
+export type PartialTableMessages = Partial<{
```
diehbria, line 13: instead, use `RecursivePartial`, and we can skip ever declaring this, and just directly use `RecursivePartial<TableMessages>` as the typescript type on the prop

diehbria (changes requested):

packages/table/src/utils/messages.ts
square-li: Fixed it.

packages/components/src/components/iot-table/iot-table.tsx
```diff
@@ -11,7 +11,9 @@ import {
   ProviderWithViewport,
 } from '@iot-app-kit/core';
 import { v4 as uuidv4 } from 'uuid';
-import { createTableItems, Item, TableProps } from '@iot-app-kit/table';
+import { createTableItems, Item, TableProps, DefaultTableMessages, TableMessages } from '@iot-app-kit/table';
+import merge from 'lodash/merge';
+import { RecursivePartial } from '../../testing/types';
```
diehbria, line 16: move `RecursivePartial` to `utils/types` or some other location that makes more sense. shouldn't use testing stuff in implementing components

diehbria (changes requested):

diehbria (approved):

PR #223, opened 2022-08-30
fix(dashboard): add touch support for context menu

jmbuss:
## Overview
This adds support for the context menu on a touch device. The trigger is a long press (500ms) on the dashboard grid area. The implementation uses a custom decorator to register the event listeners.

## Tests

## Legal

### DIFF
--- a/packages/dashboard/src/components/iot-dashboard/getDashboardPosition.ts
+++ b/packages/dashboard/src/components/iot-dashboard/getDashboardPosition.ts
@@ -2,9 +2,30 @@ import { Position } from '../../types';
 
 export const DASHBOARD_CONTAINER_ID = 'container';
 
-export const getDashboardPosition = (event: MouseEvent): Position => {
-  let totalOffsetX = event.offsetX;
-  let totalOffsetY = event.offsetY;
+const getOffsets = (event: MouseEvent | TouchEvent | PointerEvent) => {
+  if (event instanceof TouchEvent) {
+    const ev = event.touches[0];
+
+    const rect = (ev.target as HTMLElement).getBoundingClientRect();
+    const offsetX = ev.clientX - rect.x;
+    const offsetY = ev.clientY - rect.y;
+
+    return {
+      offsetX,
+      offsetY,
+    };
+  }
+
+  return {
+    offsetX: event.offsetX,
+    offsetY: event.offsetY,
+  };
+};
+
+export const getDashboardPosition = (event: MouseEvent | TouchEvent | PointerEvent): Position => {
+  const { offsetX, offsetY } = getOffsets(event);
+  let totalOffsetX = offsetX;
+  let totalOffsetY = offsetY;
   let currElement: HTMLElement | null = event.target as HTMLElement;
 
   while (currElement && currElement.id != DASHBOARD_CONTAINER_ID) {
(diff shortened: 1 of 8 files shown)


packages/dashboard/src/components/iot-dashboard/getDashboardPosition.ts
```diff
@@ -2,9 +2,30 @@ import { Position } from '../../types';
 
 export const DASHBOARD_CONTAINER_ID = 'container';
 
-export const getDashboardPosition = (event: MouseEvent): Position => {
-  let totalOffsetX = event.offsetX;
-  let totalOffsetY = event.offsetY;
+const getOffsets = (event: MouseEvent | TouchEvent | PointerEvent) => {
+  if ((event as TouchEvent).touches) {
```
diehbria, line 6: is there a way to utilize narrowing (https://www.typescriptlang.org/docs/handbook/2/narrowing.html) to achieve this without asserting `as`?

diehbria (changes requested):

packages/dashboard/src/decorators/press.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const defaultOptions: Required<PressOptions> = {
+  duration: 500,
+  threshold: 5,
+  trigger: 'touch',
+  target: 'self',
+};
+
+type EventTypess = {
+  start: string;
```
diehbria, line 24: are these actually any string? or is there a strict set of values they can be

packages/dashboard/src/decorators/press.ts
```diff
@@ -0,0 +1,128 @@
+import { ComponentInterface } from '@stencil/core';
+import { noop } from 'lodash';
+import { Position } from '../types';
+import { distance } from '../util/distance';
+
+export type PressEvents = 'press';
+
+type PressTrigger = 'mouse' | 'touch' | 'all';
+export type PressOptions = {
+  duration?: number;
```
diehbria, line 10: can you document the units expected? duration in MS?

packages/dashboard/src/decorators/press.ts
```diff
@@ -0,0 +1,128 @@
...
+      x: ev.clientX,
+      y: ev.clientY,
+    };
+    pressTimeout = setTimeout(() => {
+      cb.call(component, event);
+      reset();
+    }, duration);
+  };
+  const pressEndHandler = (event: Event) => {
+    if (startPosition === null || event.type === 'mouseup' || event.type === 'pointerup') {
```
diehbria, line 93: a common bug i see is where people handle null explicitly, and then eventually something changes and an `undefined` somehow gets passed down. we can just use `== null` and `!= null` instead and always just handle null and undefined - unless there is a really good reason otherwise.

jmbuss: The typing on start position is defined as `let startPosition: Position | null = null;`. Unless someone overrides an assignment with an any or something similar this shouldn't be a problem? I can change the check to see if the variable is falsey to guard against that case though.

packages/dashboard/src/components/iot-dashboard/getDashboardPosition.ts
jmbuss: yup! good callout.

diehbria (approved):

PR #225, opened 2022-08-31
feat(core)!: Refactor time series data module / add meta field to data source

diehbria:
## Overview

- Add `meta` field to dataStream type to support sources providing meta information into data sources. This will help support alarm feature within the package `@iot-app-kit/source-twinmaker` which is a work in progress. Related commit in Synchro Charts: https://github.com/awslabs/synchro-charts/commit/e3ff0e10c14b45bdd4affffb08bcdd34b39f2aaf

- BREAKING CHANGE: Refactored export from `@iot-app-kit/core` IoTAppKitDataModule to be named TimeSeriesDataModule, and removed the concept of multiple data sources per time series data module

- Remove the `name` fields in queries internally, since it is no longer necessary as each time series data module only maps to one data source, as opposed to a single repository of all time series data sources.

- Upgrade Synchro Charts to latest version, 6.0.1

## Legal

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -54,7 +54,7 @@
     "@iot-app-kit/related-table": "^1.5.0",
     "@iot-app-kit/source-iotsitewise": "^1.5.0",
     "@stencil/core": "^2.7.0",
-    "@synchro-charts/core": "^6.0.0",
+    "@synchro-charts/core": "^6.0.1",
     "styled-components": "^5.3.0"
   },
   "devDependencies": {
--- a/packages/core/src/__mocks__/data-source.ts
+++ b/packages/core/src/__mocks__/data-source.ts
@@ -17,7 +17,6 @@ export const createMockSiteWiseDataSource = (
     onRequestData?: (props: any) => void;
   } = { dataStreams: [], onRequestData: () => {} }
 ): DataSource<SiteWiseDataStreamQuery> => ({
-  name: 'site-wise',
   initiateRequest: jest.fn(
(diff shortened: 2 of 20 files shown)


sheilaXu:
the core package has another DataStream type here https://github.com/awslabs/iot-app-kit/blob/83527d6c91a38c22011626ac23d74aa3330603af/packages/core/src/data-module/types.ts#L32 should we add meta to it? or remove it and use the one from synchro charts everywhere?

boweihan (dismissed):
Looks great. Will we also need to explicitly expose this `meta` property on our existing components (iot-line-chart) or is it only going to be used by new components for now?

diehbria:
> Looks great. Will we also need to explicitly expose this `meta` property on our existing components (iot-line-chart) or is it only going to be used by new components for now?

Line chart and many other widgets in their current form, would entirely ignore it, but the idea is the datastreams with the meta would flow to them. In the future, if we need 'data source specific features' we can leverage them if the features aren't easily representable in some more generic way that isn't data source specific

diehbria:
> the core package has another DataStream type here
> 
> https://github.com/awslabs/iot-app-kit/blob/83527d6c91a38c22011626ac23d74aa3330603af/packages/core/src/data-module/types.ts#L32
> 
> should we add meta to it? or remove it and use the one from synchro charts everywhere?

I had intended to append the meta field type here.. not sure what went wrong.

Revised the code and added it.

boweihan (approved):

Issue #227, opened 2022-09-01
iot-resource-explorer jest test failures locally


diehbria:
**Describe the bug**
`iot-resource-explorer` jest tests failure, but only locally

```
@iot-app-kit/components: FAIL src/components/iot-resource-explorer/iot-resource-explorer.spec.ts
@iot-app-kit/components:   ● Test suite failed to run
@iot-app-kit/components:     Call retries were exceeded
@iot-app-kit/components:       at ChildProcessWorker.initialize (../../node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)
```

**To Reproduce**
1. run `yarn run test` from the workspace root

**Expected behavior**
All tests pass locally

**Desktop (please complete the following information):**
 - OS: macOS
 - node version: 16.17.0

PR #203, opened 2022-08-11
Npm workspaces

TheEvilDev:
## Overview
Removes dev dependencies on `yarn` and `lerna` and just uses npm workspaces.

## Tests
[Include a link to the passing GitHub action running the test suite here.]

## Legal
This project is available under the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0.html).

PR #231, opened 2022-09-02
build: npm workspaces

boweihan:
## Overview
1. Migrate to NPM workspaces - taking over https://github.com/awslabs/iot-app-kit/pull/203
2. Fix typescript errors in component tests
3. Rewrite resource-explorer tests for performance

## Legal


.github/workflows/npm-publish.yml
```diff
@@ -18,9 +18,9 @@ jobs:
 
       - name: Install and Build
         run: |
-          yarn install --frozen-lockfile
```
diehbria, line 21: is there a reason to remove this?

diehbria:
it looks like it doesn't have lock files for all the packages?

diehbria:
what was the fix for https://github.com/awslabs/iot-app-kit/issues/227 ?

NorbertNader (dismissed):

.npmrc
```diff
@@ -1,2 +1,3 @@
 CYPRESS_INSTALL_BINARY=0
 puppeteer_skip_chromium_download=true
+legacy-peer-deps=true
```
NorbertNader, line 3: Why do we need this?

NorbertNader, line 3: Why do we need this?

boweihan: I thought we needed this to manage peer dependency conflicts / backwards compatibility because we're now forcing people to use npm 8+...but it looks like the build works fine without it. Removing.

boweihan:
> what was the fix for #227 ?



> it looks like it doesn't have lock files for all the packages?

Good catch - the lockfile in core doesn't need to exist

boweihan:
> what was the fix for #227 ?

There was a type mismatch when passing props to the related table react component (no type-safety there). You could find the error by de-obfuscating the react error through the dev console (running the test in dev mode).

boweihan:
I rewrote the iot-resource-explorer cypress tests - hopefully this resolves #227

diehbria (approved):

PR #237, opened 2022-09-08
feat(dashboard): add drag and drop event handlers

jmbuss:
## Overview
Adds event handling for drag and drop interactions

Refactor internal dashboard to use those events instead of mouse events. This makes internal dashboard work on touch devices.
Refactor resizable panels to use drag and drop events.

## Legal

### DIFF
--- a/packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
+++ b/packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
@@ -15,6 +15,8 @@ import { getSelectedWidgetIds, getSelectedWidgets } from '../../util/select';
 import { getSelectionBox } from './getSelectionBox';
 import { DASHBOARD_CONTAINER_ID, getDashboardPosition } from './getDashboardPosition';
 import { DashboardMessages } from '../../messages';
+import { CustomEvent } from '../../decorators/events';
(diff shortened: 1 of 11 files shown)


packages/dashboard/src/components/iot-dashboard/iot-dashboard-internal.tsx
```diff
@@ -146,20 +166,86 @@ export class IotDashboardInternal {
...
+  onPressStart(event: PointerEvent) {
+    if (event.button !== MouseClick.Left) return;
+
+    const isUnionSelection = event.shiftKey;
+
+    const { x, y } = getDashboardPosition(event);
+    const isOnSelection = this.isPositionOnSelection({ x, y });
+
+    // context menu uses a click handler, so mousedown will fire before we can cancel the event.
+    const isAgainstContextMenuTarget =
```
diehbria, line 200: is it possible for the dashboard not to know about every single possible thing that the events are trying to achieve, and instead have something else just prevent the event from propagating? I.e. the resize box could capture the event and prevent onPressStart from ever initiating

jmbuss: Ya, I agree, this has taken on too much responsibility. Will refactor it

PR #241, opened 2022-09-13
fix: only last dataStream rendered on all charts

NorbertNader:
## Overview
This fixes an issue in the time series store that would render the last dataStream on all charts even if other properties were queried. This was caused by how lodash.merge resolves arrays.

Screen Shot 2022-09-14 at 12 30 24 AM

Simplified alarms docs.

## Legal


diehbria (changes requested):
can we add a test that would fail with the prior code?

diehbria (approved):

PR #252, opened 2022-09-23
feat(TwinMaker): Add TwinMaker packages

sheilaXu:
## Overview
Add TwinMaker packages

## Legal


TheEvilDev (dismissed):
A CR on a change like this is kind of unrealistic. Any feedbacks would need to be done as follow ups, and all of this code has gone through a review at one point or another already.

Ship it, and we can address any changes as follow-ups or just refactor as we go.

.eslintrc.js
```diff
@@ -4,8 +4,16 @@ module.exports = {
...
     jest: true,
+    node: true,
   },
-  ignorePatterns: ['cypress', 'stencil.config.ts', 'configuration', '__mocks__'],
+  ignorePatterns: [
+    'cypress',
+    'stencil.config.ts',
+    'configuration',
+    '__mocks__',
+    'packages/scene-composer/src/three/tiles3d/*',
```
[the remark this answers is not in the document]
sheilaXu: this is to ignore them from lint check, not test coverage. it's easier to match to original code when we don't change the part of the code we don't need to change

TheEvilDev (approved):

diehbria:
Add twinmaker references to the root README.md to make it discoverable

docs/SceneViewer.md
```diff
@@ -0,0 +1,134 @@
...
+
+The SceneViewer component contains the following properties that you can customize. 
+
+### `sceneLoader`
+
+The class to load scene metadata and content.
+
+**Note: When a new instance of this object is passed in, the SceneView will trigger a new loading of the whole scene. Therefore, do not recreate this object when not needed.**
+
+- Type: `SceneLoader` defined in `@iot-app-kit/source-iottwinmaker`
```
diehbria, line 34: remove list indentations for `Type` statements across docs, make consistent with existing documents

docs/SceneViewer.md
```diff
@@ -0,0 +1,134 @@
...
+
+- `dracoDecoder`
+
+  The configurations for a draco decoder. More information can be found [here](https://docs.aws.amazon.com/iot-twinmaker/latest/guide/scenes-before-starting.html)
+
+  - Type: Object
+  
+- `locale`
+
+  The language for the texts displayed in the component. Default to `en-US`.
```
diehbria, line 129: is it optional? if it has a default, it sounds like it's optional

docs/VideoPlayer.md
```diff
@@ -0,0 +1,102 @@
+# Video Player
+
+The VideoPlayer component allows you to stream a video from the Kinesis Video Streams. It supports the following configuration:
```
diehbria, line 3: is there more detail we could provide? My understanding is it's more than _just_ a video player, but more like a playlist?

examples/react-app/src/components/SceneViewer.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const onSelectionChanged = useCallback((e) => {
+    console.log('onSelectionChanged event fired with data: ', e)
+  }, []);
+
+  const onWidgetClick = useCallback((e) => {
+    console.log('onWidgetClick event fired with data: ', e)
+  }, []);
+
+  return (
+    <div className="SceneViewer" >
```
diehbria, line 25: is there a reason the Scenviewer component requires a div wrapper? why not just have it built into the component exported from scene composer package?

package.json
```diff
@@ -16,7 +16,9 @@
       "packages/related-table",
       "packages/table",
       "packages/components",
-      "packages/react-components"
+      "packages/source-iottwinmaker",
+      "packages/react-components",
+      "packages/scene-composer"
```
diehbria, line 21: should scene composer component come before the other component packages? The list of packages is sensitive as it determines the build order

packages/react-components/src/video-player/utils/dateRangeUtils.tsx
```diff
@@ -0,0 +1,46 @@
...
+) => {
+  let startAndEndTime = undefined;
+  if ('relative' === newDateRange.type) {
+    let offset = 0;
+    switch (newDateRange.unit) {
+      case 'second':
+        offset = newDateRange.amount;
+        break;
+      case 'minute':
+        offset = newDateRange.amount * 60;
```
diehbria, line 15: can we utilize constants like we do in the other packages, i.e. `MINUTE_IN_MS`, `DAY_IN_MS`, etc.

packages/react-components/src/video-player/utils/videoProgressUtils.tsx
```diff
@@ -0,0 +1,22 @@
...
+  endTimeStamp: number
+) => {
+  return ((currentStart + currentTime * 1000 - startTimeStamp) / (endTimeStamp - startTimeStamp)) * 100;
+};
+
+export const getVideoProgressSeekTime = (percentage: number, startTimeStamp: number, endTimeStamp: number) => {
+  return startTimeStamp + (percentage * (endTimeStamp - startTimeStamp)) / 100;
+};
+
+export const getVideoProgressTooltip = (seekTime: number, startTime: number) => {
```
diehbria, line 16: declare typescript return types

packages/scene-composer/src/common/errors.ts
```diff
@@ -0,0 +1,52 @@
...
+  constructor(
+    level: ErrorLevel,
+    code: ErrorCode,
+    message: string,
+    context?: Record<string, string>,
+    innerError?: unknown,
+  ) {
+    super(`[${level}][${code}] ${message}`);
+    this.details = {
+      level: level,
```
diehbria, line 45: prefer object shorthand i.e. `{ level, code, ... }`

packages/scene-composer/src/components/SceneComposerInternal.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <SceneComposerInternal
+          config={{ mode: 'Editing' }}
+          sceneLoader={createSceneLoaderMock(invalidTestScenes.invalidSpecVersionScene)}
+        />,
+      );
+    });
+
+    await new Promise((resolve) => setTimeout(resolve, 1));
+
+    expect(container).toMatchSnapshot();
```
diehbria, line 125: instead of a bunch of snapshorts, can we instead assert that there is the error we expect?

packages/scene-composer/src/components/SceneComposerInternal.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+    act(() => {
+      container = renderer.create(
+        <SceneComposerInternal
+          config={{ mode: 'Editing' }}
+          sceneLoader={createSceneLoaderMock(invalidTestScenes.invalidSpecVersionScene)}
+        />,
+      );
+    });
+
+    await new Promise((resolve) => setTimeout(resolve, 1));
```
diehbria, line 123: we can instead do something like `await flushPromises()`  https://www.npmjs.com/package/flush-promises as we do in the other packages

packages/scene-composer/src/components/SceneComposerInternal.spec.tsx
```diff
@@ -0,0 +1,220 @@
+import * as React from 'react';
```
diehbria, line 1: overall these tests heavily rely on snapshots, which are not very good tests to give people certainty that changes are correct, as they have a tendancy to fail due to _any_ changes, and are often blindly updated.

Please instead assert on the expected output, hopefully in a way that they only break if the thing the test is trying to assert on, is actually wrong

diehbria (changes requested):

package.json
sheilaXu: I don't see any build issues... the composer also has source-iottwinmaker as a dependency, so feel it could be the last

.stylelintignore
```diff
@@ -4,3 +4,4 @@ www/
 loader/
 node_modules/
 coverage/
+examples/
```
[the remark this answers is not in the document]
sheilaXu: i will check how many changes needed to include it as followup

NorbertNader (changes requested):
needs version bump before publishing

NorbertNader (dismissed):

package.json
diehbria: Shouldn't `scene-composer` be a dependency of `react-components`? and why does `react-components` rly on `source-twnmaker`?

sheilaXu: I believe `source-twinmaker` is used in the video player test to create the video data object.
how do we want to export composer from react-components? i can do it as followup. like this in its index file?
```
export * from '@iot-app-kit/scene-composer';
```

diehbria (approved):

examples/react-app/src/components/SceneViewer.tsx
sheilaXu: by default, Scenviewer takes the full screen. this div provides a style to it to make it fit in it's spot in the sample app

sheilaXu:
> Add twinmaker references to the root README.md to make it discoverable

addressed in https://github.com/awslabs/iot-app-kit/pull/256

docs/SceneViewer.md
sheilaXu: addressed in https://github.com/awslabs/iot-app-kit/pull/256

docs/SceneViewer.md
sheilaXu: marked as optional in https://github.com/awslabs/iot-app-kit/pull/256

docs/VideoPlayer.md
mukeshsahay: I think the details added are sufficient and clear for the users to understand the purpose and usage of the video player component.
From user point of view it is regular video player that can play from KVS using KVS stream name or using TwinMaker Entity and Component information.
Internally, in the TwinMaker mode we fetch the KVS stream name and other information like when the video is available, when not available and when available on edge. And then accordingly setup the video player with sources. I think we should not include these information in the user guide for the component. 

I'm open to recommendations though - Please suggest.

packages/scene-composer/src/assets/svgs/icons/OrbitCameraIcon.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+import { Orbit } from '../../auto-gen/icons';
+
+export const OrbitCameraSvg = (
+  <span>
```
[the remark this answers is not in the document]
sheilaXu: to prevent the auto set styles from awsui

packages/scene-composer/src/assets/svgs/icons/PanCameraIcon.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+import { Pan } from '../../auto-gen/icons';
+
+export const PanCameraSvg = (
+  <span>
```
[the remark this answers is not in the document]
sheilaXu: to prevent the auto set styles from awsui

packages/scene-composer/src/components/DefaultErrorFallback/__tests__/__snapshots__/DefaultErrorFallback.spec.tsx.snap
```diff
@@ -0,0 +1,93 @@
...
+    <Styled(Component)
+      header={
+        <Unknown
+          variant="h2"
+        >
+          Error
+        </Unknown>
+      }
+    >
+      <Unknown>
```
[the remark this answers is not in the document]
sheilaXu: fixed in https://github.com/awslabs/iot-app-kit/pull/278

packages/scene-composer/src/components/SceneComposerInternal.spec.tsx
sheilaXu: addressed in https://github.com/awslabs/iot-app-kit/pull/278

packages/scene-composer/src/augmentations/three/Anchor.ts
```diff
@@ -0,0 +1,73 @@
...
+ * The generic anchor object.
+ */
+export class Anchor extends THREE.Object3D {
+  private log = new DebugLogger('Anchor');
+
+  protected visualMap = new Map<string, IVisual>();
+  protected _visualState: DefaultAnchorStatus | string = DefaultAnchorStatus.Info;
+  protected _isSelected = false;
+
+  // @ts-ignore
```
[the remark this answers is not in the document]
sheilaXu: removed in https://github.com/awslabs/iot-app-kit/pull/278

packages/scene-composer/src/common/errors.ts
sheilaXu: addressed in https://github.com/awslabs/iot-app-kit/pull/278

packages/scene-composer/src/common/errors.ts
```diff
@@ -0,0 +1,52 @@
...
+  message: string;
+  context?: Record<string, string>;
+}
+
+// TODO: Handle I18N error message
+export class SceneComposerRuntimeError extends CustomError {
+  public details: IErrorDetails;
+  public innerError?: unknown;
+
+  constructor(
```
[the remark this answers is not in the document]
sheilaXu: addressed in https://github.com/awslabs/iot-app-kit/pull/278

packages/scene-composer/src/components/SceneComposerInternal.spec.tsx
sheilaXu: We've been trying to mock the child components that will change for unrelated changed, and they are working fine so far. so we can keep them like this for now and switch to assert on the expected output once we see issues

packages/scene-composer/src/components/SceneComposerInternal.spec.tsx
sheilaXu: We've been trying to mock the child components that will change for unrelated changed, and they are working fine so far. so we can keep them like this for now and switch to assert on the expected output once we see issues

packages/scene-composer/src/common/constants.ts
```diff
@@ -0,0 +1,134 @@
...
+import { Component, LightType } from '../models/SceneModels';
+import { InfoIconSvgString, WarningIconSvgString, ErrorIconSvgString, VideoIconSvgString } from '../assets';
+import { IValueDataBindingProviderState, DefaultAnchorStatus, DistanceUnit, Vector3 } from '../interfaces';
+import { CameraControlImpl } from '../store/internalInterfaces';
+
+/******************************************************************************
+ * Document Constants
+ ******************************************************************************/
+
+// TODO: [4553] remove in preview build
```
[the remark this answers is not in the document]
sheilaXu: I don't quite remember when is the preview build... but it doesn't hurt to keep it for a longer time, maybe after this year's reinvent so we don't break things now..

PR #266, opened 2022-10-05
Fix: Refactor to improve ViewCursorWidget performance

mumanity:
## Overview
ViewCursorWidget Refactor
* Fixes performance
* Fixes tag placement
* Removes unused code from previous implementation. 

## Legal


packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.spec.tsx
```diff
@@ -2,15 +2,14 @@
 import * as THREE from 'three';
 import React from 'react';
 import renderer from 'react-test-renderer';
+import { act } from '@testing-library/react';
 import { useLoader, useThree } from '@react-three/fiber';
+import { MockTransformControls } from '../../../../../tests/__mocks__/MockTransformControls';
 
-import { ViewCursorWidget } from '../../../../../src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget';
+import { ViewCursorWidget } from '../viewpoint/ViewCursorWidget';
 import { Viewpoint } from '../../../../../src';
```
jwills-jdubs, line 10: All references to Viewpoint can now be removed. The View Cursor was the only reason it wasn't removed when it's feature was removed.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
```diff
@@ -1,76 +1,94 @@
-import * as THREE from 'three';
-import React, { useContext, useEffect, useMemo, useRef } from 'react';
-import { extend, useLoader, useThree } from '@react-three/fiber';
-import { SVGLoader, SVGResult } from 'three-stdlib';
+import React, { useCallback, useContext, useMemo, useRef } from 'react';
+import { useFrame, useLoader, useThree } from '@react-three/fiber';
+import { AlwaysDepth, Box3, Color, DoubleSide, Group, Mesh, MeshBasicMaterial, Object3D, ShapeGeometry, Vector3 } from 'three';
```
jwills-jdubs, line 3: I'm going to make a non-blocking suggestion here:
I strongly advise that we use import * as THREE from 'three'; and I'll explain my reasoning as to why.

1. There are now possible name collisions. In our specific case 3 libraries including ours, have their own version of Vector3 (React-Three/Fiber, Ours in interfaces I believe, and Three)
2. The namespace is a ThreeJS standard. The author, all documentation and examples using ThreeJS will do the namespace style

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
```diff
@@ -1,76 +1,94 @@
...
-    >
-      {moveVisual}
-      {editVisual}
-    </viewCursor>
-  );
-};
+  /* istanbul ignore next */
+  useFrame(({ raycaster }) => {
+    const sceneMeshes: Object3D[] = [];
+    scene.traverse((child) => {
```
jwills-jdubs, line 77: if this is the only place scene is being used you can get this from the state passed into useFrame such as { raycaster, scene }

jwills-jdubs (changes requested):

packages/scene-composer/src/assets/auto-gen/icons/Camera.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { SVGProps } from 'react';
 
 const SvgCamera = (props: SVGProps<SVGSVGElement>) => (
+  /* istanbul ignore next */
```
jwills-jdubs, line 5: Looks like these are causing problems in the build or tests :/

packages/scene-composer/src/assets/auto-gen/icons/Light.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { SVGProps } from 'react';
```
TheEvilDev, line 2: These are auto-generated, tehy shouldn't be included in the CR. Check your `.gitignore` file.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
TheEvilDev: While I see where @jwills-jdubs is coming from, I respectfully disagree. `import * as Whatever from 'anylib';` is generally considered poor practice in Javascript, as it forces you to include entire libraries in the bundling process or relying on Tree Shaking to break things up.

The name collision issue can easily be worked around like:

```
import { AlwaysDepth as ThreeAlwaysDepth } from 'three';
```

Should never import more than you need.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
```diff
@@ -1,76 +1,94 @@
...
-    return svgIconToWidgetVisual(svgData, ViewpointState.Deselected, false, { userData: { isCursor: true } });
-  }, []);
+  const svg = cursorStyle === 'edit' ? ViewCursorEditIcon : ViewCursorMoveIcon;
+  const data = useLoader(SVGLoader, svg.dataUri);
 
-  const editVisual = useMemo(() => {
-    const svgData: SVGResult = useLoader(SVGLoader, ViewCursorEditIcon.dataUri);
-    return svgIconToWidgetVisual(svgData, ViewpointState.Selected, false, { userData: { isCursor: true } });
-  }, []);
+  /* istanbul ignore next */
```
TheEvilDev, line 20: Are we not testing this part because it's hard, or because it's unnecessary? If it's hard, and we don't have time for it, I'd prefer we actually just drop coverage rather than creating a false sense of security.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
```diff
@@ -1,76 +1,94 @@
...
+          side: DoubleSide
+        });
+        const shapes = SVGLoader.createShapes(path);
+        shapes.forEach((line) => {
+          const geometry = new ShapeGeometry(line);
+          const mesh = new Mesh(geometry, material);
+          resetObjectCenter(mesh);
+          svgGroup.add(mesh);
+        });
+      }
```
TheEvilDev, line 49: This logic still feels out of place here, moving to utility functions that break up some of this complexity could make both this Component and this logic easier to test.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
```diff
@@ -1,76 +1,94 @@
...
+          depthFunc: AlwaysDepth,
+          side: DoubleSide
+        });
+        path.subPaths.forEach((childPath) => {
+          const geometry = SVGLoader.pointsToStroke(childPath.getPoints(), path?.userData?.style);
+          if (geometry) {
+            const mesh = new Mesh(geometry, material);
+            resetObjectCenter(mesh);
+            svgGroup.add(mesh);
           }
```
TheEvilDev, line 66: Ditto, consider refactoring this into smaller more testable pieces.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
TheEvilDev: We have to get the raycaster on frame? This seems unfortunate. 

There are 2 instances where we need to do the collision detection here:
1. When the user clicks the mouse, so we know what to attach things to
2. When the mouse moves over an object (so we can adjust the rotation of the cursor)

For the second case, I can see `useFrame` might actually be more efficient, since the mouse can definitely move faster than 60 times/second, but this creates possible lag as the orientation of the cursor can take a few milliseconds to catch up to the actual mouse position. The problem is while we target 60 frames per second, slower machines with weaker graphics cards may not be able to render at 60 fps, and the aforementioned lag is even more pronounced in those situations.

I'd actually prefer for the second scenario, if we added a debounce on the mouse move to make the lag at least consistent.

packages/scene-composer/src/components/WebGLCanvasManager.tsx
```diff
@@ -95,6 +71,15 @@ export const WebGLCanvasManager: React.FC = () => {
...
   };
 
+  const esc = useCallback(() => {
+    window.addEventListener('keyup', (e: KeyboardEvent) => {
+      if (e.key === 'Escape' && !!addingWidget) {
+        setAddingWidget(undefined);
+      }
+    });
+    return window?.removeEventListener('keyup', setAddingWidget as any);
+  }, [addingWidget])
```
TheEvilDev, line 81: I'm not sure I agree with adding key handling to the window, but this is not blocking since we're only doing it while the editing state is "active".

Keep in mind, in both the appkit and grafana case, we do not own the page, and attaching anything to window risks polluting the global namespace of the host application. We should instead add this key handler to like the canvas or the overall viewer node.

packages/scene-composer/src/components/WebGLCanvasManager.tsx
```diff
@@ -104,6 +89,9 @@ export const WebGLCanvasManager: React.FC = () => {
   }, [gridHelperRef.current]);
 
   useEffect(() => {
+    setCursorVisible(!!addingWidget);
+    setCursorStyle(addingWidget ? 'edit' : 'move');
```
TheEvilDev, line 93: This should probably be handled in state, and not in the WebGLCanvasManager.

packages/scene-composer/src/components/three-fiber/EntityGroup/index.tsx
```diff
@@ -111,7 +115,7 @@ const EntityGroup = ({ node }: IEntityGroupProps) => {
         dispose={null}
         onPointerDown={onPointerDown}
         onPointerUp={onPointerUp}
-        userData={{ nodeRef: !isEnvironmentNode(node) ? nodeRef : undefined, componentTypes }} // Do not add ref for environment nodes
+        userData={{ nodeRef: !isEnvironmentNode(node) ? nodeRef : undefined, targetRef, componentTypes }} // Do not add ref for environment nodes
```
TheEvilDev, line 118: What is the "targetRef"?

packages/scene-composer/src/components/three-fiber/ModelRefComponent/GLTFModelComponent.tsx
```diff
@@ -174,19 +160,16 @@ export const GLTFModelComponent: React.FC<GLTFModelProps> = ({
     if (addingWidget) {
       const parent = findNearestViableParentAncestorNodeRef(e.object) || clonedModelScene;
       const { position } = getIntersectionTransform(e.intersections[0]);
-      const newWidgetNode = createNodeWithPositionAndNormal(addingWidget, position, cursorLookAt, parent);
-
+      const targetParent = getObject3DBySceneNodeRef(parent.userData.targetRef);
```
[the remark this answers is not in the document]
TheEvilDev: Why is this needed?

packages/scene-composer/src/utils/nodeUtils.ts
```diff
@@ -31,7 +31,7 @@ export const findComponentByType = (
  * @returns {boolean} true if the object is a Viable Parent
  */
 const isViableParent = (object: THREE.Object3D): boolean => {
-  return !!object.userData.nodeRef;
+  return !!object.userData.targetRef;
```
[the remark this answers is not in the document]
TheEvilDev: Yeah I don't agree with this design, we should revisit. Flagging models with metadata about what scene node they are makes sense because it creates a reverse mapping so we can get objects by node, or node by objects.

This is a hack that alters this behavior to solve the problem for a single feature.

TheEvilDev (changes requested):
We need to reconsider the `targetParent` workaround here. If we are going to hack this for one feature it should just be a new property that that one feature can ignore some nodes in it's calculation.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.spec.tsx
mumanity: Update coming in next commit. Please sanity check removals.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
mumanity: @TheEvilDev You good with this? I implemented the initial feedback since it made sense, but @jwills-jdubs makes a really good point about name collisions.

packages/scene-composer/src/assets/auto-gen/icons/Camera.tsx
mumanity: Fixed.

packages/scene-composer/src/components/three-fiber/EntityGroup/index.tsx
mumanity: Since submodels are separate scenes we need to track where the original ref was (hence `targetRef`). When we look at the parent of a submodel its parent is the root of its separate scene & that was what caused the odd placement of the tags when clicking on submodels. By passing an optional `targetRef`, we're able to accurately tell the tag where to be placed in the world.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
jwills-jdubs: So when this gets packaged we only take what we need from the lib? If we really don't take it all anyway it makes sense to me.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
mumanity: `resetObjectCenter` is used to keep an `svg` centered to the mouse pointer so it would be more appropriately tested with end-to-end tests. Unit test coverage for this would be useless IMO. However, if you don't agree we can discuss. 

The point you made below about moving the `shape` rendering into utils makes sense & would be more testable in that aspect.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
mumanity: Great point.

packages/scene-composer/src/components/WebGLCanvasManager.tsx
mumanity: I've been trying to figure out how to get it off the `window`, but haven't had luck. Will continue exploring this, but not promising that it will be fixed in the upcoming commit.

packages/scene-composer/src/components/WebGLCanvasManager.tsx
mumanity: Good call out.

packages/scene-composer/src/components/toolbars/floatingToolbar/items/AddObjectMenu.tsx
```diff
@@ -250,7 +248,11 @@ export const AddObjectMenu = () => {
           parentRef: mustBeRoot ? undefined : getRefForParenting(),
         } as unknown as ISceneNodeInternal;
 
-        appendSceneNode(node);
+        if (enhancedEditingEnabled && !modelType) {
```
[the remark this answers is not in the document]
mumanity: What's the blast radius & how would we test that the suggested change works properly?

packages/scene-composer/src/utils/nodeUtils.ts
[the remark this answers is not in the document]
mumanity: Will address.

jwills-jdubs:
> We need to reconsider the `targetParent` workaround here. If we are going to hack this for one feature it should just be a new property that that one feature can ignore some nodes in it's calculation.

They can't be completely ignored. We need to be a child of the submodel but we need to parent to the model in the original scene. The submodels live in a different sub scene which strangely effects world position. I can give more details if you'd like but the metadata tagging is a clean solution and not purely special cased. 

The way it is meant to work is that all known objects have a `nodeRef` for hierarchical parenting (already what drives enhanced editing) and a `targetRef` (maybe needs renaming) for 3D world/local space conversion

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
mumanity: Really good callouts on both sides guys. I'll convert to use @TheEvilDev's suggestion of `import { AlwaysDepth as ThreeAlwaysDepth } from 'three';`

packages/scene-composer/src/components/WebGLCanvasManager.tsx
jwills-jdubs: is it possible to have the canvas listen for the keypress? you can get it from the R3F state 
```
const { gl } = useThree()
gl.domElement.addEventListener('keypress', () => {});
```

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
mumanity: `useFrame` is actually ideal for what we're doing here as this [hook](https://gracious-keller-98ef35.netlify.app/docs/api/hooks/useFrame/) is designed specifically for effects. 

Raycasting based on mouse movement requires having to constantly look at where the mouse is intersecting in order to know how to cast the `svg`. Without it, we can't place the `svg` in the canvas.

If mouse movement wasn't involved then we wouldn't have a need for the `useFrame`.

packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
mumanity: `svg` testing is included in upcoming push.

PR #267, opened 2022-10-06
chore(CameraView): Removes TopBar Feature Flag Dependency for Camera View

jwills-jdubs:
This will improve the following:

1. No longer require a feature flag so no special build of grafana is required once feature ships
2. If there are no cameras we don't show the UI as it is confusing
3. This restores not showing the View Options if there are no Motion Indicators
4. Hides the top bar if neither exist for more screen real estate

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/TopBar.tsx
+++ b/packages/scene-composer/src/components/panels/TopBar.tsx
@@ -3,14 +3,13 @@ import styled from 'styled-components';
 import { ButtonDropdown, SpaceBetween } from '@awsui/components-react';
 import { useIntl } from 'react-intl';
 
-import { COMPOSER_FEATURES, KnownComponentType } from '../../interfaces';
+import { KnownComponentType } from '../../interfaces';
 import { sceneComposerIdContext } from '../../common/sceneComposerIdContext';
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/components/panels/TopBar.tsx
```diff
@@ -59,29 +62,35 @@ export const TopBar: FC = () => {
...
-      </ButtonDropdown>
-      {cameraViewEnabled && (
-        <ButtonDropdown data-testid={'camera-views'} items={cameraItems} onItemClick={setActiveCameraOnItemClick}>
-          {intl.formatMessage({ defaultMessage: 'Cameras', description: 'camera views dropdown button text' })}
-        </ButtonDropdown>
-      )}
-    </StyledSpaceBetween>
-  );
+  if (showTopBar) {
+    return (
```
mumanity, line 66: You could write this as ```{showTopBar && (<JSX_Content />)}``` & it will only populate if `showTopBar` is true. That way you can remove the need to return an empty fragment (`<></>`).

mumanity (approved):

sheilaXu:
update the PR title to follow the same commit message pattern since that will be the commit message when you squash and merge the PR by default (unless you want to update the message during merge)

PR #269, opened 2022-10-06
test(react-components, VideoPlayer): fix flaky test for videoPlayer

square-li:
## Overview
Due to a bug in `viewportStartDate` from `@iot-app-kit/core`, `propertiesNotChanged` might return `false` for unchanged properties. Add `currentTime` param `viewportStartDate` to fix this issue.

### Details
Basically `viewportStartDate` uses `Date.now()` when the viewport passed in is a duration (no `start`). And function `propertiesNotChanged` compares viewports as following:
```
viewportStartDate(this.props.viewport).getTime() === viewportStartDate(prevProps.viewport).getTime() 
```
Usually `viewportStartDate(this.props.viewport).getTime()` and `viewportStartDate(prevProps.viewport).getTime()` should return the same value for the same viewport. However, if there’s some async ops or it is running in a slow machine (like Github actions runner), the return value can be different - since `Date.now()` returns a new timestamp. Then this behaviour leads to `propertiesNotChanged` return false and causes `getKvsStreamSrcFn` called twice instead expected once.

## Legal

### DIFF
--- a/packages/core/src/common/viewport.ts
+++ b/packages/core/src/common/viewport.ts
@@ -2,11 +2,11 @@ import { MinimalViewPortConfig } from '@synchro-charts/core';
 import { isMinimalStaticViewport } from './predicates';
 import { parseDuration } from './time';
 
-export const viewportStartDate = (viewportConfig: MinimalViewPortConfig): Date =>
+export const viewportStartDate = (viewportConfig: MinimalViewPortConfig, currentDate?: Date): Date =>
   isMinimalStaticViewport(viewportConfig)
     ? new Date(viewportConfig.start)
-    : new Date(Date.now() - parseDuration(viewportConfig.duration));
+    : new Date((currentDate?.getTime() || Date.now()) - parseDuration(viewportConfig.duration));
 
-export const viewportEndDate = (viewportConfig: MinimalViewPortConfig): Date => {
-  return isMinimalStaticViewport(viewportConfig) ? new Date(viewportConfig.end) : new Date(Date.now());
+export const viewportEndDate = (viewportConfig: MinimalViewPortConfig, currentDate?: Date): Date => {
+  return isMinimalStaticViewport(viewportConfig) ? new Date(viewportConfig.end) : currentDate || new Date(Date.now());
 };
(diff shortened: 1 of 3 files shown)


packages/react-components/src/video-player/videoPlayer.tsx
```diff
@@ -449,12 +452,12 @@ export class VideoPlayer extends React.Component<IVideoPlayerProps, IVideoPlayer
     }
   };
 
-  private errorHandlingForLiveMode = (errorMessage: string) => {
+  private errorHandlingForLiveMode(errorMessage: string) {
```
sheilaXu, line 455: this change may cause `this` to be lost, better to keep the arrow function to be safe

packages/react-components/src/video-player/videoPlayer.tsx
```diff
@@ -465,7 +468,7 @@ export class VideoPlayer extends React.Component<IVideoPlayerProps, IVideoPlayer
     }
   };
 
-  private setVideoPlayerForLiveMode = async () => {
+  private async setVideoPlayerForLiveMode() {
```
sheilaXu, line 471: same here

packages/react-components/src/video-player/videoPlayer.tsx
square-li: > this change may cause `this` to be lost, better to keep the arrow function to be safe

yes good catch, i was using this to trace call stack. forget to remove it. Thanks! Removed.

sheilaXu (approved):

NorbertNader (dismissed):

packages/core/src/common/viewport.ts
```diff
@@ -2,11 +2,11 @@ import { MinimalViewPortConfig } from '@synchro-charts/core';
 import { isMinimalStaticViewport } from './predicates';
 import { parseDuration } from './time';
 
-export const viewportStartDate = (viewportConfig: MinimalViewPortConfig): Date =>
+export const viewportStartDate = (viewportConfig: MinimalViewPortConfig, currentTime?: Date): Date =>
   isMinimalStaticViewport(viewportConfig)
     ? new Date(viewportConfig.start)
-    : new Date(Date.now() - parseDuration(viewportConfig.duration));
+    : new Date((currentTime?.getTime() || Date.now()) - parseDuration(viewportConfig.duration));
```
boweihan, line 8: currentDate instead of currentTime?

boweihan (changes requested):

packages/core/src/common/viewport.ts
square-li: Right Date is better

boweihan (approved):

PR #276, opened 2022-10-12
fix(composer): use arrow data as texture to avoid public path issue

sheilaXu:
## Overview
use arrow data as texture to avoid public path issue

## Legal


packages/scene-composer/src/components/three-fiber/MotionIndicatorComponent/helpers.ts
```diff
@@ -3,7 +3,15 @@ import { useEffect, useRef } from 'react';
 import * as THREE from 'three';
 
 import { Component, Vector3 } from '../../../models/SceneModels';
-import arrow from '../../../assets/icons/arrow.svg';
+import { getDataUri } from '../../../utils/svgUtils';
+
+const arrowSvg = `
```
mumanity, line 8: Any reason this isn't stored with the other svgs in `assets/svgs`?

mumanity (approved):

PR #309, opened 2022-11-02
Feat(Scene Hierarchy): Radio Button Implementation & Selection Bug Fix 

mumanity:
## Overview
(Fix) Scene Composer:  Implementing radio buttons for single-select
(Fix) Scene Composer: Scene Hierarchy selection on single click

## Legal


packages/scene-composer/src/components/Tree/tree.scss
```diff
@@ -25,145 +25,121 @@ $active-color: #ec7211;
...
+
+          // styled to match Polaris: https://polaris.a2z.com/components/radio-group/
+          &:checked {
+            appearance: none;
+            background-color: #fff;
+            color: #fff;
+            border: 4px solid #00a1c9;
+            border-radius: 50%;
+            padding: 3px;
+          }
```
TheEvilDev, line 67: These should be pulling styles from Polaris directly, as opposed to being hard coded here. I'm ok with shipping this, but we should spend a bit of time seeing if we can use their CSS variables here instead.

TheEvilDev (approved):

packages/scene-composer/src/components/Tree/tree.scss
mumanity: We can create our own component. Unfortunately the Polaris radio button doesn't allow for passing unique identifiers so React throws the "Each child should have a unique key prop" error.

packages/scene-composer/src/components/Tree/TreeItem.tsx
```diff
@@ -23,30 +25,34 @@ export interface TreeItemProps extends TreeItemInnerProps, ComponentPropsWithRef
...
   className = '',
   onActivated = () => {},
 }) => {
+  const sceneComposerId = useSceneComposerId();
+
   const toggle = useCallback(
     (e) => {
-      onSelected(!selected, e);
+      (children as any)?.filter((child) => {
+        return child ? useStore(sceneComposerId).getState().setSelectedSceneNodeRef(child.props?.objectRef) : null;
```
sheilaXu, line 38: use hook inside hook is anti-pattern

sheilaXu (changes requested):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -143,8 +143,7 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
   const select = useCallback(
     (objectRef?: string) => {
       if (sceneComposerId) {
-        const setSelectedSceneNodeRef = useStore(sceneComposerId).getState().setSelectedSceneNodeRef;
-        setSelectedSceneNodeRef(objectRef);
+        useStore(sceneComposerId).getState().setSelectedSceneNodeRef(objectRef);
```
sheilaXu, line 146: this file can be cleaned up a bit.. useStore can be all grouped together..

packages/scene-composer/src/components/Tree/TreeItem.tsx
sheilaXu: nvm, this pattern is introduced recently to fix performance issue. I will look into if we can fix the performance issue from the useStore hook side instead of calling hook inside hook

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
sheilaXu: same here. not an issue for this PR

TheEvilDev:

PR #320, opened 2022-11-04
fix(composer): submodel and hierarchy search fix

sheilaXu:
## Overview
- fix the issue where only root nodes can be searched
- only show the matching nodes in a flat list, not their children
- gate sub-model feature with flag
- show SubModelTree for ModelRef component and fix duplicated trees when sub model is promoted to a node

search issue:

https://user-images.githubusercontent.com/9315598/199861938-7d02f86d-83d8-4029-a0f8-80f2d30256e4.mov

search fix:

https://user-images.githubusercontent.com/9315598/199862110-74d44955-e29c-4945-80eb-0519a830c8e4.mov

sub-model issue:

https://user-images.githubusercontent.com/9315598/199862147-eb260182-0119-4aa0-9c38-20eb6cbf4ba3.mov

sub-model fix:

https://user-images.githubusercontent.com/9315598/199862174-4c73a1c3-f544-435a-a673-688a84b6a7b3.mov

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
+++ b/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
@@ -1,6 +1,7 @@
 import React, { FC, createContext, useContext, useCallback, useState } from 'react';
 import { DndProvider } from 'react-dnd';
 import { HTML5Backend } from 'react-dnd-html5-backend';
+import { isEmpty } from 'lodash';
 
 import { useSceneComposerId } from '../../../common/sceneComposerIdContext';
 import { findComponentByType, isEnvironmentNode } from '../../../utils/nodeUtils';
@@ -57,18 +58,15 @@ export const useSceneHierarchyData = () => {
   return useContext(Context);
 };
 
-const toSceneHeirarchyNode = ({
-  ref,
-  name,
-  parentRef,
-  childRefs = [],
-  components,
-}: ISceneNodeInternal | Readonly<ISceneNodeInternal>) => {
+const toSceneHeirarchyNode = (
+  { ref, name, parentRef, childRefs = [], components }: ISceneNodeInternal | Readonly<ISceneNodeInternal>,
+  hideChild?: boolean,
+) => {
   return {
     objectRef: ref,
     name,
     componentTypes: components.map((c) => c.type),
-    childRefs,
+    childRefs: hideChild ? [] : childRefs,
     parentRef,
   } as ISceneHierarchyNode;
 };
@@ -102,22 +100,24 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
       ? unfilteredNodeMap
       : Object.values(unfilteredNodeMap).filter((node) => searchMatcher(node, searchTerms));
 
-  const rootNodeRefs = Object.values(nodeMap)
+  const unfilteredRootNodeRefs = Object.values(unfilteredNodeMap)
     .filter((item) => !item.parentRef && (!isEnvironmentNode(item) || !isViewing()))
     .map((item) => item.ref);
 
-  const rootNodes: Readonly<ISceneNodeInternal>[] = rootNodeRefs
-    .map(getSceneNodeByRef)
-    .filter((node) => node !== undefined && searchMatcher(node, searchTerms))
-    .map((item) => item as ISceneNodeInternal)
-    .sort(sortNodes);
+  const rootNodes: Readonly<ISceneNodeInternal>[] = !isEmpty(searchTerms)
+    ? Object.values(nodeMap)
+    : unfilteredRootNodeRefs
+        .map(getSceneNodeByRef)
+        .filter((node) => node !== undefined && searchMatcher(node, searchTerms))
+        .map((item) => item as ISceneNodeInternal)
+        .sort(sortNodes);
 
   const getChildNodes = useCallback(
     (parentRef?: string) => {
       const nodeMap = useStore(sceneComposerId).getState().document.nodeMap;
       const results = Object.values(nodeMap)
         .filter((node) => node.parentRef === parentRef)
-        .map(toSceneHeirarchyNode)
+        .map((node) => toSceneHeirarchyNode(node))
         .sort(sortNodes);
 
       return results;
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -57,18 +58,15 @@ export const useSceneHierarchyData = () => {
...
-const toSceneHeirarchyNode = ({
-  ref,
-  name,
-  parentRef,
-  childRefs = [],
-  components,
-}: ISceneNodeInternal | Readonly<ISceneNodeInternal>) => {
+const toSceneHeirarchyNode = (
+  { ref, name, parentRef, childRefs = [], components }: ISceneNodeInternal | Readonly<ISceneNodeInternal>,
+  hideChild?: boolean,
```
TheEvilDev, line 63: I'm not sure this belongs here, this is kind of polluting the object model. Hiding children in search results is a view concern, it shouldn't affect the underlying data model.

Suggestion: We could move the filtering logic to the getChildNodes hook instead, and just make that aware of the search filters or similar.

TheEvilDev: This might be the simplest way to achieve this though, so it's fine.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -102,22 +100,24 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
...
+        .map((item) => item as ISceneNodeInternal)
+        .sort(sortNodes);
 
   const getChildNodes = useCallback(
     (parentRef?: string) => {
       const nodeMap = useStore(sceneComposerId).getState().document.nodeMap;
       const results = Object.values(nodeMap)
         .filter((node) => node.parentRef === parentRef)
-        .map(toSceneHeirarchyNode)
+        .map((node) => toSceneHeirarchyNode(node))
```
TheEvilDev, line 120: This change creates an unnecessary wrapping function. Super minor, but it's technically another pointer the GC needs to clean up.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -196,7 +196,7 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
     <DndProvider backend={HTML5Backend}>
       <Context.Provider
         value={{
-          rootNodes: rootNodes.map(toSceneHeirarchyNode),
+          rootNodes: rootNodes.map((node) => toSceneHeirarchyNode(node, !isEmpty(searchTerms))),
```
TheEvilDev, line 199: Looking at this, I would almost prefer to see a "filteredNodes" property passed through context instead, but we can save that for when we rework search properly.

TheEvilDev (approved):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -47,7 +48,8 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
       (type as unknown as IModelRefComponentInternal)?.modelType !== ModelType.Environment,
   );
 
-  const showSubModel = isValidModelRef && !!model && !isViewing();
+  const [{ variation: subModelSelectionEnabled }] = useFeature(COMPOSER_FEATURES[COMPOSER_FEATURES.SubModelSelection]);
```
mumanity, line 51: Can you just pass `useFeature(COMPOSER_FEATURES.SubModelSelection)`? Curious if the extra nesting is really necessary.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -97,11 +99,9 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
             getChildNodes(key).map((node, index) => (
               <React.Fragment key={index}>
                 <SceneHierarchyTreeItem key={node.objectRef} enableDragAndDrop={enableDragAndDrop} {...node} />
-                {showSubModel && node?.componentTypes?.includes(KnownComponentType.SubModelRef) && (
-                  <SubModelTree parentRef={key} expanded={false} object3D={model!} selectable />
-                )}
               </React.Fragment>
             ))}
+          {showSubModel && <SubModelTree parentRef={key} expanded={false} object3D={model!} selectable />}
```
mumanity, line 104: Just a heads up that my incoming change will move this back into the `React.Fragment` with fixes for hierarchy placement. Non-blocking, but please let me know if you have any concerns with this regarding your fix.

mumanity (approved):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
sheilaXu: we shouldn't put it into the Fragment because this tree is for the root ModelRef node, not for the child SubModelRef nodes. the current code breaks submodel feature because when I create a new scene and add a model to it, I won't be able to see the tree and cannot add sub model ref component to the scene, so this tree never show up

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
sheilaXu: I'm following how useFeature is used for other features, not quite sure why we did this.... probably because this COMPOSER_FEATURES enum used to be int enum not string enum..

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
sheilaXu: the map function auto provide a number as second parameter, so if I don't wrap it, there will be error

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
sheilaXu: a filter callback? that's probably more scalable. i can have a try in my next change

PR #326, opened 2022-11-07
fix(composer):Scene Hierarchy radio buttons & bug fix for selection on single click

mumanity:
## Overview
Implements Scene Hierarchy radio buttons for single-click select & bug fix for selection on single click
- Includes some test cleanup

## Legal


packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -85,7 +85,7 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
       expandable={childRefs.length > 0}
       selected={selected === key}
       selectionMode={selectionMode}
-      onSelected={isViewing() ? onActivated : onToggle}
+      onSelected={onToggle}
```
sheilaXu, line 88: is this the new expected behavior to not move camera on select in viewing mode?

packages/scene-composer/src/components/Tree/TreeItem.tsx
```diff
@@ -23,33 +28,45 @@ export interface TreeItemProps extends TreeItemInnerProps, ComponentPropsWithRef
...
   onSelected = /* istanbul ignore next */ () => {},
   selectable = true,
   selected = false,
   className = '',
   onActivated = () => {},
 }) => {
+  const sceneComposerId = useSceneComposerId();
+  const setSelectedSceneNodeRef = useStore(sceneComposerId).getState().setSelectedSceneNodeRef;
+
+  const selectFromChildren = () => {
```
sheilaXu, line 41: `useCallback`?

packages/scene-composer/src/components/Tree/TreeItem.tsx
```diff
@@ -23,33 +28,45 @@ export interface TreeItemProps extends TreeItemInnerProps, ComponentPropsWithRef
...
+    (children as any)?.filter((child) => {
```
sheilaXu, line 42: should this be `forEach`? doesn't seem to return a filtered result

mumanity: Good call out. Will fix.

packages/scene-composer/src/components/Tree/TreeItem.tsx
mumanity: This is invoked directly from a function in the same component that uses the `useCallback` there shouldn't be any need for one here.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mumanity: As I understand it, the camera should only move to target on a double-click. Single-click should update what is populated in the inspect panel so that the user can maintain view on the item they want. Ex: Double-click to target the fork lift, then single-click the directional light. The user can then modify the light intensity, color, etc... to see the effect it has on the scene.

sheilaXu: this is more for viewing mode, not editing mode. for editing mode, we don't move camera with single click, but the current behavior for viewing mode is that we will move camera to the object when it's single clicked

mumanity: Will revert this change. If someone wants it to behave differently we'll get some documentation for it.

sheilaXu (approved):

PR #327, opened 2022-11-07
fix(Composer): Safe bounding box and 3D cursor fix

jwills-jdubs:
## Overview
Push out for the 3D cursor was causing spinning and movement due to it being executed in useFrame. This is not actually necessary for the cursor due to it's visibility through objects

Helpers, most notably the Camera Helper make the scene too large and trying to focus on a camera as well as the whole scene while active will zoom incorrectly when finding the best viewing angle

Keeps the system cursor visible with 3D cursor as per UX decision for visibility at any scale and distance. The 3D cursor is a nice helper that we can improve upon later. Adding the cursor makes this non-blocking

## Legal


packages/scene-composer/src/utils/objectThreeUtils.ts
```diff
@@ -68,8 +68,34 @@ export function enableShadow(component: IModelRefComponentInternal, obj: THREE.O
...
+  obj.position.multiplyScalar(-1);
+};
+
+export const getSafeBoundingBox = (obj: THREE.Object3D): THREE.Box3 => {
+  // Because LineSegments have absurd sizes in ThreeJS we need to account for these in the scene and ignore them.
+  // Map all Line Segments to their parent for re-parenting
+  const lineMap: Map<THREE.Object3D, THREE.Object3D> = new Map<THREE.Object3D, THREE.Object3D>();
+  obj.traverse((child) => {
+    if ((child as THREE.LineSegments).isLineSegments) {
+      if (child.parent) {
```
mumanity, line 83: No need to nest. This can be expressed as a single `if` check. `if (child as THREE.LineSegments).isLineSegments && child.parent) ...`

packages/scene-composer/src/utils/objectThreeUtils.ts
```diff
@@ -68,8 +68,34 @@ export function enableShadow(component: IModelRefComponentInternal, obj: THREE.O
...
+
+  // Severe the connection
+  lineMap.forEach((line) => {
+    line.removeFromParent();
+  });
+
+  const safeBoundingBox = new THREE.Box3().setFromObject(obj);
+
+  // Re-connect
+  lineMap.forEach((line, parent) => {
```
mumanity, line 97: Same note as above. Looks like you want to have this covered with promise.

mumanity:
Approving as risk is minimal. Will revisit after upcoming deadline.

mumanity (approved):

PR #329, opened 2022-11-07
fix(dashboard): add node polyfill and svg plugin

jmbuss:
## Overview
There are many nodejs apis used in source-iotsitewise and synchro-charts that need to be polyfilled for dashboard to run correctly. Some of these are crypto, os, fs, http, etc. These are included in the `rollup-plugin-polyfill-node` plugin. Also included in this PR is a fix for the grid svg that we use as the background for the dashboard grid component. To include this in the distributable, it is converted to an inline asset using `postcss-url`.

## Legal


packages/dashboard/package.json
```diff
@@ -36,8 +36,8 @@
     "@storybook/addon-essentials": "^6.5.12",
     "@storybook/addon-interactions": "^6.5.12",
     "@storybook/addon-links": "^6.5.12",
-    "@storybook/builder-webpack5": "^6.5.12",
-    "@storybook/manager-webpack5": "^6.5.12",
+    "@storybook/builder-webpack4": "^6.5.13",
```
mpopv, line 39: Should we upgrade other storybook packages to `^6.5.13` as well so they stay in sync?

mpopv (approved):

diehbria (approved):

PR #342, opened 2022-11-09
Opacity filter zero

TheEvilDev:
## Overview
Opacity filter was incorrectly showing objects when opacity was set to 0. This was a bug in the boolean logic to check if it should apply alpha or not.

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/ColorOverlayComponent/index.tsx
+++ b/packages/scene-composer/src/components/three-fiber/ColorOverlayComponent/index.tsx
@@ -45,12 +45,14 @@ const ColorOverlayComponent: React.FC<IColorOverlayComponentProps> = ({
     /* istanbul ignore next */ (obj) => {
       if (obj instanceof Mesh && ruleColor) {
         if ('color' in obj.material) {
-          if (ruleColor?.color) {
-            obj.material.color = ruleColor.color.clone().convertSRGBToLinear();
-          }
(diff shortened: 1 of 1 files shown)


jwills-jdubs (dismissed):

diehbria (changes requested):
make sure the commits follow conventional commit and that the PR is a request to merge one well formatted commit (commit message considered part of the review)

diehbria (approved):

packages/scene-composer/src/components/three-fiber/ColorOverlayComponent/index.tsx
```diff
@@ -45,12 +45,14 @@ const ColorOverlayComponent: React.FC<IColorOverlayComponentProps> = ({
...
-            obj.material.color = ruleColor.color.clone().convertSRGBToLinear();
-          }
-          if (ruleColor?.alpha && ruleColor?.alpha !== 1) {
-            obj.material.transparent = true;
-            obj.material.opacity = ruleColor.alpha;
+          if (ruleColor) {
+            if (ruleColor.color) {
+              obj.material.color = ruleColor.color.clone().convertSRGBToLinear();
+            }
+            if ((ruleColor.alpha || ruleColor.alpha === 0) && ruleColor?.alpha !== 1) {
```
sheilaXu, line 52: would it be more clear to combine the first 2 conditions into `isNumber(ruleColor.alpha)`?

PR #356, opened 2022-11-11
fix(composer): Fixes the duplication of sub models

jwills-jdubs:
## Overview
Sub Models previously relied on `appendSceneNodeInternal` preventing.a duplicate based purely on the reference Id and the Object3D.id of the original sub model. The problem is both ThreeJS and our Ids are not persisted by design. 

The solution was to add a property specifically to sub models that reference the "ID" of the sub model object. The only persisted case here is the name and changing that within the model in something like Blender is still a way to mess this up however I do believe unless you specifically name objects the most 3D applications will provide a new name. We then check all nodes in the scene for the property matching the name. If we find it we do not add the sub model to the scene as was occurring when within the same session.

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SubModelTree/index.tsx
+++ b/packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SubModelTree/index.tsx
@@ -27,7 +27,7 @@ const SubModelTree: FC<SubModelTreeProps> = ({
   visible: defaultVisible = true,
 }) => {
   const sceneComposerId = useSceneComposerId();
-  const { appendSceneNodeInternal } = useSceneDocument(sceneComposerId);
+  const { appendSceneNodeInternal, document } = useSceneDocument(sceneComposerId);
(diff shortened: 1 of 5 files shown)


packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SubModelTree/index.tsx
```diff
@@ -76,12 +85,14 @@ const SubModelTree: FC<SubModelTreeProps> = ({
...
-      properties: [],
+      properties: {
+        subModelId: object3D.name,
+      },
     } as ISceneNodeInternal;
 
     appendSceneNodeInternal(node);
     setSceneNodeObject3DMapping(nodeRef, object3D); // Cache Reference
-  }, [object3D]);
+  }, [object3D, document.nodeMap]);
```
mumanity, line 95: Typically I'd flag this, but approving as this is made available via `useSceneDocument`.

mumanity (approved):

PR #361, opened 2022-11-14
fix(composer): Fix to restore drag-&-drop placement after refresh

mumanity:
## Overview
Fix to restore drag-&-drop placement after refresh

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
+++ b/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
@@ -62,14 +62,12 @@ export const useSceneHierarchyData = () => {
 
 const toSceneHeirarchyNode = (
   { ref, name, parentRef, childRefs = [], components }: ISceneNodeInternal | Readonly<ISceneNodeInternal>,
-  canExpand: boolean,
   hideChild?: boolean,
 ) => {
   return {
     objectRef: ref,
     name,
     componentTypes: components.map((c) => c.type),
-    hasChildren: canExpand,
     childRefs: hideChild ? [] : childRefs,
     parentRef,
   } as ISceneHierarchyNode;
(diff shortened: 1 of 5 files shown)


jwills-jdubs (dismissed):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -105,38 +104,27 @@ const sortNodes = (a, b) => {
...
+    document,
+    selectedSceneNodeRef,
+    getSceneNodeByRef,
+    setSelectedSceneNodeRef,
+    updateSceneNodeInternal,
+    getObject3DBySceneNodeRef,
+    setCameraTarget,
+    removeSceneNode,
+    isEditing,
+  } = useStore(sceneComposerId)((state) => state);
```
sheilaXu, line 117: using individual hooks is a performance improvement from https://github.com/awslabs/iot-app-kit/pull/283/files#diff-9eae39af074be35fac13ff65406838d84bce07db7e48db1e4654f15a227e23d9

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -193,44 +190,41 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
...
-    [sceneComposerId],
+    [removeSceneNode],
   );
 
   return (
     <DndProvider backend={HTML5Backend}>
       <Context.Provider
         value={{
-          rootNodes: rootNodes.map((node) => toSceneHeirarchyNode(node, !isEmpty(searchTerms))),
+          rootNodes: rootNodes.map((item) => toSceneHeirarchyNode(item, item.childRefs.length > 0)),
```
sheilaXu, line 227: the hideChild flag is to hide child of the nodes found from search, so we only display a flat list of matching nodes. this change will break it

sheilaXu (changes requested):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
mumanity: Local testing shows that search is working properly.

sheilaXu: you should only see the nodes matching the search, not their children (unless the children matches to)

mumanity: Current state is broken. This branch still displays children, however, it also restores the ability to search at all.

sheilaXu (approved):

PR #363, opened 2022-11-14
fix(composer): SubModel child fix

jwills-jdubs:
### Overview
Forces all sub model children to be physically parented to it's parent. This makes it so that children don't move with the sub model but the render correctly. There is a feature flag to maintain previous implementations so that we can fix later

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/EntityGroup/index.tsx
+++ b/packages/scene-composer/src/components/three-fiber/EntityGroup/index.tsx
@@ -11,9 +11,10 @@ import {
 } from '../../../store';
 import { sceneComposerIdContext, useSceneComposerId } from '../../../common/sceneComposerIdContext';
 import { getChildrenGroupName, getEntityGroupName } from '../../../utils/objectThreeUtils';
-import { KnownComponentType } from '../../../interfaces';
+import { COMPOSER_FEATURES, KnownComponentType } from '../../../interfaces';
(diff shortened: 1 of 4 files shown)


mumanity (dismissed):

mumanity (approved):

packages/scene-composer/src/components/three-fiber/EntityGroup/index.tsx
```diff
@@ -22,12 +23,35 @@ interface IEntityGroupProps {
...
+    const childRefs: string[] = [];
+    if (node.childRefs) {
+      childRefs.push(...node.childRefs);
+
+      if (subModelChildrenDisabled) {
+        node.childRefs.forEach((childRef) => {
+          const child = getSceneNodeByRef(childRef);
+          const childIsSubModel = !!findComponentByType(child, KnownComponentType.SubModelRef);
+          if (child && childIsSubModel) {
+            childRefs.push(...child.childRefs);
```
jiajzhou, line 42: You only handled 1 level of nesting, can there be multiple levels of nestings? e.g. a GLTF model contains another GLTF model and both having submodels, would this code work?

jwills-jdubs: This is a good call out and I almost did that. However by design sub models will not be able to have sub model children. And because of that as long as they become a child of something in real space a.k.a whatever its parent is by definition we are safe

PR #366, opened 2022-11-15
fix(composer): One instance of sub models allowed per instance of the model

jwills-jdubs:
## Overview
It was my misunderstanding that we needed to limit multiple instances of the sub models on a per ModelRef basis. This fixes that and also removes the unnecessary property

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SubModelTree/index.tsx
+++ b/packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SubModelTree/index.tsx
@@ -7,8 +7,9 @@ import Tree, { TreeItem } from '../../../../Tree';
 import { ISubModelRefComponent, KnownComponentType } from '../../../../../interfaces';
 import { ISceneComponentInternal, ISceneNodeInternal, useEditorState, useSceneDocument } from '../../../../../store';
 import { useSceneComposerId } from '../../../../../common/sceneComposerIdContext';
-import './SubModelTree.scss';
+import { findComponentByType } from '../../../../../utils/nodeUtils';
 
+import './SubModelTree.scss';
 import TreeItemLabel from './SubModelTreeItemLabel';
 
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SubModelTree/index.tsx
```diff
@@ -55,11 +56,16 @@ const SubModelTree: FC<SubModelTreeProps> = ({
...
+    const duplicates = getSceneNodeByRef(parentRef)?.childRefs.filter((childRef) => {
+      const child = getSceneNodeByRef(childRef);
+      const childSubModelComponent = findComponentByType(
+        child,
+        KnownComponentType.SubModelRef,
+      ) as ISubModelRefComponent;
+      return childSubModelComponent?.selector === object3D.name;
+    });
+
+    if (duplicates && duplicates.length > 0) {
```
mumanity, line 68: You can implement optional chaining here `if (duplicates?.length > 0)`.

mumanity (approved):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SubModelTree/index.tsx
jwills-jdubs: I did that but it fails because undefined cannot be checked for > 0 apparently

jwills-jdubs: Or it is an unreliable check.

mumanity: True. 1undefined` is true in javascript. If that's what you're working around, then it should be updated to `duplicates !== undefined && duplicates.length > 0` for readability & preventing linters from re-writing it.

PR #367, opened 2022-11-15
fix(composer): Fix to persist drag/drop actions after refresh

mumanity:
## Overview
Updating import for `updateSceneNodeInternal` to come from `useSceneDocument` to achieve persist on refresh. 

## Legal


TheEvilDev (approved):
I don't expect this to be any different, but worth a try.

PR #371, opened 2022-11-16
fix(composer): Fixes expand button on hierarchies

jwills-jdubs:
## Overview
We weren't using the childRefs prop and we needed to just fetch from the node

## Legal


packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -25,7 +25,6 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
   name: labelText,
   componentTypes,
   enableDragAndDrop,
-  childRefs = [],
```
mumanity, line 28: This will break drag & drop if it is removed.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/model/ISceneHierarchyNode.ts
```diff
@@ -1,7 +1,6 @@
 interface ISceneHierarchyNode {
   objectRef: string;
   name: string;
-  childRefs: string[];
```
mumanity, line 4: This is needed.

mumanity (changes requested):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
jwills-jdubs: It's not used anywhere

packages/scene-composer/src/components/panels/SceneHierarchyPanel/model/ISceneHierarchyNode.ts
jwills-jdubs: Not used anywhere

mumanity (approved):

PR #373, opened 2022-11-16
fix(composer): Fix for useEffect error

mumanity:
## Overview
Fix to unsubscribe in `useEffect` that was triggering error on drag/drop. Includes fix for type safety in `move` function.

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
+++ b/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
@@ -79,12 +79,18 @@ export const useChildNodes = (parentRef: string) => {
   const [childNodes, setChildNodes] = useState([] as ISceneHierarchyNode[]);
 
   useEffect(() => {
+    let mounted = true;
     (async () => {
       setLoading(true);
       const results = await getChildNodes(parentRef);
-      setChildNodes(results);
-      setLoading(false);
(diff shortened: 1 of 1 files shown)


packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -192,7 +197,7 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
         // remove child ref from parent
         if (!oldParentRef) {
           const newRoots = document.rootNodeRefs.filter((ref) => ref !== objectRef);
-          updateDocumentInternal({ rootNodeRefs: newRoots });
+          updateDocumentInternal({ rootNodeRefs: newRoots } as RecursivePartial<Pick<ISceneDocumentInternal, 'unit'>>);
```
sheilaXu, line 200: you don't need to case. i updated the type of `updateDocumentInternal` and there shouldn't be type error here

mumanity: It's triggering an error in my IDE. Applied to appropriate type from what was inferred. Casting shouldn't hurt anything, but helps prevent errors. If this is the wrong type I'll gladly update it.

sheilaXu (approved):

PR #379, opened 2022-11-17
fix(composer): CSS Cleanup for Grafana

mumanity:
## Overview
CSS overrides for Grafana environment to correct hover/selected borders & radio buttons. 

## Legal


packages/scene-composer/src/components/RadioButton/radio-button.scss
```diff
@@ -1,13 +1,33 @@
...
-    background-color: #fff;
-    color: #fff;
-    border: 4px solid #00a1c9;
-    border-radius: 50%;
-    padding: 3px;
+    @include radio;
+  }
+}
+
+// Grafana override
```
sheilaXu, line 19: I don't think we should mention Grafana in here or in commit/PR message. It is just one of the AppKit client, but this would fix the override by all clients

mumanity: Fixed

sheilaXu (approved):

PR #380, opened 2022-11-17
fix(composer): Set tree hierarchy items to auto-collapsed on load

mumanity:
## Overview
Setting tree hierarchy to be collapsed rather than expanded on page load. 

## Legal


packages/scene-composer/src/components/panels/CommonPanelComponents.tsx
```diff
@@ -214,7 +214,7 @@ export interface ExpandableInfoSectionProps {
 export const ExpandableInfoSection: React.FC<React.PropsWithChildren<ExpandableInfoSectionProps>> = ({
   title,
   withoutSpaceBetween = false,
-  defaultExpanded = true,
+  defaultExpanded = false,
```
sheilaXu, line 217: are we sure we want to default collapse the sections in the right side component details panel?

sheilaXu:
when I reorder the expanded tree to be a child of another node, the tree will be auto collapsed. is this expected?

packages/scene-composer/src/components/panels/CommonPanelComponents.tsx
mumanity: Good catch! Fixed.

jwills-jdubs (approved):

jwills-jdubs (approved):

PR #385, opened 2022-11-21
feat(dashboard-side-panel): Create helper functions

square-li:
- Create helper functions `getProperty` and `setProperty` for fetching and updating widget states.
- Create `useInput` hook 

with these, we can easily create components to update widget states. For example, we have 
```
const widget: Widget = {
  ...
  annotations: {
    y: [
      {
        value: 10,
        color: 'black',
      },
    ],
  },
}; 
```
if we want to add an input field to change the color of the YAnnotation, all I need to do is
```
const AnnotationSettings = () => {
  const [color, updateColor] = useInput<string>('annotations.y.0.color');

  return (
      <Input
        value={color}
        onChange={(event) => {
          updateColor(event.detail.value);
        }}
      />
   
  );
};
```
and it will automatically get the color value.
image
when users update it value, it will create a new widget state without modify the old state, which is very helpful when adapting Redux in the next steps.

## Overview

## Legal

### DIFF
--- a/packages/dashboard/src/components/side-panel/utils/index.ts
+++ b/packages/dashboard/src/components/side-panel/utils/index.ts
@@ -1,0 +1,17 @@
+import { useEffect, Dispatch, useState } from 'react';
+// import { useDispatch } from 'react-redux';
+import { get, set, cloneDeep } from 'lodash';
+
+export const useInput: <T>(path: string) => [T, Dispatch<T>] = (path) => {
+  const selectedWidget = {}; // TODO: get selected widget from state manager or context
+  const [value, updateValue] = useState(get(selectedWidget, path));
+  // const dispatch = useDispatch();
+  useEffect(() => {
+    if (selectedWidget && get(selectedWidget, path) !== value) {
+      const deepClone = cloneDeep(selectedWidget);
+      const newWidget = set(deepClone, path, value);
+      // TODO: dispatch(onUpdateSelectedWidgetAction(newWidget))
+    }
+  }, [value]);
+  return [value, updateValue];
+};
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/components/side-panel/utils/index.ts
```diff
@@ -0,0 +1,57 @@
+import { useEffect, useState } from 'react';
+// import { useDispatch } from 'react-redux';
+
+type ReturnValueType<T = any> = T;
```
diehbria, line 4: Why? seems like an extra layer of obfuscation for no reason

packages/dashboard/src/components/side-panel/utils/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  // TODO: get selectedWidget from state
+  const selectedWidget = {
+    foo: 'foo',
+  };
+  const [value, updateValue] = useState<ReturnValueType>(getProperty(path, selectedWidget));
+  // const dispatch = useDispatch();
+  useEffect(() => {
+    if (getProperty(path, selectedWidget) !== value) {
+      const newWidget = setProperty(path, value, selectedWidget);
+      console.log('New widget states:', newWidget);
```
diehbria, line 52: remove console statement

packages/dashboard/src/components/side-panel/utils/index.ts
```diff
@@ -0,0 +1,57 @@
+import { useEffect, useState } from 'react';
+// import { useDispatch } from 'react-redux';
+
+type ReturnValueType<T = any> = T;
+export const getProperty: <ReturnValueType>(path: string, object: ReturnValueType) => unknown = (path, object) => {
```
diehbria, line 5: could we just use https://lodash.com/docs/4.17.15#get?

diehbria (changes requested):

diehbria (approved):

PR #390, opened 2022-11-29
fix(composer): Allows nodes to be dropped at root level

mumanity:
## Overview
Fix to allow dragging nodes to root. 
* Includes fix for expand/collapse button for submodels as children

## Legal

### DIFF
--- a/packages/scene-composer/src/components/Tree/tree.scss
+++ b/packages/scene-composer/src/components/Tree/tree.scss
@@ -42,6 +42,11 @@ $active-color: #ec7211;
     position: relative;
     user-select: none;
 
+    &:first-child {
+      padding-top: 1rem; 
+      padding-bottom: 1rem; 
+    }
+    
     &.drop,
     .drop {
       border-bottom: 3px solid #fff;
--- a/packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
+++ b/packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useCallback, useContext, useState } from 'react';
+import React, { FC, useCallback, useContext, useMemo, useState } from 'react';
 import { Object3D } from 'three';
 
 import ISceneHierarchyNode from '../../model/ISceneHierarchyNode';
@@ -34,19 +34,18 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
 
   const model = getObject3DBySceneNodeRef(key) as Object3D | undefined;
   const [childNodes] = useChildNodes(key);
-  const isValidModelRef = componentTypes?.find(
-    (type) =>
-      type === KnownComponentType.ModelRef &&
-      (type as unknown as IModelRefComponentInternal)?.modelType !== ModelType.Environment,
-  );
-
-  const [{ variation: subModelSelectionEnabled }] = useFeature(COMPOSER_FEATURES[COMPOSER_FEATURES.SubModelSelection]);
-  const showSubModel = subModelSelectionEnabled === 'T1' && isValidModelRef && !!model && !isViewing();
   const sceneComposerId = useContext(sceneComposerIdContext);
   const { getSceneNodeByRef } = useSceneDocument(sceneComposerId);
   const node = getSceneNodeByRef(key);
+  const component = findComponentByType(node, KnownComponentType.ModelRef) as IModelRefComponentInternal;
+  const componentRef = component?.ref;
+  const isValidModelRef = useMemo(() => {
+    return component && component?.modelType !== ModelType.Environment;
+  }, [component]);
+  const [{ variation: subModelSelectionEnabled }] = useFeature(COMPOSER_FEATURES[COMPOSER_FEATURES.SubModelSelection]);
+  const showSubModel = subModelSelectionEnabled === 'T1' && isValidModelRef && !!model && !isViewing();
   const isSubModel = !!findComponentByType(node, KnownComponentType.SubModelRef);
-  const componentRef = findComponentByType(node, KnownComponentType.ModelRef)?.ref;
+
   const { searchTerms } = useSceneHierarchyData();
   const isSearching = searchTerms !== '';
 
@@ -81,7 +80,7 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
       labelText={<SceneNodeLabel objectRef={key} labelText={labelText} componentTypes={componentTypes} />}
       onExpand={onExpandNode}
       expanded={expanded}
-      expandable={node && (node.childRefs.length > 0 || !!showSubModel) && !isSearching}
+      expandable={((node && node.childRefs.length > 0) || showSubModel) && !isSearching}
       selected={selected === key}
       selectionMode={selectionMode}
       onSelected={isViewing() ? onActivated : onToggle}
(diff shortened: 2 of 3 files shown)


packages/scene-composer/src/hooks/useEditorHelper.ts
```diff
@@ -53,6 +53,7 @@ export function useEditorHelper<T>(
     }
   });
 
+  /* istanbul ignore next */
```
mumanity, line 56: Logic in `useFrame` is tightly coupled to canvas rendering & should go through either manual or integration tests rather than unit tests.

TheEvilDev: Disagree, this logic is pretty easy to mock out and test, I don't think we should ignore it. Wheras an integration test in this place has no access to the logic, and this is typically tied to animations.

Really I question whether we need this use frame at all. We need to evaluate whether something should be visible or not 60 times a second?

TheEvilDev (changes requested):
Don't ignore coverage in the useFrame.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -192,50 +192,53 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
...
+        // Update the node position to remain in its world space
+        partial.transform = {
+          position: maintainedTransform.position.toArray(),
+          rotation: maintainedTransform.rotation.toVector3().toArray(),
+          scale: maintainedTransform.scale.toArray(),
+        };
+      }
+      // update node to have new parent
+      if (!newParentRef) {
+        updateSceneNodeInternal(objectToMoveRef, { parentRef: newParentRef });
```
sheilaXu, line 235: when the node is dropped at root, we should also add it to `rootNodeRefs`

packages/scene-composer/src/hooks/useEditorHelper.ts
mumanity: Removed the ignore, but this looks like it's helping Three.js communicate with React via their rendering cycle.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
mumanity: There's no need to modify the root as it gets updated with the re-renders. Validated that the updates were persistent in refresh via local console testing.

TheEvilDev:
From testing in the Storybook, the issue of dropping at root nodes still seems to be present. Let's discuss.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -81,7 +78,7 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
       labelText={<SceneNodeLabel objectRef={key} labelText={labelText} componentTypes={componentTypes} />}
       onExpand={onExpandNode}
       expanded={expanded}
-      expandable={node && node.childRefs.length > 0 && !isSearching}
+      expandable={((node && node.childRefs.length > 0) || isValidModelRef) && !isSearching}
```
sheilaXu, line 81: is this change to show expand button for submodel tree? if so, we should use `showSubModel` to be consistent with when the tree is actually rendered

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -34,19 +34,16 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
...
-  );
-
-  const [{ variation: subModelSelectionEnabled }] = useFeature(COMPOSER_FEATURES[COMPOSER_FEATURES.SubModelSelection]);
-  const showSubModel = subModelSelectionEnabled === 'T1' && isValidModelRef && !!model && !isViewing();
   const sceneComposerId = useContext(sceneComposerIdContext);
   const { getSceneNodeByRef } = useSceneDocument(sceneComposerId);
   const node = getSceneNodeByRef(key);
+  const component = findComponentByType(node, KnownComponentType.ModelRef) as IModelRefComponentInternal;
+  const componentRef = component?.ref;
+  const isValidModelRef = component && component.modelType !== ModelType.Environment;
```
sheilaXu, line 42: useMemo so we don't recalculate every time?

sheilaXu (changes requested):

packages/scene-composer/src/components/Tree/tree.scss
```diff
@@ -42,6 +42,11 @@ $active-color: #ec7211;
     position: relative;
     user-select: none;
 
+    &:first-child {
```
sheilaXu, line 45: with https://github.com/awslabs/iot-app-kit/pull/411 you should only need this css change

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mumanity: This is taking into consideration the logic for `EnvironmentModel' as those aren't supposed to have expand/collapse functionality.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mumanity: Good call out. Fixed.

TheEvilDev (dismissed):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
sheilaXu: so the problem your are trying to solve is when submodel tree is the only child item, the expand button is not shown. you should just need to handle this issue by showing the button when submodel tree is shown which is when `showSubModel` is true. 

`showSubModel` handles environment model and also viewing mode. with your change, the expand button will be shown when the model has no child at all in viewing mode

sheilaXu: so the problem your are trying to solve is when submodel tree is the only child item, the expand button is not shown. you should just need to handle this issue by showing the button when submodel tree is shown which is when `showSubModel` is true.

`showSubModel` handles environment model and also viewing mode. with your change, the expand button will be shown when the model has no child at all in viewing mode

sheilaXu (changes requested):

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mumanity: Incorrect. The problem that I am trying to solve is for the expand/collapse button to be shown when **any** child is present. Using `showSubModel` incorrectly associates this with only submodel logic, while `isValidModelRef` is more explicit as to what is/is not valid (submodel or not). 

Both changes work in composer & viewer which were tested on my end. Are you seeing different behavior on your's when pulling this branch locally? If so, please sync up with me so that we can determine why we would see inconsistent behavior. 

For the sake of readability & best practices, we should not have this explicitly tied to submodels.

sheilaXu (approved):

PR #391, opened 2022-11-29
chore(PublishStorybook): Publish Storybook Site to build artifacts

TheEvilDev:
## Overview
Adds a github action workflow that allows a reviewer to download the prebuilt Storybook site for Scene composer at the current revision.

You can download these artifacts from the workflow, and using a local static site tool like `httpserver` run the storybook website and play with new features before approving the CR, without having to check it out and run it locally.

## Legal

### DIFF
--- a/.github/workflows/publish-storybook.yml
+++ b/.github/workflows/publish-storybook.yml
@@ -1,0 +1,32 @@
+name: Publish Storybook
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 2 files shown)


haweston:
Seems like the steps to use the review tool should be captured in a README or something outside this PR?

sheilaXu (approved):

PR #393, opened 2022-11-29
chore(composer): add script to auto build composer change

sheilaXu:
## Overview
add script to auto build composer change

## Legal

### DIFF
--- a/packages/scene-composer/.eslintrc.js
+++ b/packages/scene-composer/.eslintrc.js
@@ -72,5 +72,5 @@ module.exports = {
     jest: true,
     es6: true,
   },
-  ignorePatterns: ['src/three/GLTFLoader.js', 'src/three/tiles3d/*'],
+  ignorePatterns: ['src/three/GLTFLoader.js', 'src/three/tiles3d/*', 'tools/watch-build.js'],
 };
--- a/packages/scene-composer/package.json
+++ b/packages/scene-composer/package.json
@@ -27,9 +27,11 @@
   },
   "scripts": {
     "build": "run-s compile",
+    "build:watch": "node ./tools/watch-build.js",
     "clean": "rimraf dist ./build/* node_modules",
     "compile": "run-s extract-msgs convert-svg tttsc copy-assets",
     "tttsc": "ttsc -p tsconfig.build.json",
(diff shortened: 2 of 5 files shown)


packages/scene-composer/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "scripts": {
     "build": "run-s compile",
+    "build:watch": "node ./tools/watch-build.js",
```
TheEvilDev, line 30: Javascript conventions would be just name this script "watch" it's a very common operation.

packages/scene-composer/.eslintrc.js
```diff
@@ -72,5 +72,5 @@ module.exports = {
     jest: true,
     es6: true,
   },
-  ignorePatterns: ['src/three/GLTFLoader.js', 'src/three/tiles3d/*'],
+  ignorePatterns: ['src/three/GLTFLoader.js', 'src/three/tiles3d/*', 'tools/watch-build.js'],
```
TheEvilDev, line 75: Did you look at just running the watch command with typescript instead of nodemon? This is a built in feature:

https://www.typescriptlang.org/docs/handbook/configuring-watch.html

packages/scene-composer/package.json
sheilaXu: we have another `test:watch` command, so I'm trying to be consistent with that

packages/scene-composer/.eslintrc.js
sheilaXu: nodemon was the first showing when I'm searching for watch files, so I stated with it.

I just tried this tsc watch feature, it works fine too. but since our build script includes all those steps `"run-s extract-msgs convert-svg tttsc copy-assets"` it won't be able to auto build for svg file changes for example. 

I could add both commands and let people choose which one to use - either faster but only for ts file changes, or slower but all changes

TheEvilDev: This is just rerunning the entire build for every change, that's very heavy. This is ok for now, but we should look for a better option. Typically these sort of scripts aren't needed.

TheEvilDev (approved):

TheEvilDev (approved):

PR #394, opened 2022-11-30
feat(dashboard): set up basic layout for side panel

square-li:
## Overview
- import @cloudscape-design/components
- set up expandable section 
- add a shared components `ExpandableSectionHeader` to allow use to handle actions that targets to the whole section. Eg. add a new threshold.
- fix jest transforms with @swc/jest

## Preview
image

## Legal

### DIFF
--- a/packages/dashboard/src/components/side-panel/index.css
+++ b/packages/dashboard/src/components/side-panel/index.css
@@ -1,0 +1,3 @@
+.iot-side-panel {
+  height: 100%;
+}
--- a/packages/dashboard/src/components/side-panel/index.tsx
+++ b/packages/dashboard/src/components/side-panel/index.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { Container, Header } from '@cloudscape-design/components';
+import PropertiesAlarmsSection from './sub-components/properties-alarms-section';
+import ThresholdsSection from './sub-components/thresholds-section';
+import ChartSettings from './sub-components/chart-settings';
+import DataSettings from './sub-components/data-settings';
+import './index.css';
+
+const IoTSidePanel = () => {
+  return (
+    <Container header={<Header variant="h3">Configurations</Header>} className={'iot-side-panel'}>
+      <PropertiesAlarmsSection />
+      <ThresholdsSection />
+      <ChartSettings />
+      <DataSettings />
+    </Container>
+  );
+};
+
+export default IoTSidePanel;
--- a/packages/dashboard/src/components/side-panel/shared/expandableSectionHeader.tsx
+++ b/packages/dashboard/src/components/side-panel/shared/expandableSectionHeader.tsx
@@ -1,0 +1,25 @@
+import React, { createElement, FC, PropsWithChildren } from 'react';
+import { Icon, IconProps } from '@cloudscape-design/components';
+import './styles.css';
+
+type ExpandableSectionHeaderProps = {
+  variant?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5';
+  onClickButton?: (e: Event) => void;
+  iconName?: IconProps.Name;
+};
+const ExpandableSectionHeader: FC<PropsWithChildren<ExpandableSectionHeaderProps>> = (props) => {
+  const { children, onClickButton, iconName = 'add-plus', variant = 'h5' } = props;
+  const element = createElement(variant, { className: 'expandable-section-header-text' }, children);
+  return (
+    <>
+      {element}
+      {onClickButton && (
+        <span className="expandable-section-header-icon">
+          <Icon name={iconName} />
+        </span>
+      )}
+    </>
+  );
+};
+
+export default ExpandableSectionHeader;
--- a/packages/dashboard/src/components/side-panel/sub-components/chart-settings.tsx
+++ b/packages/dashboard/src/components/side-panel/sub-components/chart-settings.tsx
@@ -1,0 +1,13 @@
+import React from 'react';
+import { ExpandableSection } from '@cloudscape-design/components';
+import ExpandableSectionHeader from '../shared/expandableSectionHeader';
+
+const ChartSettings = () => {
+  return (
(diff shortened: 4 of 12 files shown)


packages/dashboard/src/components/side-panel/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Container } from '@cloudscape-design/components';
+import PropertiesAlarmsSection from './sub-components/properties-alarms-section';
+import ThresholdsSection from './sub-components/thresholds-section';
+import ChartSettings from './sub-components/chart-settings';
+import DataSettings from './sub-components/data-settings';
+import './index.css';
+
+const IoTSidePanel = () => {
+  return (
+    <Container header={<h2>Configurations</h2>} className={'iot-side-panel'}>
```
tjuranek, line 11: Why aren't we using the `<Header />` from cloudscape here?

https://cloudscape.design/components/container/?tabId=playground

packages/dashboard/src/components/side-panel/index.css
```diff
@@ -0,0 +1,3 @@
+.iot-side-panel {
+  height: 100%;
```
tjuranek, line 2: Make sure we try this on different screen sizes, especially tablet/mobile if we're trying to support it. Percentage heights get funky sometimes with mobile browsers.

packages/dashboard/src/components/side-panel/shared/expandableSectionHeader.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FC } from 'react';
+import { Icon } from '@cloudscape-design/components';
+import './styles.css';
+
+const ExpandableSectionHeader: FC<{ text: string; onClickButton?: (e: Event) => void }> = ({ text, onClickButton }) => {
+  return (
+    <>
```
tjuranek, line 7: Shouldn't this have some semantic name? If it's a header can we take in a variant, but provide a default? If it's a header, `onClickButton` doesn't make sense as a prop name.

`type HeaderVariant = 'h1' | 'h2' | ... 'h6';`

packages/dashboard/src/components/side-panel/shared/expandableSectionHeader.tsx
```diff
@@ -0,0 +1,18 @@
...
+      {text}
+      {onClickButton && (
+        <span className="expandable-section-header-icon">
+          <Icon name="add-plus" />
```
tjuranek, line 11: Good opportunity to take the icon name in as a prop but provide a default.

packages/dashboard/src/components/side-panel/sub-components/chart-settings.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { ExpandableSection } from '@cloudscape-design/components';
+
+const ChartSettings = () => {
+  return (
+    <ExpandableSection headerText="Chart settings" defaultExpanded>
```
tjuranek, line 6: In the sub-components I'd have a `const [expanded, setExpanded] = useState<boolean>(true);` instead of using defaultExpanded. 

Then you can pass something to toggle it in through the onClick and have that working right away. Removing defaultExpanded will get rid of another prop in the future.

packages/dashboard/src/components/side-panel/sub-components/properties-alarms-section.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { ExpandableSection } from '@cloudscape-design/components';
+import ExpandableSectionHeader from '../shared/expandableSectionHeader';
+
+const PropertiesAlarmsSection = () => {
+  return (
+    <ExpandableSection
+      headerText={<ExpandableSectionHeader text="Properties & Alarms" onClickButton={() => {}} />}
```
tjuranek, line 8: This is inconsistently either a `string` or an `ExpandableSectionHeader` in these sub-components?

packages/dashboard/src/components/side-panel/sub-components/thresholds-section.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { ExpandableSection } from '@cloudscape-design/components';
+import ExpandableSectionHeader from '../shared/expandableSectionHeader';
+
+const ThresholdsSection = () => {
+  return (
+    <ExpandableSection
+      headerText={<ExpandableSectionHeader text="Thresholds" onClickButton={() => {}} />}
```
tjuranek, line 8: These are supposed to be ExpandableSections, can we add that functionality in this same pull request?

tjuranek (changes requested):
I'd make a few minor changes to this before merging in, overall looks good though!

packages/dashboard/src/components/side-panel/index.tsx
square-li: you're right. Header component is better here.

packages/dashboard/src/components/side-panel/shared/expandableSectionHeader.tsx
square-li: agree

packages/dashboard/src/components/side-panel/shared/expandableSectionHeader.tsx
square-li: It is more like a header with an icon button that allows users to click the icon button for other functionality, like adding a new property to the properties setting section. So we do need onClickButton for that.

packages/dashboard/src/components/side-panel/sub-components/chart-settings.tsx
square-li: We don't need to control state of expandable section. ExpandableSection handles that by itself.

packages/dashboard/src/components/side-panel/sub-components/thresholds-section.tsx
square-li: Same - ExpandableSection handles that functionality.

tjuranek (approved):
LGTM! 👍

PR #408, opened 2022-11-30
docs(SceneComposer): Cleaning up stories and adding Scene Composer wr…

TheEvilDev:
…apper/i18n

## Overview
Adds locale changer to global stories, refactors new SceneViewer Stories, and introduces new MDX driven Scene Composer Stores.

Still have outstanding issue with loading models, so you should use the work around of loading the old stories first, and then play with the new ones while I investigate the issue.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/interfaces/interfaces.tsx
+++ b/packages/scene-composer/src/interfaces/interfaces.tsx
@@ -39,7 +39,7 @@ export interface ISceneDocumentSnapshot {
    *
    * @param specVersion - version of the document format
    */
-  serialize(specVersion: string): string;
+  serialize(specVersion: string, ...stringifyArgs): string;
 }
 
 export interface ISceneDocument {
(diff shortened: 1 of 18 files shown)


jwills-jdubs (approved):

packages/scene-composer/src/interfaces/interfaces.tsx
```diff
@@ -39,7 +39,7 @@ export interface ISceneDocumentSnapshot {
    *
    * @param specVersion - version of the document format
    */
-  serialize(specVersion: string): string;
+  serialize(specVersion: string, ...stringifyArgs): string;
```
sheilaXu, line 42: why do we have those changes for `stringifyArgs`?

TheEvilDev: These are optional arguments, and what it lets me do is save scene files in pretty-print.

```
JSON.stringify(object); // results in the object being serialized, but single line
JSON.stringify(object, null, '\t'); // allows you to pretty print
```

Doesn't change existing behavior, but gives more options for end users.

PR #412, opened 2022-12-02
feat(dashboard): add context menu, send to back, bring to front, copy, paste

jmbuss:
## Overview
Adds a right click / long press context menu with the actions: copy, paste, delete, send to back, bring to front.

Adds implementation for copy, paste,  send to back, bring to front.

## Legal

### DIFF
--- a/packages/dashboard/jest.config.ts
+++ b/packages/dashboard/jest.config.ts
@@ -28,12 +28,11 @@ export default {
       },
     ],
   },
-  transformIgnorePatterns: [],
   setupFilesAfterEnv: ['mutationobserver-shim'],
   //transform: {
   //   '.+\\.ts$': 'ts-jest',
   //   '^.+\\.tsx?$': 'ts-jest',
   //   '^.+\\.(js|jsx)$': 'babel-jest',
   // },
-  // transformIgnorePatterns: ['node_modules/(?!@awsui/components-react)/'],
+  transformIgnorePatterns: [],
 };
--- a/packages/dashboard/src/components/contextMenu/contextMenuOptions.ts
+++ b/packages/dashboard/src/components/contextMenu/contextMenuOptions.ts
@@ -1,0 +1,113 @@
+import { ContextMenuMessages, keyboardShortcuts } from '../../messages';
+
+type ContextMenuOptionConfiguration = {
+  id: string;
+  action: () => void;
+  text: string;
+  disabled: boolean;
+  hotkey?: string;
+};
+
+type ContextMenuSectionConfiguration = {
+  id: string;
+  options: ContextMenuOptionConfiguration[];
+};
+
+export type ContextMenuConfiguration = ContextMenuSectionConfiguration[];
+
+export type ContextMenuConfigurationProps = {
+  messages: ContextMenuMessages;
+  actions: {
+    copyAction: () => void;
+    pasteAction: () => void;
+    deleteAction: () => void;
+    bringToFrontAction: () => void;
+    sendToBackAction: () => void;
+  };
+  state: {
+    hasSelectedWidgets: boolean;
+    hasCopiedWidgets: boolean;
+  };
+};
+
+type OptionCreator = (props: ContextMenuConfigurationProps) => ContextMenuOptionConfiguration;
+
+type ContextMenuConfigurationCreator = {
+  id: string;
+  options: OptionCreator[];
+};
+
+const createCopyOption: OptionCreator = (props: ContextMenuConfigurationProps): ContextMenuOptionConfiguration => {
+  const { actions, messages, state } = props;
+  return {
+    id: 'copy',
+    text: messages.copyText,
+    hotkey: keyboardShortcuts.copy,
+    disabled: !state.hasSelectedWidgets,
+    action: actions.copyAction,
+  };
+};
+
+const createPasteOption: OptionCreator = (props: ContextMenuConfigurationProps): ContextMenuOptionConfiguration => {
+  const { actions, messages, state } = props;
+  return {
+    id: 'paste',
+    text: messages.pasteText,
+    hotkey: keyboardShortcuts.paste,
+    disabled: !state.hasCopiedWidgets,
+    action: actions.pasteAction,
+  };
+};
+
+const createDeleteOption: OptionCreator = (props: ContextMenuConfigurationProps): ContextMenuOptionConfiguration => {
+  const { actions, messages, state } = props;
+  return {
+    id: 'delete',
+    text: messages.deleteText,
+    disabled: !state.hasSelectedWidgets,
+    action: actions.deleteAction,
+  };
+};
+
+const createBringToFrontOption: OptionCreator = (
+  props: ContextMenuConfigurationProps
+): ContextMenuOptionConfiguration => {
+  const { actions, messages, state } = props;
+  return {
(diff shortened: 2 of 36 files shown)


packages/dashboard/jest.config.ts
```diff
@@ -28,12 +28,14 @@ export default {
...
   },
-  transformIgnorePatterns: [],
   setupFilesAfterEnv: ['mutationobserver-shim'],
   //transform: {
   //   '.+\\.ts$': 'ts-jest',
   //   '^.+\\.tsx?$': 'ts-jest',
   //   '^.+\\.(js|jsx)$': 'babel-jest',
   // },
-  // transformIgnorePatterns: ['node_modules/(?!@awsui/components-react)/'],
+  transformIgnorePatterns: [
```
square-li, line 37: this issue has been fixed by introducing `@swc/jest` at this [commit](https://github.com/awslabs/iot-app-kit/commit/1a4f2a490ada12135f9591b2c6b0716bcc31a004) 
you might want to just rebase on that

packages/dashboard/src/components/contextMenu/contextMenuOptions.ts
```diff
@@ -0,0 +1,114 @@
...
+    text: messages.sendToBackText,
+    hotkey: keyboardShortcuts.sendToBack,
+    disabled: !state.hasSelectedWidgets,
+    action: actions.sendToBackAction,
+  };
+};
+
+const configuration: ContextMenuConfigurationCreator[] = [
+  {
+    id: 'section1',
```
square-li, line 101: nit: maybe use more semantical naming?

packages/dashboard/src/components/contextMenu/menu.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { useEffect, useState } from 'react';
+import { usePopper } from 'react-popper';
```
mpopv, line 2: Non-blocking: Is this a use case that the Cloudscape popover can't handle? https://cloudscape.design/components/popover/?tabId=playground

packages/dashboard/src/components/contextMenu/option.tsx
```diff
@@ -0,0 +1,27 @@
...
+export type ContextMenuOptionProps = {
+  disabled: boolean;
+  text: string;
+  hotkey?: string;
+  action: MouseEventHandler<Element>;
+};
+
+const ContextMenuOption: React.FC<ContextMenuOptionProps> = ({ disabled, text, hotkey, action }) => {
+  return (
+    <div
```
mpopv, line 14: Non-blocking: Could we make these minimally accessible elements instead of divs?

square-li (changes requested):

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -54,13 +67,19 @@ const selectedRect = (selection: Selection | undefined): Rect | undefined => {
 };
 import SidePanel from '../side-panel';
 
-const InternalDashboard = () => {
+type InternalDashboardProps = {
+  // merge(this.messageOverrides, DefaultDashboardMessages)
```
mpopv, line 71: Not sure if leaving this in is intentional?

packages/dashboard/jest.config.ts
jmbuss: 👍 perfect, I had introduced this to fix an issue when I added popperjs to the package. Looks like that makes this ignore unnecessary.

packages/dashboard/src/messages.ts
```diff
@@ -0,0 +1,42 @@
...
+  contextMenu: {
+    copyText: 'Copy',
+    pasteText: 'Paste',
+    deleteText: 'Delete',
+    bringToFrontText: 'Bring to Front',
+    sendToBackText: 'Send to Back',
+  },
+};
+
+const mod = isMacLike ? '⌘' : 'Ctrl';
```
mpopv, line 35: Nice!

packages/dashboard/src/styles/variables.css
```diff
@@ -8,7 +8,7 @@
   --selection-border-touch-width: 8px;
   --selection-radius-small: 4px;
   --stack-order-grid-image: -999999999;
-  --stack-order-selection-box: 9999999999999;
+  --stack-order-selection-box: 99999999;
```
mpopv, line 11: Could we make these into something more comprehensible? Hard to know what these indices are trying to accomplish with these values, and I don't think they really need to be this high.

packages/dashboard/src/util/browser.ts
```diff
@@ -0,0 +1 @@
+export const isMacLike = /(Mac|iPhone|iPod|iPad)/i.test(navigator.platform);
```
mpopv, line 1: [This web API is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform), and I think there might be a better way to do this - see:
https://stackoverflow.com/questions/3902635/how-does-one-capture-a-macs-command-key-via-javascript
https://stackoverflow.com/questions/24469943/detecting-command-c-keypress

packages/dashboard/src/components/contextMenu/menu.tsx
jmbuss: Hmm I would have to look at all of the features in cloudscape popover. I'm using popperjs for positioning, I don't want to have to contend with any cloudscape styling.

square-li (approved):
LGTM!

packages/dashboard/src/components/contextMenu/option.tsx
jmbuss: 👍 I'll convert these to the list notation tags

packages/dashboard/src/styles/variables.css
jmbuss: I've taken these from the original implementation. I'll have a look at refactoring the CSS

packages/dashboard/src/components/internalDashboard/index.tsx
jmbuss: Nope missed this!

packages/dashboard/src/components/contextMenu/option.tsx
mpopv: I think they'd need to be <a> or something keyboard-navigable https://stackoverflow.com/questions/3486110/make-a-list-item-clickable-html-css

mpopv: Either that or give them tabindex

packages/dashboard/src/util/browser.ts
jmbuss: This is being used in this scenario which the MDN documentation says is the "least bad option". haha. This is used to display the hotkeys for copy paste. 

> But there is one case where, among the options you could use, navigator.platform may be the least-bad option: When you need to show users advice about whether the modifier key for keyboard shortcuts is the ⌘ command key (found on Apple systems) rather than the ⌃ control key (on non-Apple systems)

Let me add a note above this method though that in other contexts we should have a different detection method like the documentation suggests.

packages/dashboard/src/components/contextMenu/option.tsx
jmbuss: I see, let me test that functionality.

PR #413, opened 2022-12-02
docs(StorybookEnhancements): Load AWS credentials from profile, load local scenes dynamically

TheEvilDev:
## Overview
- Adds an option to load AWS Credentials from a local profile, so you don't have to type them in all the time when doing testing.
- Removes direct dependency on `testScenes` which was statically loading all scenes, and instead lazy loads them from the public folder. It also dynamically lists all available local scenes so developers don't need to do anything except drop a `.scene.json` file into the public folder to create different test scenes.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


divya-sea (approved):
Confirmed with the author that if developers have multiple profiles configured then probably it will fail. Workaround is to restart with a different profile or update the credential fields yourself while it's running.

PR #417, opened 2022-12-05
fix(composer): support Windows dependency file paths in GLTF loader

hwandersman:
## Overview
When a GLTF file is exported on the Windows platform and has texture image files referenced in sub-folders, the generated GLTF file uses backslash character for folder hierarchy: `textures\\image.png`

The GLTFLoader uses this url in an S3 path like `s3://bucket/textures\\image.png` which cannot be resolved by S3.

## Verifying Changes

Used storybook to load an existing scene in my AWS account. The scene has a model GLTF with textures set up as the following:
```
"images": [
    {
      "uri": "textures\\plate_baseColor.jpeg"
    },
    {
      "uri": "textures\\tire_mat4_baseColor.jpeg"
    },
    {
      "uri": "textures\\brakes1_baseColor.jpeg"
    },
    {
      "uri": "textures\\brakes1_normal.png"
    }
  ],
```

Before this fix the textures are not rendered and an error is shown in the browser Console: 
```
THREE.GLTFLoader: Couldn't load texture textures\brakes1_baseColor.jpeg

Failed to load resource: the server responded with a status of 404 (Not Found)
```

After the fix all textures are loaded successfully.

### Testing

Added a 3D model of a Tesla car that can be used for testing. Upload it to your workspace's S3 bucket / resource library and add it to a scene. With this fix all textures in the scene should load.

## Legal


TheEvilDev (changes requested):
We should have a story that covers this case.

hwandersman:
> We should have a story that covers this case.

The issue is only when loading a model from S3, so we need a story that requires manual setup with an AWS account. Added an item in the backlog to do this.

TheEvilDev (approved):

PR #420, opened 2022-12-07
feat(dashboard): add dynamic widget to render different appkit components based on the provided tag

jmbuss:
## Overview
Renders a iot appkit react component based on the component tag provided to the widget object

## Legal


packages/dashboard/src/components/widgets/dynamicWidget.tsx
```diff
@@ -0,0 +1,127 @@
...
+  readOnly,
+  isEditing,
+  invalidTagErrorHeader,
+  invalidTagErrorSubheader,
+}) => {
+  const Component = ComponentMap[componentTag];
+  const componentIsRegistered = typeof Component !== 'undefined';
+
+  // eslint-disable-next-line
+  const props: any = {
```
mpopv, line 96: Non blocking: Is this the place where you were having difficulties with incompatible types across components? Do we think this is an intractable problem, or are we punting on it?

mpopv (approved):

square-li (approved):

packages/dashboard/src/components/widgets/dynamicWidget.tsx
jmbuss: Punting on it for now. It will be a little easier to solve when we move the package to react-components since that package already depends on components. It was a mess to depend on core / components / react-components to put together a single type in dashboard to be passed into react-components. We might still have to create some type in the components package, but we can start with a definition in react-components where those wrappers are defined.

PR #423, opened 2022-12-08
feat(dashboard): initial resource explorer v2

mpopv:
rec

## Overview
This PR implements a basic resource explorer for the left pane, based on the resource explorer component from the components package, using only React, source-iotsitewise, and Cloudscape components. It also implements the resource explorer in the dashboard story with mock data pulled from a real SiteWise demo dataset.

You can drill down into the asset tree and ascend to previous levels using the panel links and breadcrumb controls. Assets which have children will be linked.

This PR includes a dummy placeholder searchbar that currently does nothing, but matches the mockup for resource explorer search, serving as a placeholder demonstrating the anticipated final look of this component.

Next steps
- Support for asset properties
- Drag and drop onto widgets
- More comprehensive tests
- Search (how do we want to do this? switch to pulling entire asset tree at once?)
- Consider how we are interacting with source-iotsitewise. There some weirdness involved here stemming from incompatibilities due to source-iotsitewise being tightly coupled to the previous implementation of resource explorer. We may want to take some logic here and move it to source-iotsitewise or consider interacting with SiteWise APIs directly.

## Legal


packages/dashboard/src/components/dashboard/index.test.tsx
```diff
@@ -10,6 +10,7 @@ describe('Dashboard', () => {
     document.body.appendChild(container);
 
     const args: IotDashboardProps = {
+      query: undefined,
```
jmbuss, line 13: Is this related in any way to the query we'll use to build timeseries data?

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -370,7 +371,11 @@ const InternalDashboard: React.FC<InternalDashboardProps> = ({ messageOverrides
       </div>
       <div className="iot-dashboard-panes-area">
         <ResizablePanes
-          leftPane={<div className="dummy-content">Resource explorer pane</div>}
+          leftPane={
+            <div className="iot-resource-explorer-pane">
+              {query && <IotResourceExplorer query={query.assetTree.fromRoot()} />}
```
jmbuss, line 376: should we have an empty state here if no query is provided?

packages/dashboard/src/components/resourceExplorer/useBuildProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+  return parsed;
+};
+
+export const useBuildProvider = (query: TreeQuery<SiteWiseAssetTreeNode[], BranchReference, void>) => {
+  const [provider, setProvider] = useState<MaybeSiteWiseAssetTreeSessionInterface>(undefined);
+  const [errors, setErrors] = useState<ErrorDetails[]>([]);
+  const [rootItems, setRootItems] = useState<AssetSummary[] | undefined>(undefined);
+
+  useEffect(() => {
+    const nextProvider = query.build(uuidv4());
```
jmbuss, line 27: is there any reason we would need to track the session of this query?

packages/dashboard/src/store/state.ts
```diff
@@ -1,6 +1,8 @@
 import { DashboardConfiguration, Widget } from '../types';
+import { ListAssetsResponse } from '@aws-sdk/client-iotsitewise';
 
 export type DashboardState = {
+  query: null | ListAssetsResponse;
```
jmbuss, line 5: Is this the query result from the executed query passed into the internal dashboard?

packages/scene-composer/package.json
```diff
@@ -154,10 +154,10 @@
   "jest": {
```
jmbuss, line 154: should this file be included?

jmbuss (commented):
Left some clarification comment questions!

packages/dashboard/src/components/dashboard/index.test.tsx
mpopv: Yep, same thing! [It's the query from `source-iotsitewise`.](https://github.com/awslabs/iot-app-kit/blob/main/docs/AWSIoTSiteWiseSource.md) You should be able to hook into it to get timeseries data too.

packages/dashboard/src/components/internalDashboard/index.tsx
mpopv: Yes, we probably should. Also, I need change the name of that prop from `query` to `assetTree` or something; that's a holdover from the original resource explorer which called the asset tree prop 'query' as well, which is really confusing because the query is something different. Thanks for the reminder :)

packages/dashboard/src/components/resourceExplorer/useBuildProvider.ts
mpopv: Not sure what you mean

packages/dashboard/src/store/state.ts
mpopv: Ah, I think I forgot to remove this. I don't think we should put the query in dashboard state at all because the source-iotsitewise provider tracks its state internally. I will fix.

packages/scene-composer/package.json
mpopv: Hm, this appears to be autogenerated; maybe I ran the wrong test command at some point. I'll remove.

packages/dashboard/src/components/resourceExplorer/useBuildProvider.ts
jmbuss: I remember seeing session id as part of the build interface. Wasn't sure if that was important for any cleanup or reusing the same session if we update the component?

packages/dashboard/src/components/dashboard/index.test.tsx
jmbuss: cool. I'll plan on replacing the test query I've been using with this!

packages/dashboard/src/components/resourceExplorer/useBuildProvider.ts
mpopv: I was a little confused on that too. In the docs for the chart components, it says `(Optional) The ID of the widget. A widget is a visualization that you use the [insert name here] chart component to create` -- but the resource explorer isn't a widget, and there's no explanation in the resource explorer docs on why you would want to provide a widgetId, so I left it out for now. We can always add it in later easily if that was an oversight.

jmbuss: 👍 sounds good!

PR #428, opened 2022-12-14
feat(dashboard): add component palette

jmbuss:
## Overview
Adds the component palette to the dashboard for creating widgets.

The drag preview ux works by adding a custom drag layer component which uses the drag layer hook from `react-dnd`. The custom drag layer is where we can put any other draggable elements that need to be visualized. It contains a switch statement that renders different components based on the drag item type defined by the drag hook from `react-dnd`.

The drag preview for the component palette uses the same dynamic widget that the grid uses.

## Legal


packages/dashboard/src/components/grid/grid-dot.svg
```diff
@@ -1,6 +1,3 @@
-<svg height="10" width="10"
-     version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-     viewBox="0 0 10 10" xml:space="preserve" preserveAspectRatio="none">
-
-  <circle cx="1.2" cy="1.2" r="0.4" stroke="#e3e3e3" stroke-width="0.3" fill="#e3e3e3" shape-rendering="crispEdges"/>
-</svg>
+<svg viewBox="0 0 40 40" xmlns="http://www.w3.org/2000/svg">
+  <circle cx="2" cy="2" r="2" fill="#e3e3e3" />
```
mpopv, line 2: Nice!

packages/dashboard/src/components/grid/index.css
```diff
@@ -1,23 +1,26 @@
...
+
 .container {
   position: absolute;
   height: 100%;
 
-  /* margin: 0.5em; */
-
   /**  establish a depth context */
   z-index: 1;
+  transition: all 50ms;
```
mpopv, line 9: IMO we should never use transition all because that introduces the possibility of accidental unwanted transitions later. Can we target the propertie(s) we want to transition?

packages/dashboard/src/components/palette/component.css
```diff
@@ -0,0 +1,36 @@
+@import "../../styles/variables.css";
+
+.palette-component {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  color: var(--colors-component-palette);
+  fill: var(--colors-component-palette);
+  stroke: var(--colors-component-palette);
+  transition: all 250ms;
```
mpopv, line 10: Same here

mpopv (approved):

packages/dashboard/src/components/grid/index.css
jmbuss: Ya good idea. I'll change these.

PR #449, opened 2022-12-29
feat(dashboard): add text/ link widget to dashboard

jmbuss:
## Overview
Adds in a text widget to the dashboard.

To support this 1 major change was made. The Widget type was refactored to remove all of the AppKit specific properties. It now only contains identification properties (id, componentTag), sizing properties (height, width), and positioning properties (x, y, z). All Appkit specific properties were moved to their own type extending Widget. The same was done for Text widget. This is being classified as a primitive widget type.

_This should not cause merge conflicts unless you were modifying the `types.ts` file._

Text widget supports keyboard shortcuts when editing directly in the widget:

- mod+b : bold
- mod+i : italics
- mod+u : underline
- mod+= : font size + 1
- mod+- : font size - 1
- mod+shift+l : convert to link

widgets-text-with-link-test


square-li (approved):
LGTM

PR #452, opened 2023-01-05
fix(composer): Auto expand hierarchy when selecting node on scene

mukeshsahay:
## Overview
Currently, when a node is selected on the scene there is no change in the hierarchy except the selection of that node. If that node is collapsed under its parent, it remains so. This makes it difficult for the user to locate/view that node in the hierarchy.

When a scene object is selected, all parent nodes of it should auto expand.

**Steps to reproduce**:

1. Open the Scene Composer or Scene Viewer
2. Select an object on the scene

**Old Behavior**: Hierarchy does not auto expands

**New Behavior**: All parent nodes of it should auto expand

**Fix**: Pass a list of the nodes (parent nodes) that needs to be expanded based on the selected object to the Scene Hierarchy panel. Track the expanded or collapsed node in the hierarchy using this list. 

https://user-images.githubusercontent.com/43108986/210682355-0caf271c-4978-4feb-84a5-2a8a4146f3eb.mov

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -49,8 +47,33 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
   const { searchTerms } = useSceneHierarchyData();
   const isSearching = searchTerms !== '';
 
+  const isExpanded = () => {
+    return nodesToExpand.indexOf(key) > -1;
+  };
+
+  const removeNodeFromExpandList = (nodeRef: string) => {
+    if (nodesToExpand.indexOf(nodeRef) > -1) {
+      nodesToExpand.splice(nodesToExpand.indexOf(nodeRef), 1);
```
sheilaXu, line 56: It's not a good practice for the child to change the value of the prop passed from parent. the data flow should be top to down, not both.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyPanel.tsx
```diff
@@ -21,10 +23,36 @@ const strings = defineMessages({
...
+    const pathToRoot: string[] = [];
+    while (currentNode) {
+      currentNode = getSceneNodeByRef(currentNode.parentRef);
+      if (currentNode && pathToRoot.indexOf(currentNode.ref) === -1) {
+        pathToRoot.push(currentNode.ref);
+      }
+    }
+    pathToRoot.forEach((node) => {
+      if (expandedNodes.current.indexOf(node) === -1) {
+        expandedNodes.current.push(node);
```
sheilaXu, line 50: will you get duplicates in the array if you select -> deselect -> reselect the same node?

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyPanel.tsx
```diff
@@ -35,7 +63,7 @@ const SceneHierarchyPanel = () => {
             </Toolbar>
           )}
           <Main>
-            <SceneHierarchyTree enableDragAndDrop={canReorderHierarchy === 'T1'} />
+            <SceneHierarchyTree enableDragAndDrop={canReorderHierarchy === 'T1'} nodesToExpand={nodesToExpand} />
```
sheilaXu, line 66: it should be better to trigger a function that expends all parent nodes instead of calculating an array when selection changed, so there is no direct relation between the expanded and selected node

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mukeshsahay: Code not required with the changed approach.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyPanel.tsx
mukeshsahay: No.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyPanel.tsx
mukeshsahay: Changed the approach to handle auto expand. SceneHierarchyDataProvider is now maintaining a list of nodes from selected node to root and is updated when selection changes. This is used by the hierarchy tree/item to handle the default expand state.
User still has the ability to expand/collapse nodes manually using the tree view.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -49,6 +58,16 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
...
 
+  useEffect(() => {
+    /**
+     * Use default state only if node is not expanded
+     * If node is already expanded, then skip this
+     *  */
+    if (!expanded) {
+      setExpanded(defaultExpanded);
+    }
+  }, [defaultExpanded, pathFromSelectedToRoot]);
```
sheilaXu, line 69: do we need `pathFromSelectedToRoot` as a dependency here?

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
```diff
@@ -96,7 +115,12 @@ const SceneHierarchyTreeItem: FC<SceneHierarchyTreeItemProps> = ({
           {childNodes.map((node, index) => (
             <React.Fragment key={index}>
               {!isSearching && (
-                <SceneHierarchyTreeItem key={node.objectRef} enableDragAndDrop={enableDragAndDrop} {...node} />
+                <SceneHierarchyTreeItem
+                  key={node.objectRef}
+                  enableDragAndDrop={enableDragAndDrop}
+                  {...node}
+                  expanded={pathFromSelectedToRoot ? pathFromSelectedToRoot.indexOf(node.objectRef) > -1 : false}
```
sheilaXu, line 122: can be simplified to `pathFromSelectedToRoot.includes(node.objectRef)`

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/index.tsx
```diff
@@ -22,7 +22,12 @@ const SceneHierarchyTree: FC<SceneHierarchyTreeProps> = ({ enableDragAndDrop }:
   return (
     <EnhancedTree acceptDrop={AcceptableDropTypes} onDropped={dropHandler} droppable={enableDragAndDrop}>
       {rootNodes.map((root) => (
-        <SceneHierarchyTreeItem key={root.objectRef} enableDragAndDrop={enableDragAndDrop} {...root} />
+        <SceneHierarchyTreeItem
+          key={root.objectRef}
+          enableDragAndDrop={enableDragAndDrop}
+          {...root}
+          expanded={pathFromSelectedToRoot ? pathFromSelectedToRoot.indexOf(root.objectRef) > -1 : false}
```
sheilaXu, line 29: same here

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mukeshsahay: Yes, it is needed to update the default state when the selected node changes (which in turn updates pathFromSelectedToRoot).

sheilaXu: but if `pathFromSelectedToRoot` changes that impacts the current item, then `defaultExpanded` (the expended value from props) will be changed correct?

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mukeshsahay: I think the purpose is same just different approach. Also, the performance of both the approaches are identical. Is there any additional value other than coding style change?

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/index.tsx
mukeshsahay: same as above

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
sheilaXu: shorter code is more readable in general, especially for boolean

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mukeshsahay: There is a case where if user manually collapses a node which was expanded, the defaultExpanded state does not changes. In this case, it relies on pathFromSelectedToRoot.

sheilaXu: in that case, will `pathFromSelectedToRoot` be changed? isn't `pathFromSelectedToRoot` only depending on current selection?

mukeshsahay: I missed to mention that selection change is the next step after user manually collapses the node. And in that case, `pathFromSelectedToRoot` is updated.

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx
mukeshsahay: Done

packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/index.tsx
mukeshsahay: Done

sheilaXu (approved):

PR #454, opened 2023-01-08
feat(dashboard): add color picker

square-li:
## Overview
Add native color picker for updating threshold colors
image

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/dashboard/src/components/sidePanel/sections/index.css
+++ b/packages/dashboard/src/components/sidePanel/sections/index.css
@@ -19,6 +19,14 @@
   border-radius: 0.57rem;
 }
 
+/* hide default color input */
+.threshold-content-color-picker > input {
+  opacity: 0;
+  height: 100%;
+  width: 100%;
+  cursor: pointer;
+}
+
 .threshold-container > div > div:last-child {
   flex-grow: 1;
   justify-content: end;
--- a/packages/dashboard/src/components/sidePanel/sections/thresholdsSection.tsx
+++ b/packages/dashboard/src/components/sidePanel/sections/thresholdsSection.tsx
@@ -1,4 +1,4 @@
-import React, { FC, MouseEventHandler } from 'react';
+import React, { ChangeEventHandler, FC, MouseEventHandler } from 'react';
 import {
   Button,
   Checkbox,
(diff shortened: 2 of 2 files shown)


diehbria:
Color picker seems to be a bit overkill - has support for an alpha channel, both hex and RGBA inputs etc - while one of our main design principles is simplicity. are we sure this is the best color picker widget to go with?

.stylelintrc
```diff
@@ -2,11 +2,13 @@
   "extends": "stylelint-config-standard",
   "rules": {
     "no-empty-source": null,
-    "selector-class-pattern": "^([a-z][a-z0-9]*)(-[a-z0-9]+)*$",
+    "selector-class-pattern": "^([a-z][a-z0-9]*)((-|_)[a-z0-9]+)*$",
```
square-li, line 5: This is for modifying Cloudscape-UI css classes.

packages/dashboard/src/components/sidePanel/sections/index.css
```diff
@@ -19,6 +19,11 @@
   border-radius: 0.57rem;
 }
 
+/* remove default popover padding */
+.threshold-content-color-picker div.awsui_body_xjuzf_10pru_325.awsui_body-overflow-visible_xjuzf_10pru_339 {
```
mpopv, line 23: Is there any other way we can do this? Or is this necessary? This is probably going to break on every Cloudscape version, according to what the team has said about how they autogenerate classnames

jmbuss, line 23: is this a reliable selector? I used data attributes to do this in the text widget commit.

jmbuss (changes requested):

packages/dashboard/src/components/sidePanel/sections/index.css
square-li: Replaced react-color lib with native color picker <input type="color">. This has been removed.

square-li: Replaced react-color lib with native color picker <input type="color">. This has been removed.

square-li:
> Color picker seems to be a bit overkill - has support for an alpha channel, both hex and RGBA inputs etc - while one of our main design principles is simplicity. are we sure this is the best color picker widget to go with?

Yes you are right. Switched to the native color picker `<input type="color">` instead of relying on 3rd party lib.

mpopv (approved):

jmbuss (approved):

PR #481, opened 2023-01-13
feat(dashboard): resource explorer asset property drag and drop onto widgets

mpopv:
Screen Shot 2023-01-13 at 9 44 12 AM

## Overview
Previously, in #423, we added a resource explorer on the left sidebar that allows the customer to load and view their hierarchy of SiteWise assets. This change adds the ability to also load aliased asset properties for those assets and drag-and-drop them onto a chart widget to begin sourcing data for that chart from the asset property.

To use the SiteWise ListAssetProperties API this change relies on, we needed to update `@aws-sdk/client-iotsitewise`, which also required updating Storybook's builder to Webpack 5 due to `@aws-sdk/client-iotsitewise` dropping support for IE 11 and shipping with non-Webpack-4 compliant syntax like nullish coalescing.

## Next steps
This is the first step of a multi-PR incremental delivery. Next steps include:
1. Dispatch widget updates to Redux
2. Handle multiple assets / asset properties in the same chart
3. Determine how to handle data types (Identify in resource explorer? Explicitly 'reject' somehow when dropped onto a chart that displays a different type of data?)
4. Cache IotSiteWiseClient call results - currently, this makes a request for asset properties on every crumb click. We should cache results similar to how `@aws-sdk/source-iotsitewise` does to prevent duplicate calls.
5. Unify invocations of IotSiteWiseClient - either hoist into the top-level dashboard or else build directly into `@aws-sdk/source-iotsitewise`.
6. Tests including integration tests for full user flow of load resources => navigate to asset property => drag onto chart => see data populate.
7. Style updates - currently there are some simple styles written for the demo, but these should be updated to use Cloudscape constants, and we should migrate the side panel from using Table to using Tiles or a similar custom structure.

## Legal


packages/dashboard/.env.example
```diff
@@ -11,4 +11,4 @@ DEMO_TURBINE_ASSET_1_PROPERTY_3=xxxxx
 DEMO_TURBINE_ASSET_1_PROPERTY_4=xxxxx
 
 AGGREGATED_DATA_PROPERTY=xxxxx
-AGGREGATED_DATA_PROPERTY_2=xxxxx
\ No newline at end of file
+AGGREGATED_DATA_PROPERTY_2=xxxxx
```
mpopv, line 14: Prettier was insistent on this

packages/dashboard/src/components/widgets/widget.tsx
```diff
@@ -1,19 +1,29 @@
...
+  }[];
+};
+
 export type WidgetProps = {
   readonly: boolean;
   query?: SiteWiseQuery;
   isSelected: boolean;
   cellSize: number;
   widget: Widget;
+  widgets: Widget[];
```
jmbuss, line 26: Is this used in the component?

packages/dashboard/src/components/widgets/widget.tsx
```diff
@@ -29,8 +39,31 @@ const WidgetComponent: React.FC<WidgetProps> = ({
...
 
+  // TODO: Replace with Redux dispatch
+  const [assets, setAssets] = useState<null | AssetsParam>(null);
+  const [internalWidget, setInternalWidget] = useState(widget);
+
+  const [, drop] = useDrop(
+    () => ({
+      accept: ItemTypes.ResourceExplorerAssetProperty,
+      drop: ({ queryAssetsParam }: { queryAssetsParam: AssetsParam }) => {
+        setAssets(queryAssetsParam);
```
jmbuss, line 50: The update widget action should be available, is it possible to dispatch that from here appending the queryAssetsParam to assets?

packages/dashboard/src/components/resourceExplorer/getCurrentAssets.ts
```diff
@@ -2,6 +2,17 @@ import { BranchReference } from '@iot-app-kit/source-iotsitewise';
 import { AssetSummary, AssetHierarchy } from '@aws-sdk/client-iotsitewise';
 import { HIERARCHY_ROOT_ID } from '.';
 import { AssetNode, BranchReferenceWithAssetIds, MaybeSiteWiseAssetTreeSessionInterface } from './types';
+import { ExtendedPanelAssetSummary, EitherAssetSummary } from '.';
+
+const rootAssetsHeaderItem: ExtendedPanelAssetSummary = {
+  isHeader: true,
+  name: 'Root Assets',
```
jmbuss, line 9: Is this name displayed on the dashboard? If so can we include a section in messageOverrides?

packages/dashboard/src/components/resourceExplorer/components/style.css
```diff
@@ -0,0 +1,36 @@
+.resource-explorer-panel-asset-property-drag-handle {
+  cursor: grab;
+  padding: 2px 10px;
+  outline: 1px solid #efefef;
```
jmbuss, line 4: We might have a variable for this color

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
```diff
@@ -1,50 +1,87 @@
...
+    <div className="resource-explorer-panel-asset-property-drag-ghost">
+      <Icon name="expand" /> <Box variant="awsui-key-label">{item.name}</Box>
+    </div>
+  );
+};
+
+const PanelEmpty: React.FC = () => (
   <Box textAlign="center" padding={{ bottom: 's' }} variant="p" color="inherit">
-    No resources to display.
+    Asset has no properties or child assets.
```
jmbuss, line 22: i18n

jmbuss (changes requested):

jmbuss:
Is it possible to remove the box shadow on the drag handle?

packages/dashboard/src/components/widgets/widget.tsx
mpopv: I am planning on doing this as a followup, originally with a separate branch I wrote but now potentially integrating it with Bill's branch that does this.

packages/dashboard/src/components/widgets/widget.tsx
mpopv: Oops, it was at one point in development but is no longer. Removed.

packages/dashboard/src/components/widgets/widget.tsx
```diff
@@ -1,19 +1,29 @@
...
 import { DashboardMessages } from '../../messages';
 
 import { DashboardConfiguration, Widget } from '../../types';
 import { gestureable } from '../internalDashboard/determineTargetGestures';
 import DynamicWidgetComponent from './dynamicWidget';
+import { ItemTypes } from '../dragLayer/itemTypes';
 
 import './widget.css';
 
+export type AssetsParam = {
```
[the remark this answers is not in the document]
mpopv: Nope, fixed to use that type instead.

packages/dashboard/src/components/resourceExplorer/getCurrentAssets.ts
mpopv: Added to messageOverrides

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
mpopv: Added to messageOverrides

mpopv:
> Is it possible to remove the box shadow on the drag handle?

Sure, was mostly trying to imitate Cloudscape drop shadow there; the actual visual structure of the sidebar is likely to change anyway after design review.

packages/dashboard/src/components/resourceExplorer/components/style.css
```diff
@@ -0,0 +1,36 @@
...
+  margin: -10px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border-radius: 5px;
```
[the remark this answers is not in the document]
mpopv: Linked to a variable

packages/dashboard/src/components/resourceExplorer/components/style.css
mpopv: Linked to a variable

packages/dashboard/src/components/widgets/widget.tsx
mpopv: The reason I want to separate it out is that it's a little more complicated -- we want to be able to drag multiple assets and multiple asset properties onto a widget, so it has to both be able to append assets as well as determine if asset properties being passed belong to assets already attached to the widget, and if so, append the property to the existing asset rather than appending a duplicate asset.

jmbuss: Ahhh ok that makes sense. Plus with dan's work on kpis, sometimes we want to replace the asset property instead of append it.

jmbuss (approved):

packages/dashboard/src/components/resourceExplorer/getCurrentAssetProperties.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  let client: IoTSiteWiseClient;
+  try {
+    client = new IoTSiteWiseClient({
+      region: REGION,
+      credentials: getEnvCredentials(),
+    });
+  } catch (e) {
+    console.log(e);
```
NorbertNader, line 29: What do you think of using console.warn here instead? Maybe even just throw an error? Thinking of how we'd use this for collecting metrics.

packages/dashboard/src/components/resourceExplorer/index.tsx
```diff
@@ -35,17 +48,26 @@ export const IotResourceExplorer = ({ treeQuery }: IotResourceExplorerProps) =>
...
-    const currentItems = getCurrentItems(provider, currentBranchId);
-    setPanelItems(currentItems);
-  }, [provider?.branches, provider?.assetNodes, currentBranchId]);
+    (async () => {
+      const [currentAssets, currentAssetProperties]: [EitherAssetSummary[], EitherAssetSummary[]] = await Promise.all([
+        getCurrentAssets(provider, currentBranchId, messageOverrides),
+        getCurrentAssetProperties(provider, currentBranchId, messageOverrides),
+      ]);
+
+      console.log({ currentAssets, currentAssetProperties });
```
NorbertNader, line 65: What is the reasoning behind this console.log?

Issue #462, opened 2023-01-11
Dashboard: Lookup asset definition


square-li:
as a user I can see the asset property information (name, unit and color) for a widget when i select it

PR #484, opened 2023-01-18
feat(dashboard): add properties section to setting panel

square-li:
## Overview
https://github.com/awslabs/iot-app-kit/issues/462

Redux actions
- add new redux action UpdateAssetQuery
  - this action will assign `propertyId` as the default `refId` if `refId` is not provided, which is useful when applying `styleSettings` to widget.
- add new redux action UpdateAssetDescription to store DescribeAsset results
- add a side effect to UpdateAssetQuery, which will call IotSitewiseClient to describe unknown assets and then store the response by UpdateAssetDescription

UI
- add property section
- add property component
- change property color by color picker
- display property name/alias/dataType/unit when property is found in AssetDescription
- add MessageOverride for i18n

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/dashboard/src/components/sidePanel/sections/propertiesAlarmSection/index.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { FC } from 'react';
```
square-li, line 1: This file was `propertiesAlarmSection.tsx`. I moved it here with a new component `PropertyComponent` and renamed it.

packages/dashboard/src/components/sidePanel/index.css
```diff
@@ -2,4 +2,5 @@
   height: 100%;
   max-height: 100vh;
   overflow-y: scroll;
+  min-width: 400px;
```
mpopv, line 5: Does this cause any issues with the resizable panes? What happens if you resize the side panel down to <400px?

packages/dashboard/src/components/sidePanel/sections/thresholdsSection.tsx
```diff
@@ -55,14 +55,9 @@ const ThresholdComponent: FC<{ path: string; deleteSelf: () => void }> = ({ path
       <div className="threshold-container">
         <SpaceBetween size="xs" direction={'horizontal'}>
           if
-          <Select
-            className="threshold-content-select-label"
-            options={ComparisonOperatorOptions}
-            selectedOption={selectedOption}
-            onChange={onUpdateComparator}
-          />
+          <Select options={ComparisonOperatorOptions} selectedOption={selectedOption} onChange={onUpdateComparator} />
```
square-li, line 58: Removed unused className and then reformatted by prettier. No actual changes.

packages/dashboard/src/store/actions/updateAssetQuery/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  };
+};
+
+export const updateAssetQuery = (state: DashboardState, action: UpdateAssetQueryAction): DashboardState => {
+  const findAndUpdateWidgetById = (widgets: Widget[]) =>
+    widgets.map((w) => {
+      if (w.id === action.payload.widget.id) {
+        return {
+          ...action.payload.widget,
+          assets: action.payload.assetQuery,
```
mpopv, line 55: Will this delete asset(s) previously attached to the widget and replace them with new asset(s) from the payload? Do we always want to do that, or are there circumstances in which we might want to add assets to a widget with some assets already? I'm thinking particularly of dragging multiple assets/properties from the resource explorer onto a single chart, do we want to 'add' each to the chart, or replace what's already there?

mpopv: (Not necessarily saying the way you're doing it here is wrong, totally fine if this is a first step and that's a feature to be added later)

packages/dashboard/src/components/sidePanel/index.css
square-li: It doesn't cause any issue with resizable panes. When it <400px, setting panel will change to vertically scrollable. This was mainly for limiting the setting panel becoming too narrow to pushing contents around.

packages/dashboard/src/store/actions/updateAssetQuery/index.ts
square-li: UpdateAssetQuery simply replaces widget's assets list with the new payload. We might want to have 2 extra actions like RemoveAssetQuery and AddAssetQuery to handle removing and adding assets. Or we need to handle them when creating the payload.

mpopv (approved):

PR #485, opened 2023-01-18
feat(dashboard): add readonly view for dashboard

jmbuss:
## Overview
Add a readonly property for the dashboard which makes the dashboard only render the grid area. It also disables the context menu and the drag and drop functionality.

This adds a new story for readonly mode. The Save button in the Main dashboard story will write the current state to local storage to be used in the readonly story.


jebek (approved):
LGTM

Should `readonly` be camel case?

square-li (approved):

PR #492, opened 2023-01-20
feat(dashboard): dispatch dropped asset property to Redux

mpopv:
## Overview
This change switches the addition of asset properties from an internal state value in `widget.tsx` to a Redux dispatch to the actual widget state.

There are a couple of issues here -- the first is that converting a `Widget` to an `AppKitWidget` is kind of clunky, and the second is that there appears to be a bug in our type annotations; onUpdateAssetQueryAction is typed as if it expects an `AssetQuery[]` but it actually needs to be passed an `AssetQuery`. I'm choosing to punt on both of them for now because I think we may want to step back and think more holistically about data structure, and I'm writing a doc to try and do that.

## Legal

### DIFF
--- a/packages/dashboard/src/components/widgets/widget.tsx
+++ b/packages/dashboard/src/components/widgets/widget.tsx
@@ -1,13 +1,10 @@
+import React from 'react';
 import { SiteWiseQuery } from '@iot-app-kit/source-iotsitewise';
-import React, { useState, useEffect } from 'react';
-import { useDrop } from 'react-dnd';
 import { DashboardMessages } from '../../messages';
-
 import { DashboardConfiguration, Widget } from '../../types';
 import { gestureable } from '../internalDashboard/determineTargetGestures';
 import DynamicWidgetComponent from './dynamicWidget';
-import { ItemTypes } from '../dragLayer/itemTypes';
-import { AssetQuery } from '@iot-app-kit/core';
+import { WidgetDropTarget } from './widgetDropTarget';
 
 import './widget.css';
 
@@ -17,7 +14,6 @@ export type WidgetProps = {
   isSelected: boolean;
   cellSize: number;
   widget: Widget;
-  widgets: Widget[];
   viewport: DashboardConfiguration['viewport'];
   messageOverrides: DashboardMessages;
 };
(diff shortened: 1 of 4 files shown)


jmbuss (changes requested):
Can we move the drop logic from the this widget component into some appkit widget wrapper that is rendered by dynamic widget? This component was meant to just handle the xyz positioning. 

That should help with the typing since dynamic widget already has to figure out what type of widget we are trying to render. This way dynamic widget identifies the type as an AppKitWidget and then renders the wrapper and the chart from the component map. The wrapper would be able to define an appkit widget as the prop type and contain the logic for appending / replacing the assetquery.

packages/dashboard/src/components/widgets/widgetDropTarget.tsx
```diff
@@ -0,0 +1,39 @@
...
+      accept: ItemTypes.ResourceExplorerAssetProperty,
+      drop: ({ queryAssetsParam }: { queryAssetsParam: AssetQuery }) => {
+        const newAssetQueries = queryAssetsParam;
+
+        const appKitWidget = structuredClone(widget) as AppKitWidget;
+        appKitWidget.widgetId = widget.id;
+
+        dispatch(
+          onUpdateAssetQueryAction({
+            // TODO: There is a mismatch between type expectations and what actually results in data being retrieved here; this expects an AssetQuery[] but actually requires an AssetQuery
```
square-li, line 28: What if we are adding two properties from different asset?

mpopv: I agree this is a problem; I want to discuss the best way to handle it from a data structure perspective with the team before forging ahead with a specific solution.

mpopv:
> Can we move the drop logic from the this widget component into some appkit widget wrapper that is rendered by dynamic widget? This component was meant to just handle the xyz positioning.
> 
> That should help with the typing since dynamic widget already has to figure out what type of widget we are trying to render. This way dynamic widget identifies the type as an AppKitWidget and then renders the wrapper and the chart from the component map. The wrapper would be able to define an appkit widget as the prop type and contain the logic for appending / replacing the assetquery.

Done!

packages/dashboard/src/components/widgets/widget.tsx
```diff
@@ -33,50 +29,29 @@ const WidgetComponent: React.FC<WidgetProps> = ({
...
-      <DynamicWidgetComponent
-        readOnly={readOnly}
-        query={query}
-        viewport={viewport}
-        widget={internalWidget}
-        isSelected={isSelected}
-        widgetsMessages={messageOverrides.widgets}
-      />
-    </div>
+    <WidgetDropTarget widget={widget}>
```
jmbuss, line 33: I think this should be one of the options that dynamic widget renders. Text widgets are one of the options Dynamic widget renders but shouldn't be a drop target.

jmbuss (changes requested):

PR #496, opened 2023-01-23
fix: propertyName with multi hyphen not working

sheilaXu:
## Overview
when the propertyName used in rules has multiple hyphen characters, the rule won't be evaluated properly because only one occurrence of the hyphen is escaped. this change is to escape all occurrences.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/tests/utils/dataBindingUtils.spec.ts
+++ b/packages/scene-composer/tests/utils/dataBindingUtils.spec.ts
@@ -276,7 +276,7 @@ describe('ruleEvaluator', () => {
   });
 
   it('should property names with special characters should work correctly', async () => {
-    const key1 = 'my-temperature';
+    const key1 = 'my-temperature-prop';
     const key2 = 'myRPM';
     const value = {
       [key1]: 22,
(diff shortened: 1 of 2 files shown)


packages/scene-composer/tests/utils/dataBindingUtils.spec.ts
```diff
@@ -276,7 +276,7 @@ describe('ruleEvaluator', () => {
   });
 
   it('should property names with special characters should work correctly', async () => {
-    const key1 = 'my-temperature';
+    const key1 = 'my-temperature-prop';
     const key2 = 'myRPM';
```
divya-sea, line 280: what happens when customer uses underscore and special characters in the propertyId? Will rule still work as expected?

sheilaXu: underscore won't be touched and will still work as expected. hyphen is the only special char in property name and is fixed now

divya-sea: Confirmed from offline conversation. The change will work the following cases 
no dash in the propertyId
Multiple dash in the propertyId
single dash in the propertyId

divya-sea (approved):

PR #513, opened 2023-01-23
feat(ViewportComponents): React components for managing viewports

TheEvilDev:
DRAFT: This is not ready for prime time, just sending for initial feedback.

## Overview
Provide an explanation of what the change is:

- Adds a react context and supporting components for managing a viewport for a group of components
- Removes the need for passing viewport property to individual charts, which get it via the `useViewport` hook
- Initial storybookjs setup for testing these changes
- `ViewportControls` implementation based on the Polaris DateRange Picker component

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/react-components/src/viewport-manager/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <ViewportContext.Provider
+      value={{
+        reset,
+        update,
+        viewport,
+      }}
+    >
+      <WebglContext>{children}</WebglContext>
```
diehbria, line 57: should remove the webglcontext component from the viewport manager

TheEvilDev, line 57: Don't use WebglContext here, this won't work nicely if you use or nest multiple viewport managers.

packages/react-components/src/viewport-controls/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+    amount,
+    unit: reverseMapUnit(unitRef),
+  };
+};
+
+const ViewportControls = ({ value: defaultValue, onChange, ...props }: DateRangePickerProps) => {
+  const { update, reset, viewport } = useViewport();
+
+  const value: DateRangePickerProps.Value | null = useMemo(() => {
+    const relativeViewport = viewport as { duration: string };
```
NorbertNader, line 46: nit: this is interesting but maybe we could achieve the same with predicates? 

So instead of 
```
if (relativeViewport?.duration) {
```
something like
```
isMinimalStaticViewport = (viewport: MinimalViewPortConfig): viewport is MinimalLiveViewport =>
  'duration' in viewport;

if (isRelativeViewport(viewport)) {
```

packages/react-components/src/viewport-controls/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const absoluteViewport = viewport as { start: Date; end: Date };
+    if (relativeViewport?.duration) {
+      const { amount, unit } = parseDuration(relativeViewport.duration);
+
+      return {
+        amount,
+        type: 'relative',
+        unit,
+      } as DateRangePickerProps.RelativeValue;
+    } else if (absoluteViewport?.start && absoluteViewport?.end) {
```
NorbertNader, line 56: nit: similar to above, core exports a `isMinimalStaticViewport` predicate

packages/react-components/src/enhancers/make-viewport-aware.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export interface IViewportProps {
+  viewport: Viewport;
+}
+
+function makeViewportAware<ElementType, PropType extends IViewportProps>(BaseComponent: any) {
+  const cp = forwardRef<ElementType, Omit<PropType, 'viewport'>>((props, ref) => {
+    const { viewport: sharedViewport } = useViewport(); // if inside a <ViewportManager /> tag, we want to use the shared viewport, otherwise we fallback to the original behavior where you ned a custom viewport for each element.
+
+    return <BaseComponent ref={ref} {...props} viewport={sharedViewport} />;
```
NorbertNader, line 13: probably only concern here is to still allow to override the viewport with `props .viewport || sharedViewport` for backward compatibility

TheEvilDev:
Feedback was incorporated and featured included in https://github.com/awslabs/iot-app-kit/pull/565

Issue #143, opened 2022-07-06
viewport skitter bug


boweihan:
**Describe the bug**
Widget viewport resets to live mode on first load, even after it is already interactive and gestures have been applied.

**To Reproduce**

Create a web page with an app kit component configured.

1. Refresh page
2. IMMEDIATELY pan to the left
3. IMMEDIATELY pan to the left again
4. Wait from a small moment
5. Viewport hops back into live mode, incorrectly

**Expected behavior**
Once a component is interactive, any panning should remain intact.

**Screenshots**
![bug-jumpy mp4](https://user-images.githubusercontent.com/10264198/177613822-2685c373-05a9-4366-896d-69937382f6ad.gif)

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Chrome

PR #515, opened 2023-01-25
fix: use viewport from provider #143

NorbertNader:
## Overview
Fixes skitter bug https://github.com/awslabs/iot-app-kit/issues/143
 - use viewport from provider instead of props, this was overriding actual viewport when interacting with chart 
 - use latest @synchro-charts/core with fix for live mode (do not emit dateRangeChange if not interacting with chart)

## Legal


jmbuss (commented):
If we are not using the prop, should we remove it and the watcher for it?

jmbuss (approved):

NorbertNader:
We still need the viewport, just don’t pass it directly to the synchro-charts components. Instead they are passed down from the provider.

PR #517, opened 2023-01-25
feat(dashboard): add BaseSettingSection

square-li:
## Overview
Add BaseSettingSection replacing Chart setting section.

image

### DIFF
--- a/packages/dashboard/src/components/sidePanel/index.tsx
+++ b/packages/dashboard/src/components/sidePanel/index.tsx
@@ -1,16 +1,14 @@
 import React, { FC } from 'react';
-import { Container, Header } from '@cloudscape-design/components';
-import PropertiesAlarmsSection from './sections/propertiesAlarmSection';
+import { Container, Header, SpaceBetween } from '@cloudscape-design/components';
 import './index.css';
 import { useSelector } from 'react-redux';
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/components/sidePanel/index.tsx
```diff
@@ -27,16 +25,18 @@ const SidePanel: FC<{ messageOverrides: DashboardMessages }> = ({ messageOverrid
...
-          <DataSettings />
-        </>
-      )}
+      <SpaceBetween size={'xs'} direction={'vertical'}>
+        <BaseSettings messageOverrides={messageOverrides} />
+        {isTextWidget && <TextSettings messageOverride={messageOverrides} />}
+        {isTextWidget && <LinkSettings messageOverride={messageOverrides} />}
+        {isAppKitWidget && (
+          <>
+            {/* WIP <PropertiesAlarmsSection messageOverrides={messageOverrides} />*/}
```
square-li, line 34: Uncomment these in a following PR.

jmbuss (approved):
LGTM, I think we should consider a refactor where we define some functional component for each like widget type(s) which organizes / renders the appropriate sections. This way we don't have so many conditional blocks as we add more types. Then maybe some switch statement based on the selected widget.

PR #519, opened 2023-01-27
feat(dashboard): initial resource explorer search

mpopv:
## Overview
- Previously, the initial version of the resource explorer would make a call to listAssetProperties whenever we move through the resource explorer onto an asset with properties. This is both inefficient and incompatible with a search that must keep a record of all properties, so here I've added a simple cache so we only attempt to retrieve an asset's properties once. This is simple kludge to enable an initial implementation of search, ultimately we want to be pre-loading the whole asset tree.
- Adds a searchbar that allows you to search all assets and properties. The checkboxes for assets and properties do properly filter the search but the alarms checkbox does not function yet.

## Legal

### DIFF
--- a/packages/dashboard/src/components/resourceExplorer/components/searchbar.tsx
+++ b/packages/dashboard/src/components/resourceExplorer/components/searchbar.tsx
@@ -1,33 +1,112 @@
-import React, { useState, Dispatch, SetStateAction } from 'react';
+import React, { useState, Dispatch, SetStateAction, useEffect } from 'react';
 import SpaceBetween from '@cloudscape-design/components/space-between';
 import Autosuggest from '@cloudscape-design/components/autosuggest';
 import Checkbox from '@cloudscape-design/components/checkbox';
+import { BaseChangeDetail } from '@cloudscape-design/components/input/interfaces';
+import { MaybeSiteWiseAssetTreeSessionInterface } from '../types';
+import { AssetPropertiesCache } from '../useAssetProperties';
+import { ExtendedPanelAssetSummary } from '..';
+import { DashboardMessages } from '../../../messages';
 
-const checkboxOnChange = (setShowFn: Dispatch<SetStateAction<boolean>>) => {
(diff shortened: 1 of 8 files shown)


packages/dashboard/src/components/resourceExplorer/components/searchbar.tsx
```diff
@@ -1,29 +1,106 @@
...
+  const onSearchChange = ({ detail }: { detail: BaseChangeDetail }) => {
+    setCrumbsToSearch();
+    setAutosuggestValue(detail.value);
+    let nextPanelItems: ExtendedPanelAssetSummary[] = [];
+    try {
+      const reduceResult: SearchOption[] = [];
+      nextPanelItems = searchOptions.reduce((acc, option) => {
+        if (!option.value.includes(detail.value)) return acc;
+        const nextPanelItem = { ...option, name: option.value };
+        console.log(nextPanelItem);
```
square-li, line 89: remove console.log

package.json
```diff
@@ -33,7 +33,7 @@
     "fix:stylelint": "stylelint '**/*.css' --fix",
     "test": "npm-run-all -p test:unit test:eslint test:stylelint test:git",
     "test:integration": "npm run test:integration -ws --if-present",
-    "test:eslint": "eslint --ext .js,.ts,.tsx . --max-warnings=46",
+    "test:eslint": "eslint --ext .js,.ts,.tsx . --max-warnings=49",
```
square-li, line 36: why are we rising this number?

square-li (changes requested):
Overall looks good with some minor issues.

packages/dashboard/src/components/resourceExplorer/components/searchbar.tsx
mpopv: Removed this

package.json
mpopv: We don't need to; removed this change

square-li (approved):

NorbertNader (approved):

PR #523, opened 2023-01-31
feature(dashboard): add axis setting section

square-li:
## Overview

add axis setting section 
- toggle showing X and Y axis
- update Y axis' label content

preview:
image

## Legal


packages/dashboard/src/components/sidePanel/shared/styles.css
```diff
@@ -27,3 +27,17 @@
   height: 1.57rem;
   border-radius: 0.57rem;
 }
+
+/* hide default color input */
+.color-picker-container > input {
+  opacity: 0;
```
mpopv, line 33: If this is intended to be noninteractive we might want to guarantee that with other things like `pointer-events: none`

mpopv (approved):

packages/dashboard/src/components/sidePanel/index.scss
```diff
@@ -1,6 +1,9 @@
+@import "~@cloudscape-design/design-tokens/index";
```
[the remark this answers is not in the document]
square-li: yes. otherwise it will fail on building&bundling.

packages/dashboard/src/components/sidePanel/shared/styles.css
square-li: the input will still be interactive. this was for hiding the default native color picker button as below:
image

PR #526, opened 2023-01-31
chore(core,source-iotsitewise): add aggregation layer to cache 

corteggiano:
## Overview
- add a layer to the cache so that we can support multiple aggregations in the future
```
[dataStreamId: string]: {
        rawData: DataStream | undefined,
        resolutions: {
                [resolution: number]: { 
                        [aggregationType in AggregateType]?: DataStreamStore | undefined; 
                } | undefined;
        } | undefined;
} | undefined;
```

## Verifying Changes
still receiving and displaying data properly :) 
image

## Legal

### DIFF
--- a/packages/core/src/data-module/TimeSeriesDataModule.spec.ts
+++ b/packages/core/src/data-module/TimeSeriesDataModule.spec.ts
@@ -542,7 +542,7 @@ describe('error handling', () => {
 
   const CACHE_WITH_ERROR: DataStreamsStore = {
     [DATA_STREAM_INFO.id]: {
-      [DATA_STREAM_INFO.resolution]: {
+      rawData: {
         id: DATA_STREAM.id,
         resolution: DATA_STREAM.resolution,
         dataCache: EMPTY_CACHE,
@@ -557,7 +557,7 @@ describe('error handling', () => {
 
   const CACHE_WITHOUT_ERROR: DataStreamsStore = {
     [DATA_STREAM_INFO.id]: {
-      [DATA_STREAM_INFO.resolution]: {
+      rawData: {
         id: DATA_STREAM.id,
         resolution: DATA_STREAM.resolution,
         dataCache: EMPTY_CACHE,
--- a/packages/core/src/data-module/TimeSeriesDataModule.ts
+++ b/packages/core/src/data-module/TimeSeriesDataModule.ts
@@ -80,19 +80,24 @@ export class TimeSeriesDataModule<Query extends DataStreamQuery> {
   }) => {
     const requestedStreams = await this.dataSourceStore.getRequestsFromQueries({ queries, request });
 
-    const isRequestedDataStream = ({ id, resolution }: RequestInformation) =>
-      this.dataCache.shouldRequestDataStream({ dataStreamId: id, resolution: parseDuration(resolution) });
+    const isRequestedDataStream = ({ id, resolution, aggregationType }: RequestInformation) =>
+      this.dataCache.shouldRequestDataStream({
+        dataStreamId: id,
+        resolution: parseDuration(resolution),
+        aggregationType,
+      });
 
     const requiredStreams = requestedStreams.filter(isRequestedDataStream);
 
     const requests = requiredStreams
-      .map(({ resolution, id, cacheSettings, meta }) =>
+      .map(({ resolution, id, cacheSettings, aggregationType, meta }) =>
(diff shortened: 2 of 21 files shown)


packages/source-iotsitewise/src/time-series-data/data-source.spec.ts
```diff
@@ -118,13 +120,15 @@ describe('initiateRequest', () => {
             end: new Date(),
             resolution: '0',
             fetchMostRecentBeforeEnd: true,
+            aggregationType: AGGREGATE_TYPE,
```
NorbertNader, line 123: Why do we need to pass the `aggregationType` if resolution is 0?

packages/core/src/data-module/data-cache/types.ts
```diff
@@ -17,6 +18,7 @@ export type HistoricalRequest = {
 export type DataStreamStore = {
   id: string;
   resolution: number;
+  aggregationType: AggregateType;
```
NorbertNader, line 21: I think aggregationType should be conditional since we don't need it for raw data.

packages/source-iotsitewise/src/time-series-data/client/legacy/getHistoricalPropertyDataPoints.ts
```diff
@@ -76,6 +81,7 @@ const getHistoricalPropertyDataPointsForProperty = ({
       onError({
         id,
         resolution: 0,
+        aggregationType: AggregateType.AVERAGE,
```
diehbria, line 84: should be undefined, historical is always raw

diehbria (changes requested):

packages/core/src/data-module/TimeSeriesDataModule.spec.ts
```diff
@@ -1323,3 +1323,29 @@ it('passes meta field returned from data source into the initiate request', asyn
     ]
   );
 });
+
+// Expected:
```
diehbria, line 1327: likely should remove this

diehbria (dismissed):

NorbertNader (dismissed):

packages/source-iotsitewise/src/time-series-data/client/client.spec.ts
```diff
@@ -38,6 +40,7 @@ describe('getHistoricalPropertyDataPoints', () => {
         end: endDate,
         resolution: '0',
         fetchFromStartToEnd: true,
+        aggregationType: AGGREGATE_TYPE,
```
NorbertNader, line 43: is there a way to write this so we don't need the `aggregationType` since resolution is 0 here?

packages/core/src/data-module/TimeSeriesDataModule.ts
```diff
@@ -93,6 +98,7 @@ export class TimeSeriesDataModule<Query extends DataStreamQuery> {
           store: this.dataCache.getState(),
           start: viewportStartDate(viewport),
           end: viewportEndDate(viewport),
+          aggregationType: parseDuration(resolution) === 0 ? undefined : AggregateType.AVERAGE,
```
NorbertNader, line 101: Why aren't we using the `aggregationType` here?

NorbertNader (approved):

packages/source-iotsitewise/src/time-series-data/client/client.spec.ts
```diff
@@ -241,6 +243,7 @@ describe('getLatestPropertyDataPoint', () => {
       end: endDate,
       resolution,
       fetchMostRecentBeforeEnd: true,
+      aggregationType: AGGREGATE_TYPE,
```
NorbertNader, line 246: looks like we are still requesting aggregationType here even for 0 resolution

NorbertNader (changes requested):

packages/source-iotsitewise/src/time-series-data/client/client.spec.ts
corteggiano: great catch! updated :)

corteggiano: fixed! thanks for catching that

NorbertNader (dismissed):

diehbria:
are there any changes required for `iot-twinmaker` datasource?

diehbria (approved):

PR #528, opened 2023-02-01
feature(dashboard): add legend configure section

square-li:
## Overview
Add legend configuration.
image

Width input only applies when legend is positioned as `right`. So it will be disabled when legend is positioned at bottom.

## Legal

### DIFF
--- a/packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.tsx
+++ b/packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.tsx
@@ -3,7 +3,8 @@ import { DashboardMessages } from '../../../../messages';
 import { useInput } from '../../utils';
 import { NonCancelableEventHandler } from '@cloudscape-design/components/internal/events';
 import { BaseChangeDetail } from '@cloudscape-design/components/input/interfaces';
-import { ExpandableSection, Grid, Input } from '@cloudscape-design/components';
+import { ExpandableSection, Grid, Input, SpaceBetween } from '@cloudscape-design/components';
+import '../index.scss';
 
 export const BaseSettings: FC<{ messageOverrides: DashboardMessages }> = ({
   messageOverrides: {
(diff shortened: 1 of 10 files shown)


packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.tsx
```diff
@@ -24,18 +25,41 @@ export const BaseSettings: FC<{ messageOverrides: DashboardMessages }> = ({
...
-        <div className="section-item-label">{baseSettings.y}</div>
-        <Input value={`${y}`} type="number" onChange={onYChange} data-test-id="base-setting-y-input" />
-      </Grid>
-      <Grid gridDefinition={gridDefinition}>
-        <div className="section-item-label">{baseSettings.width}</div>
-        <Input value={`${width}`} type="number" onChange={onWidthChange} data-test-id="base-setting-width-input" />
-        <div className="section-item-label">{baseSettings.height}</div>
-        <Input value={`${height}`} type="number" onChange={onHeightChange} data-test-id="base-setting-height-input" />
-      </Grid>
+      <SpaceBetween size={'xs'} direction={'vertical'}>
```
square-li, line 28: Rearrange layout

packages/dashboard/src/components/sidePanel/sections/legendSettingSection/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  messageOverrides: {
+    sidePanel: { legendSettingMessages },
+  },
+}) => {
+  const [legendSettings, setLegendSettings] = useAppKitWidgetInput('legend');
+  const onWidthChange = ({ detail }: { detail: InputProps.ChangeDetail }) => {
+    setLegendSettings({ ...defaultLegendSettings, ...legendSettings, width: parseInt(detail.value) });
+  };
+  const [showLegend, setShowLegend] = useState<boolean>(!!legendSettings);
+  const [recentLegendSettings, setRecentLegendSettings] = useState<LegendConfig>();
```
square-li, line 33: when user toggles back view legend, this preserves previous settings.

packages/dashboard/src/components/sidePanel/sections/propertiesAlarmSection/propertyComponent.tsx
```diff
@@ -48,13 +48,13 @@ export const PropertyComponent: FC<PropertyComponentProps> = ({
     <Grid gridDefinition={[{ colspan: 12 }]}>
       <SpaceBetween size="xxxs" direction={'vertical'}>
         <Grid gridDefinition={[{ colspan: 9 }, { colspan: 3 }]} disableGutters>
-          <div className="threshold-content-item with-gutter grow">
-            <div className="threshold-content-item with-gutter">
+          <div className="side-panel-input with-gutter grow">
```
square-li, line 51: rename class name.

packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.tsx
```diff
@@ -24,18 +25,41 @@ export const BaseSettings: FC<{ messageOverrides: DashboardMessages }> = ({
...
+            <span>{baseSettings.x}</span>
+          </div>
+          <div className="side-panel-input with-gutter">
+            <Input value={`${x}`} type="number" onChange={onXChange} data-test-id="base-setting-x-input" />
+          </div>
+          <div className="side-panel-input">
+            <span>{baseSettings.y}</span>
+          </div>
+          <div className="side-panel-input">
+            <Input value={`${y}`} type="number" onChange={onYChange} data-test-id="base-setting-y-input" />
```
mpopv, line 40: No need to interpolate value here, I think

packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.tsx
```diff
@@ -24,18 +25,41 @@ export const BaseSettings: FC<{ messageOverrides: DashboardMessages }> = ({
...
+          </div>
+          <div className="side-panel-input with-gutter">
+            <Input value={`${width}`} type="number" onChange={onWidthChange} data-test-id="base-setting-width-input" />
+          </div>
+          <div className="side-panel-input">
+            <span>{baseSettings.height}</span>
+          </div>
+          <div className="side-panel-input">
+            <Input
+              value={`${height}`}
```
mpopv, line 55: Same here

packages/dashboard/src/components/sidePanel/sections/legendSettingSection/index.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { render } from '@testing-library/react';
+
+const widget: Widget = {
+  ...MOCK_KPI_WIDGET,
+};
+
+const state: Partial<DashboardState> = {
+  dashboardConfiguration: {
+    widgets: [widget],
+    viewport: { duration: '5m' },
```
mpopv, line 18: I wish we could make this a default, I feel like I see it everywhere and it's never really necessary to make explicit

mpopv: (No need to do anything to respond to this, just an observation)

packages/dashboard/src/components/sidePanel/sections/legendSettingSection/index.tsx
mpopv: I think we can improve on the pattern of blowing away every legend setting with `setLegendSettings(undefined)` below and then having to do this awkward thing where we 'remember' the previous settings on the client to prevent unexpected deletion of settings when a customer clicks the show legend toggle twice.

Why don't we just include a `shown` boolean in the legend settings object? We can completely delete `showLegend` and `recentLegendSettings` here because everything we need is always encoded in what we're dispatching to Redux.

mpopv (changes requested):

square-li:
close for upcoming refactoring.

PR #529, opened 2023-02-02
feat(dashboard): resource explorer simpler data fetching and alarms

mpopv:
## Overview

* Refactors data fetching when navigating to an asset's pane to a single call to DescribeAsset, to support data for Alarms. This required a cascading set of changes to associated components like searchbar.
* Adds Alarms in Resource Explorer. The drag infrastructure is partially in place because currently we don't have everything set up on the widget side to receive dropped alarm data; when that's added I will open a follow-up PR.
* Updates tests appropriately.

## Legal


packages/dashboard/src/components/resourceExplorer/components/panel.tsx
```diff
@@ -32,69 +41,99 @@ const PanelAssetPropertyDragHandle = ({ item }: { item: ExtendedPanelAssetSummar
...
-    {
-      id: 'variable',
-      header: '',
-      maxWidth: '100%',
-      cell: (item: ExtendedPanelAssetSummary) => <PanelCell item={item} />,
-    },
-  ];
+  const hasChildAssets = !!asset?.children?.length;
+  const hasProperties = !!asset?.assetProperties?.length;
+  const hasAlarms = !!asset?.assetCompositeModels?.length;
```
NorbertNader, line 92: the presence of `assetCompositeModels` doesn't necessarily mean these composite models are alarms. Here is a util that helps you get the alarm source property https://github.com/awslabs/iot-app-kit/blob/main/packages/source-iotsitewise/src/alarms/iotevents/util/getAlarmSourceProperty.ts

From there you can extract the id that you will need to describe the alarm model. You will need that to get the name of the alarm. Just like it is done here https://github.com/awslabs/iot-app-kit/blob/main/packages/source-iotsitewise/src/alarms/iotevents/siteWiseAlarmModule.ts#L53

so we should do a check like 
```
  assetModel.assetModelCompositeModels
    ?.filter(({ type }) => type === 'AWS/ALARM')
```

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
```diff
@@ -32,69 +41,99 @@ const PanelAssetPropertyDragHandle = ({ item }: { item: ExtendedPanelAssetSummar
...
+      ) : (
+        <></>
+      )}
+
+      {hasAlarms ? (
+        <details open>
+          <summary>
+            <Header>Alarms</Header>
+          </summary>
+          {asset?.assetCompositeModels?.map(({ id, name }) => (
```
NorbertNader, line 130: This id will not be the id of the alarm state property id. That is the `AWS/ALARM_STATE` and what we will need for the query.

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
mpopv: Got it, thanks!

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
mpopv: Thanks, that's helpful.

PR #532, opened 2023-02-02
chore(react-components): Add storybook and sample story book for kpi

diehbria:
## Overview
Settings up storybook for use in react-comopnents.

Unable to use latest version due to an earlier version of webpack tied to components which gets pulled in, which causes the following error on build:

> Error: NormalModuleFactory.beforeResolve (IgnorePlugin) is no longer a waterfall hook, but a bailing hook instead. Do not return the passed object, but modify it instead. Returning false will ignore the request and results in no module created.

## Legal


square-li (approved):
LGTM

PR #534, opened 2023-02-03
chore(react-components): port more KPI and associated utils from synchro-charts

diehbria:
This is part of a series of PRs to port of KPI, status, dial and gauge from SynchroCharts to IoT App Kit.

There are still more code to port over, including integ and unit tests.

## Legal


packages/react-components/src/kpi/kpiBase.tsx
```diff
@@ -0,0 +1,65 @@
...
+        ) : (
+          <>
+            <div
+              data-testid="current-value"
+              className="value-wrapper"
+              style={{ color: valueColor || DEFAULT_FONT_COLOR }}
+            >
+              <Value value={point ? point.y : undefined} unit={stream.unit} />
+            </div>
+            {point && <>at {new Date(point.x).toLocaleString()}</>}
```
boweihan, line 59: this seems like something that should be localized / override-able. I know you're just porting things over with this PR though.

boweihan (approved):

packages/react-components/src/kpi/kpiBase.tsx
diehbria: I actually want to propose to remove the `at` all together, and instead just display a full timestamp (instead of 'to locale string')

NorbertNader (approved):

PR #536, opened 2023-02-06
chore(react-components): create useTimeSeriesDataFromViewport hook an…

diehbria:
### Overview
This PR primarily focuses around getting an end to end, working KPI that queries from IoT App Kit data sources.

Screen Shot 2023-02-06 at 8 42 06 AM

- Revise `KPIBase`, the visual component of the KPI
- Creates the `KPI` component that binds to iot app kit data source
- Creates `widgetPropertiesFromInputs` util, which will be used for:
  - KPI
  - Status
  - Dial
  - Gauge
  - Other single query visualizations in the future? 
- create `useTimeSeriesDataFromViewport` hook to easily bind to IoT App Kit. Establishing patterns for all future IoT App Kit components
- reduce warning count to 41
- allow unused function args that begin with `_` which is a common pattern for destructuring data when using `rest` and `spread` operations
- port over more synchro-charts utils
- Remove contains operator ability to interpret array data as this is not a feature we wish to support at the current moment.

### Notes
- Scheduled time with UX to go over KPI alterations before anything is final
- More work on tests to be done. Some tests are intentionally skipped, as they currently fail due to the work not being completed.

## Legal


package.json
```diff
@@ -33,7 +33,7 @@
     "fix:stylelint": "stylelint '**/*.css' --fix",
     "test": "npm-run-all -p test:unit test:eslint test:stylelint test:git",
     "test:integration": "npm run test:integration -ws --if-present",
-    "test:eslint": "eslint --ext .js,.ts,.tsx . --max-warnings=44",
+    "test:eslint": "eslint --ext .js,.ts,.tsx . --max-warnings=41",
```
jmbuss, line 36: nice :D

packages/react-components/src/utils/number.ts
```diff
@@ -0,0 +1,30 @@
...
+const MAX_PRECISION = 4;
+
+/**
+ * Rounds a number to a pre-determined precision
+ *
+ * i.e. round(100000.1234) => 100000.1234
+ *      round(100000.12345678) => 100000.1234
+ *      round(.02345678) => 0.02346
+ */
+export const round = (num: number): number => {
```
jmbuss, line 12: Do we not want to use lodash round here?

jmbuss (approved):

packages/react-components/src/utils/number.ts
diehbria: willing to re-evaluate that when we tackle customization of decimal precision

PR #541, opened 2023-02-06
#500: NPM Workspaces -> Turborepo

tjuranek:
## Overview
Changes the monorepo from npm workspaces to turborepo. This gives us faster builds and easier maintenance. A dev script was added to each package so we can easily develop across multiple packages at once.

## Legal


diehbria (approved):

diehbria (changes requested):
make sure to squash commits and provide a conventional commit style message

TheEvilDev:
Can we verify this builds with NpmPM before we merge it?

PR #542, opened 2023-02-06
chore(react-components): port over status widget

diehbria:
## Overview
Ports over the core parts of the `<Status />` widget, both the visualization portion of the component, and the bindings to IoT App Kit.

Going to follow up with tests after confirming with UX on all the behaviors/configurations.

Screen Shot 2023-02-06 at 3 52 30 PM

## Legal

### DIFF
--- a/packages/react-components/src/status/status.css
+++ b/packages/react-components/src/status/status.css
@@ -1,0 +1,68 @@
+.status-widget {
+  display: flex;
+  height: 100%;
+  box-sizing: border-box;
+  flex-direction: column;
+  padding: var(--margin-medium);
+  font-size: var(--font-size-3);
+  line-height: var(--line-height-3);
+}
+
+.status-widget .description {
+  font-size: var(--font-size-1);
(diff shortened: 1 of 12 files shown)


corteggiano (approved):

packages/react-components/src/status/status.css
```diff
@@ -0,0 +1,68 @@
+.status-widget {
+  display: flex;
+  height: 100%;
+  box-sizing: border-box;
+  flex-direction: column;
+  padding: var(--margin-medium);
```
square-li, line 6: Do we consider using [@cloudscape/design-token](https://cloudscape.design/foundation/visual-foundation/design-tokens/)? Or a bigger question: how much do we want iot-app-kit to look like cloudscape components?

diehbria: yeah i think that's a good callout - right now the styling is ported from synchro charts but i think thats a good goal to move towards

PR #544, opened 2023-02-07
feat(dashboard): ClientContext and initial resource explorer alarms support

mpopv:
Screenshot 2023-02-07 at 10 46 16 AM

## Overview
This PR:

1. Wraps the dashboard in a ClientContext that allows us to useContext to access the IoTSiteWiseClient passed into the dashboard anywhere. (The Client.getInstance() pattern doesn't seem to work; looking at it, I think we're trying to do something there that requires a React Context -- as written, getInstance() should return undefined when imported in places other than where the Client was instantiated.)

2. Gets alarms with describeAsset and displays them in sidebar. When we are ready to receive dropped alarms on widgets, will be simple to update this to achieve that.

## Legal

### DIFF
--- a/packages/dashboard/src/components/dashboard/index.tsx
+++ b/packages/dashboard/src/components/dashboard/index.tsx
@@ -13,33 +13,40 @@ import { DashboardState, SaveableDashboard } from '../../store/state';
 import { PickRequiredOptional, RecursivePartial } from '../../types';
 import { SiteWiseQuery } from '@iot-app-kit/source-iotsitewise';
 import { DashboardMessages, DefaultDashboardMessages } from '../../messages';
+import { ClientContext } from './clientContext';
 
 import '@cloudscape-design/global-styles/index.css';
 import '../../styles/variables.css';
+import { IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
 
 export type IotDashboardProps = {
   messageOverrides?: RecursivePartial<DashboardMessages>;
   query?: SiteWiseQuery;
   onSave?: (dashboard: SaveableDashboard) => void;
+  client?: IoTSiteWiseClient;
 } & PickRequiredOptional<DashboardState, 'dashboardConfiguration', 'readOnly' | 'grid'>;
 
-const Dashboard: React.FC<IotDashboardProps> = ({ messageOverrides, query, onSave, ...dashboardState }) => (
-  <Provider store={configureDashboardStore({ ...dashboardState })}>
-    <DndProvider
-      backend={TouchBackend}
-      options={{
-        enableMouseEvents: true,
-        enableKeyboardEvents: true,
-      }}
-    >
-      <InternalDashboard
-        query={query}
-        onSave={onSave}
-        messageOverrides={merge(messageOverrides, DefaultDashboardMessages)}
-      />
-      <WebglContext />
-    </DndProvider>
-  </Provider>
-);
+const Dashboard: React.FC<IotDashboardProps> = ({ messageOverrides, query, onSave, client, ...dashboardState }) => {
+  return (
+    <ClientContext.Provider value={client}>
+      <Provider store={configureDashboardStore({ ...dashboardState })}>
+        <DndProvider
+          backend={TouchBackend}
+          options={{
+            enableMouseEvents: true,
+            enableKeyboardEvents: true,
+          }}
+        >
+          <InternalDashboard
+            query={query}
+            onSave={onSave}
+            messageOverrides={merge(messageOverrides, DefaultDashboardMessages)}
+          />
+          <WebglContext />
+        </DndProvider>
+      </Provider>
+    </ClientContext.Provider>
+  );
+};
 
 export default Dashboard;
--- a/packages/dashboard/src/components/dragLayer/itemTypes.ts
+++ b/packages/dashboard/src/components/dragLayer/itemTypes.ts
@@ -2,4 +2,5 @@ export enum ItemTypes {
   Grid = 'Grid',
   Component = 'Component',
   ResourceExplorerAssetProperty = 'ResourceExplorerAssetProperty',
+  ResourceExplorerAlarm = 'ResourceExplorerAlarm',
 }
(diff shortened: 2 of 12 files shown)


packages/dashboard/src/components/resourceExplorer/describeCurrentAsset.ts
```diff
@@ -0,0 +1,50 @@
...
+  assetCreationDate: undefined,
+  assetLastUpdateDate: undefined,
+  assetStatus: undefined,
+  assetDescriptio: undefined,
+} as DescribeAssetResponse;
+
+export const describeCurrentAsset = async (currentBranchId: string) => {
+  let client: IoTSiteWiseClient;
+  try {
+    client = new IoTSiteWiseClient({
```
diehbria, line 23: please remove instantiations of SDK within code, should be passed in. this pattern makes testing hard, doesn't adhere to existing patterns, and throws console warnings in the CI process

diehbria (changes requested):

packages/dashboard/src/components/resourceExplorer/describeCurrentAsset.ts
mpopv: Done!

packages/dashboard/src/components/dashboard/index.tsx
```diff
@@ -13,33 +13,40 @@ import { DashboardState, SaveableDashboard } from '../../store/state';
...
 
 import '@cloudscape-design/global-styles/index.css';
 import '../../styles/variables.css';
+import { IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
 
 export type IotDashboardProps = {
   messageOverrides?: RecursivePartial<DashboardMessages>;
   query?: SiteWiseQuery;
   onSave?: (dashboard: SaveableDashboard) => void;
+  client?: IoTSiteWiseClient;
```
jmbuss, line 26: is it possible to derive query from this client? The 2 are related in some way. I know we are going to clean up the multi client situation eventually, just want to note it

packages/dashboard/src/components/resourceExplorer/describeCurrentAsset.ts
```diff
@@ -0,0 +1,41 @@
...
+  assetLastUpdateDate: undefined,
+  assetStatus: undefined,
+  assetDescription: undefined,
+} as DescribeAssetResponse;
+
+export const sendCommand = (client: IoTSiteWiseClient, assetId: string) =>
+  client.send(new DescribeAssetCommand({ assetId }));
+
+export const describeCurrentAsset = async (currentBranchId: string, client: IoTSiteWiseClient | undefined) => {
+  if (!client) return blank;
```
jmbuss, line 22: I thought this was some new return type for undefined vs null. I was confused for a second 😄

packages/dashboard/src/components/dragLayer/itemTypes.ts
```diff
@@ -2,4 +2,5 @@ export enum ItemTypes {
   Grid = 'Grid',
   Component = 'Component',
   ResourceExplorerAssetProperty = 'ResourceExplorerAssetProperty',
+  ResourceExplorerAlarm = 'ResourceExplorerAlarm',
```
jmbuss, line 5: is this being used? I don't see any other reference to it

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
```diff
@@ -58,17 +60,19 @@ export const IotResourceExplorerPanel: React.FC<IotResourceExplorerPanelProps> =
   };
 
   const PanelCell = ({ item }: { item: ExtendedPanelAssetSummary }) => {
+    const isAlarm = !!(item?.type === 'AWS/ALARM');
```
jmbuss, line 63: should this be some function we have tests around? I'm sure the type is pretty stable, but in a case where we want to extend other string values to be an alarm we'd want to have a place to define it

jmbuss (changes requested):

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
jmbuss: I think we are using it multiple times in the component also?

packages/dashboard/src/components/dashboard/index.tsx
mpopv: Yes, I think we need to do that next. Just trying to add the minimum required right now to source the client from being passed in, per Brian's comment.

packages/dashboard/src/components/dragLayer/itemTypes.ts
mpopv: It's part of the scaffolding we'll need to drop alarms. I could remove it, but we'll just immediately add it back in.

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
mpopv: I talked to Norbert about this and he said this is the best way to verify that it's an alarm, because all alarms have this. So my understanding is this is a reliable indicator.

If we want to allow external alarms, that will require a much larger refactor, so I didn't try to attack that problem here.

packages/dashboard/src/components/dashboard/index.tsx
```diff
@@ -13,33 +13,40 @@ import { DashboardState, SaveableDashboard } from '../../store/state';
...
-        onSave={onSave}
-        messageOverrides={merge(messageOverrides, DefaultDashboardMessages)}
-      />
-      <WebglContext />
-    </DndProvider>
-  </Provider>
-);
+const Dashboard: React.FC<IotDashboardProps> = ({ messageOverrides, query, onSave, client, ...dashboardState }) => {
+  return (
+    <ClientContext.Provider value={client}>
```
square-li, line 31: currently `describeAsset` uses this object(`packages/dashboard/src/util/IotSitewiseClient.ts`) to get a client instance. Using context is fine. But we need to update `describeAsset` as well.

packages/dashboard/src/components/resourceExplorer/components/panel.tsx
mpopv: You're right, I can abstract it.

mpopv: Added abstracted function.

packages/dashboard/src/components/dashboard/index.tsx
mpopv: Agreed, but I think we can do this next.

square-li (approved):

jmbuss (approved):

PR #547, opened 2023-02-08
fix: build issues

NorbertNader:
## Overview

This will bump version to `3.0.0-alpha.2`

Fixes build issues:
1.  `✘ [ERROR] Could not resolve "./iconx.svg"` - Instead inline the svg in the component since it is only used there. Alternatives involved rollup svg plugin but seems unnecessary for just one svg.
2. `✘ [ERROR] [plugin vite:dep-pre-bundle] Missing "./jsx-runtime.js" export in "react" package` - this was related to react-dnd, fixed by bumping to 16.0.1
3. Typo in Dashboard export

## Legal


diehbria:
twin maker team had some blockers around react 18, will this effect them?

NorbertNader:
Thanks for brining this up @diehbria ! Reached out to TM team. I'll follow up once I get an answer.

NorbertNader:
Was able to test this with react 17 and it works fine. Going to hold of on upgrading react to 18.

PR #560, opened 2023-02-09
refactor(dashboard): add DashboardMessageContext and useMessage hook.

square-li:
## Overview
Previously we have to do prop drill to pass messageOverride to child components. Mid level components have to have the messageOverrides object to provide the object to children.
```
<ParentComponent messageOverrides={messageOverrides}>
  <..many layers messageOverrides={messageOverrides}>
    <ChildrenComponent messageOverrides={messageOverrides}>
  <.../>
</ParentComponent>
```

Now we can avoid the prop drill by using this hook.
```
<DashboardMessageContext.Provider value={messageOverrides}>
  <ParentComponent >
    <..many layers >
      <ChildrenComponent>
    <.../>
  </ParentComponent>
</DashboardMessageContext.Provider/>

const ChildrenComponent = ()=> {
  const someLabel = useMessage(message=>message.settings.subsettings)
  return <div> {someLabel} </div>
} 


```

## Legal

### DIFF
--- a/packages/dashboard/src/components/dashboard/index.tsx
+++ b/packages/dashboard/src/components/dashboard/index.tsx
@@ -12,7 +12,7 @@ import { configureDashboardStore } from '../../store';
 import { DashboardState, SaveableDashboard } from '../../store/state';
 import { PickRequiredOptional, RecursivePartial } from '../../types';
 import { SiteWiseQuery } from '@iot-app-kit/source-iotsitewise';
-import { DashboardMessages, DefaultDashboardMessages } from '../../messages';
+import { DashboardMessageContext, DashboardMessages, DefaultDashboardMessages } from '../../messages';
 import { ClientContext } from './clientContext';
 
 import '@cloudscape-design/global-styles/index.css';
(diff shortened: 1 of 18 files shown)


mpopv (approved):

packages/dashboard/src/components/dashboard/index.tsx
```diff
@@ -29,22 +29,24 @@ export type IotDashboardProps = {
...
+            backend={TouchBackend}
+            options={{
+              enableMouseEvents: true,
+              enableKeyboardEvents: true,
+            }}
+          >
+            <InternalDashboard
+              query={query}
+              onSave={onSave}
+              messageOverrides={merge(messageOverrides, DefaultDashboardMessages)}
```
NorbertNader, line 44: We shouldn't need this anymore right?

packages/dashboard/src/messages.ts
```diff
@@ -275,3 +276,10 @@ export const keyboardShortcuts = {
   bringToFront: ']',
   sendToBack: '[',
 };
+
+export const DashboardMessageContext = createContext<DashboardMessages>(DefaultDashboardMessages);
+type useMessageType<M extends DashboardMessages = DashboardMessages> = <T>(getMessage: (message: M) => T) => T;
+export const useMessage: useMessageType = (getMessage) => {
+  const messages = useContext(DashboardMessageContext);
+  return getMessage(messages);
```
NorbertNader, line 284: Can't we just return the context? Why do we need this callback?

NorbertNader (approved):

PR #562, opened 2023-02-09
Upgrade from NPM Workspaces to Turborepo

tjuranek:
## Overview
Changes the monorepo tooling from npm workspaces to turborepo. Adds dev scripts to each package for cross package collaboration.

## Legal

### DIFF
--- a/examples/react-app/package.json
+++ b/examples/react-app/package.json
@@ -28,7 +28,6 @@
   "scripts": {
     "start": "GENERATE_SOURCEMAP=false react-app-rewired start",
     "build": "react-app-rewired build",
-    "test": "react-app-rewired test",
     "eject": "react-scripts eject",
     "link": "./link.sh",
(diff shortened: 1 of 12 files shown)


examples/react-app/package.json
```diff
@@ -28,7 +28,6 @@
   "scripts": {
     "start": "GENERATE_SOURCEMAP=false react-app-rewired start",
     "build": "react-app-rewired build",
-    "test": "react-app-rewired test",
```
diehbria, line 31: should this be removed?

tjuranek: Yes, it's on purpose. There aren't any tests for the example package.

`turbo run test` will run `test` on every npm project under `/examples` or `/packages`. This was an easy way to avoid hard coding a filter at the root level.

diehbria (approved):

PR #563, opened 2023-02-10
feat(dashboard): P0 resource explorer

mpopv:
## Overview
Rebuilds resource explorer using Stencil component.

## Legal


packages/dashboard/src/components/resourceExplorer/index.test.tsx
```diff
@@ -1,55 +0,0 @@
-import * as React from 'react';
```
mpopv, line 1: Deleting this as it's noisy and (with the tested component unused right now) not currently adding value. Before re-adding the tested component it should be broken down into more testable constituent parts.

PR #580, opened 2023-02-13
chore(dial): Port over dial from synchro charts

diehbria:
## Overview
Ports over dial component from synchrocharts. Also refactors to remove calls to calculate bounding boxes and eliminate a lot of state mutations, in preference for a simpler, more straightforward implementation

Follow up PR will be made which will add icons, tests, etc. This is the first iteration, and does not expose the dial component publicly. 
## Legal

### DIFF
--- a/packages/react-components/src/components/dial/dialSvg.tsx
+++ b/packages/react-components/src/components/dial/dialSvg.tsx
@@ -1,0 +1,85 @@
+import React from 'react';
+
+import { DialSettings } from './types';
+import { ColorConfigurations } from './constants';
+import { useArcs } from './useArcs';
+
+const NO_VALUE_PRESENT = '-';
+
+const RADIUS = 138;
+const UNIT_SPACE = 4;
+
+export const DialSvg = ({
+  percent,
+  value,
+  color,
+  label,
+  unit,
+  isLoading,
+  settings,
+}: {
+  percent: number;
+  isLoading: boolean;
+  color: string; // hex color string
+  value?: string;
+  label?: string;
+  unit?: string;
+  settings?: DialSettings;
+}) => {
+  const { defaultRing, colorRing } = useArcs({ percent, lineThickness: settings.dialThickness, radius: RADIUS });
+  const displayLabel = label != null && label != '' && !isLoading;
+  const displayValue = !isLoading && value != null && value !== '';
+
+  const valueLength = (value != null ? value.length : 0) + (unit != null ? unit.length : 0);
+  const valueFontSize =
+    value != null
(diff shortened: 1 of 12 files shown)


packages/react-components/src/components/dial/dialSvg.tsx
```diff
@@ -0,0 +1,85 @@
...
+              <tspan fontSize={Math.min(settings.unitFontSize, valueFontSize)} dx={UNIT_SPACE}>
+                {unit}
+              </tspan>
+            )}
+          </tspan>
+        </text>
+      )}
+      {isLoading && (
+        <text y='55%' x='50%' textAnchor='middle' fontWeight='bold' fontSize={settings.fontSize}>
+          Loading
```
NorbertNader, line 63: shouldn't this be internationalized?

diehbria: This PR is incremental, but yet loading will be internationalized. I want to pull over the `useMessageOverrides` hook as a part of iot app kit API and rethink how we are handling internationalization from the perspective of component contributors

packages/react-components/src/components/dial/useArcs.ts
```diff
@@ -0,0 +1,54 @@
...
+export const useArcs = ({
+  percent,
+  radius,
+  lineThickness,
+}: {
+  percent: number;
+  radius: number;
+  lineThickness: number;
+}) => {
+  const [colorRing, setColorRing] = useState('');
```
NorbertNader, line 35: inconsistent types, here it is string but on line 49 it is `DefaultArcObject`

packages/react-components/src/components/dial/useArcs.ts
```diff
@@ -0,0 +1,54 @@
+import { DefaultArcObject, arc } from 'd3-shape';
```
NorbertNader, line 1: Can we add tests for this?

packages/react-components/src/components/dial/dialSvg.tsx
```diff
@@ -0,0 +1,85 @@
...
+          x='50%'
+          textAnchor='middle'
+          fontWeight='bold'
+          fontSize={settings.fontSize}
+          fill={ColorConfigurations.SECONDARY_TEXT}
+        >
+          {NO_VALUE_PRESENT}
+        </text>
+      )}
+      {displayLabel && (
```
NorbertNader, line 78: I don't think this is tested?

diehbria: will add one more unit test in `DialBase`

diehbria: added missing test on the case within `DialBase` that completes this coverate

PR #581, opened 2023-02-13
feat(dashboard): add input widet configuration

NorbertNader:
## Overview
Add configuration settings for input widget

https://user-images.githubusercontent.com/6397726/218612141-fc36eb0e-1386-4b44-bbbb-f17d408e47f2.mov

## Legal

### DIFF
--- a/packages/dashboard/src/components/sidePanel/index.tsx
+++ b/packages/dashboard/src/components/sidePanel/index.tsx
@@ -6,6 +6,7 @@ import { DashboardMessages } from '~/messages';
 import { AppKitComponentTags } from '~/types';
 import TextSettings from './sections/textSettingSection/text';
 import LinkSettings from './sections/textSettingSection/link';
+import InputSettings from './sections/inputSettingsSection';
 import { BaseSettings } from './sections/baseSettingSection';
 import AxisSetting from './sections/axisSettingSection';
 import ThresholdsSection from './sections/thresholdsSection/thresholdsSection';
@@ -20,6 +21,7 @@ const SidePanel: FC<{ messageOverrides: DashboardMessages }> = ({ messageOverrid
   const selectedWidget = selectedWidgets[0];
   const isAppKitWidget = AppKitComponentTags.find((tag) => tag === selectedWidget.componentTag);
   const isTextWidget = selectedWidget.componentTag === 'text';
+  const isInputWidget = selectedWidget.componentTag === 'input';
 
   return (
     <div className='iot-side-panel'>
@@ -28,6 +30,7 @@ const SidePanel: FC<{ messageOverrides: DashboardMessages }> = ({ messageOverrid
         <BaseSettings messageOverrides={messageOverrides} />
         {isTextWidget && <TextSettings messageOverride={messageOverrides} />}
         {isTextWidget && <LinkSettings messageOverride={messageOverrides} />}
+        {isInputWidget && <InputSettings messageOverride={messageOverrides} />}
         {isAppKitWidget && (
           <>
             <PropertiesAlarmsSection messageOverrides={messageOverrides} />
--- a/packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.spec.tsx
+++ b/packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.spec.tsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import createWrapper from '@cloudscape-design/components/test-utils/dom';
+import { Provider } from 'react-redux';
+import InputSettings from './index';
+import { InputWidget, RecursivePartial } from '~/types';
+import { DashboardState } from '~/store/state';
+import { DefaultDashboardMessages, DashboardMessages } from '~/messages';
+import { configureDashboardStore } from '~/store';
+import { MOCK_INPUT_WIDGET } from '../../../../../testing/mocks';
+import { expect } from '@jest/globals';
+
+const widget: InputWidget = { ...MOCK_INPUT_WIDGET };
+
+const state: Partial<DashboardState> = {
+  dashboardConfiguration: {
+    widgets: [widget],
+    viewport: { duration: '5m' },
+  },
+  selectedWidgets: [widget],
+};
+
+const TestComponent = (
+  {
+    stateOverride = state,
+    messageOverride = DefaultDashboardMessages,
+  }: {
+    stateOverride?: RecursivePartial<DashboardState>;
+    messageOverride?: DashboardMessages;
+  } = { stateOverride: state, messageOverride: DefaultDashboardMessages }
+) => (
+  <Provider store={configureDashboardStore(stateOverride)}>
+    <InputSettings messageOverride={messageOverride} />
+  </Provider>
+);
+
+it('renders without options from state', () => {
+  const props = { stateOverride: {} };
+  const { container } = render(<TestComponent {...props} />);
+  const inputSettings = createWrapper(container);
+  const options = inputSettings.findTokenGroup('[data-test-id="input-widget-token-list"]');
+
+  expect(options.findTokens().length).toBe(0);
+});
+
+it('renders with options from state', () => {
+  const { container } = render(<TestComponent />);
+  const inputSettings = createWrapper(container);
+  const options = inputSettings.findTokenGroup('[data-test-id="input-widget-token-list"]');
+
+  MOCK_INPUT_WIDGET.options.forEach(({ label }) => {
+    expect(options.getElement()).toHaveTextContent(label);
+  });
+});
+
+it('can add option', () => {
+  const { container } = render(<TestComponent />);
+  const inputSettings = createWrapper(container);
+  const addOptionButton = inputSettings.findButton('[data-test-id="input-widget-add-option-btn"]');
+  const optionInput = inputSettings.findInput('[data-test-id="input-widget-option-input"]');
+  const options = inputSettings.findTokenGroup('[data-test-id="input-widget-token-list"]');
+  const newOption = 'lorem ipsum';
+
+  expect(addOptionButton.isDisabled()).toBeTruthy();
+  expect(options.findTokens().length).toBe(3);
+  expect(options.getElement()).not.toHaveTextContent(newOption);
+
+  optionInput.setInputValue(newOption);
+
+  expect(addOptionButton.isDisabled()).toBeFalsy();
+  addOptionButton.click();
+
+  expect(options.findTokens().length).toBe(4);
+  expect(options.getElement()).toHaveTextContent(newOption);
+});
+
+it('can remove option', () => {
+  const { container } = render(<TestComponent />);
+  const inputSettings = createWrapper(container);
+  const options = inputSettings.findTokenGroup('[data-test-id="input-widget-token-list"]');
+
+  expect(options.findTokens().length).toBe(3);
+
+  options.findToken(2).findDismiss().click();
+
+  expect(options.findTokens().length).toBe(2);
+  expect(options.findToken(1).getElement()).toHaveTextContent(widget.options[0].label);
+  expect(options.findToken(2).getElement()).toHaveTextContent(widget.options[2].label);
+});
+
+it('correctly renders translations', () => {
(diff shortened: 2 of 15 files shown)


packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+  selectedWidgets: [widget],
+};
+
+const TestComponent = () => (
+  <Provider store={configureDashboardStore(state)}>
+    <InputSettings messageOverride={DefaultDashboardMessages} />
+  </Provider>
+);
+
+it('renders', () => {
```
diehbria, line 28: these tests don't give too much confidence that the component actually works. This is more or less a snapshot test with some indirection via testIds

A better test would be that as a user, i can add an option, and then see that newly added option reflected in the dashboard definition.

I would advise to avoid adding testid's on every element.

diehbria (changes requested):

packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+          onChange={addOption}
+          data-test-id="input-widget-option-input"
+        />
+        <Button disabled={!label} onClick={saveOption} data-test-id="input-widget-add-option-btn">
+          {inputSettings.add}
+        </Button>
+      </Grid>
+      <TokenGroup
+        data-test-id="input-widget-token-list"
+        alignment={'vertical'}
```
diehbria, line 56: nit: alignment="verticle"

packages/dashboard/src/components/widgets/primitives/input/index.tsx
```diff
@@ -1,22 +1,32 @@
...
 import SpaceBetween from '@cloudscape-design/components/space-between';
 import { InputWidget as InputWidgetType } from '../../../../types';
 
 export type InputWidgetProps = InputWidgetType & { readOnly: boolean };
 
 const Input: React.FC<InputWidgetProps> = ({ readOnly, ...widget }) => {
-  const [selectedOption, setSelectedOption] = useState<SelectProps.Option>(widget.options[0]);
+  const [selectedOption, setSelectedOption] = useState<SelectProps.Option | null>(widget.options[0]);
+  const options = widget.options.map(({ label }) => ({ label, value: label }));
+  const disabled = !readOnly || options.length === 0;
```
diehbria, line 12: when will it be readonly?

packages/dashboard/src/components/widgets/primitives/input/index.tsx
```diff
@@ -1,22 +1,32 @@
...
+  useEffect(() => {
+    if (options.length === 0) {
+      setSelectedOption(null);
+    } else if (!selectedOption || options.filter(({ label }) => selectedOption.label === label).length === 0) {
+      setSelectedOption(widget.options[0]);
+    }
+  }, [widget.options]);
 
   return (
     <SpaceBetween size={'xs'} direction={'horizontal'}>
```
diehbria, line 27: avoid `{'string'}`, just "string" instead.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
```diff
@@ -1,22 +1,32 @@
...
-      <Select
-        disabled={!readOnly}
-        selectedOption={selectedOption}
-        onChange={({ detail }) => setSelectedOption(detail.selectedOption)}
-        options={widget.options}
-      />
-      <Button disabled={!readOnly}>{widget.messageOverrides?.submitLabel}</Button>
+      <Select disabled={disabled} selectedOption={selectedOption} onChange={changeOption} options={options} />
+      <Button disabled={disabled}>{widget.messageOverrides?.submitLabel}</Button>
```
diehbria, line 29: should submitLabel really sometime be undefined? feels like it should be required after merging in defaults.

packages/dashboard/src/messages.ts
```diff
@@ -136,6 +136,12 @@ export type SidePanelMessages = {
     containsLabel: string;
     thresholdPlaceHolder: string;
   };
+
+  inputSettings: {
+    title: string;
+    add: string;
```
diehbria, line 142: add?

packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.spec.tsx
```diff
@@ -0,0 +1,45 @@
+import { InputWidget } from '../../../../types';
```
tjuranek, line 1: Suggestion: Use `paths` in this package. https://www.typescriptlang.org/tsconfig#paths

Would let us `import  { DashboardState } from '~/store/state';`

packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.spec.tsx
tjuranek: I agree on testing component functionality, but would agree with Norbert on his use of test ids. 

I'd want to see `.findButton('[data-test-id="input-widget-add-option-btn"]').click()` instead of only truthy assertions, unless we're checking that something rendered only.

Also, defining an enum or object to hold the testing ids - that's scoped to the component - would be an easier read.

packages/dashboard/src/components/sidePanel/utils/index.ts
```diff
@@ -46,4 +46,5 @@ export const useInput: <T>(path: string, validator?: (newValue: T) => boolean) =
 
 // this helps Typescript finding correct types given attribute names
```
tjuranek, line 47: Is there some implementation we're missing here with generics? 

```js
export const useInput<T> = useInput<T>;
```

packages/dashboard/src/components/widgets/primitives/input/index.tsx
tjuranek: I'm guessing this was to appease eslint accessibility warnings?

tjuranek (commented):
Seems good - other than testing comment just leaving suggestions I wouldn't consider blocking a merge.

packages/dashboard/src/components/sidePanel/index.tsx
```diff
@@ -28,6 +30,7 @@ const SidePanel: FC<{ messageOverrides: DashboardMessages }> = ({ messageOverrid
         <BaseSettings messageOverrides={messageOverrides} />
         {isTextWidget && <TextSettings messageOverride={messageOverrides} />}
         {isTextWidget && <LinkSettings messageOverride={messageOverrides} />}
+        {isInputWidget && <InputSettings messageOverride={messageOverrides} />}
```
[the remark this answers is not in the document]
NorbertNader: Agreed. I like this idea, it is more scalable than what we have been doing so far in the sidePanel. I would say it is out of scope for this PR but we should keep track of this and tackle it at some point.

packages/dashboard/src/components/sidePanel/utils/index.ts
NorbertNader: I am not sure but worth tackling this at some point definitely. Out of scope for this PR though.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: It will be read only when the dashboard is in view mode (no side panels and grid is locked)

NorbertNader: No, nothing to do with eslint. Talked to UX and we do not want to widget to be used in edit mode.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: Agreed, this should not be optional. This is the pattern we have been following for all the widgets so far though. Please look at `AppKitWidget`, `TextWidget` etc. I will make it required for my component but we should fix this for the other widgets as well. I won't include that in my PR but worth tracking.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: This is a pattern we have been following all over the app. Maybe we want to include a prettier/eslint config enforcing this. I would say this is out of the scope of this PR but worth tracking.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
jmbuss: We should be deep merging message overrides at the dashboard component so you shouldn't need to worry about having undefined messaging.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
diehbria: input needs to be usuable in 'view mode', operators need to be able add these notes, even if they don't have access to edit the dashboard.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
diehbria: yes we should update the eslint config

packages/dashboard/src/components/widgets/primitives/input/index.tsx
diehbria: we should fix it in this PR and start enforcing better patterns around message over rides

packages/dashboard/src/components/sidePanel/index.tsx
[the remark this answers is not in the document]
jmbuss: We will tackle the way we render this component when we define the customization api. It will look more like your suggestion @tjuranek than a bunch of if statements. We are still trying to figure out the best way to write that though

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: Looks like this won't be a simple fix since this is broken all over the dashboard.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: Exactly, that is why this code will allow the input widget to be usable in "view mode" and only prevent users from using it in "edit mode"

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: created separate PR for this https://github.com/awslabs/iot-app-kit/pull/584

packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.spec.tsx
NorbertNader: Say no more https://github.com/awslabs/iot-app-kit/pull/586

packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+      },
+    },
+  };
+  const { container } = render(<TestComponent {...props} />);
+  const inputSettings = createWrapper(container);
+  const addOptionButton = inputSettings.findButton('[data-test-id="input-widget-add-option-btn"]');
+  const optionInput = inputSettings.findInput('[data-test-id="input-widget-option-input"]');
+
+  expect(addOptionButton?.findTextRegion()?.getElement().innerHTML).toBe(addOptionLabel);
+  expect(optionInput?.findNativeInput().getElement().placeholder).toBe(optionPlaceholder);
```
diehbria, line 119: instead of searching down the DOM tree, we should be using best practices testing-library where possible. 

For example, there is a `getByPlaceholderText` API, as well as a way to say `get button that says 'lorem ipsum 2' instead of `findTextRegion().getElement().innerHTML.

NorbertNader: I am using the cloudescape testing lilbrary for this https://cloudscape.design/get-started/testing/introduction/

diehbria (approved):

PR #582, opened 2023-02-14
chore(dashboard): refactor resize and move gestures into hooks

jmbuss:
## Overview
Refactor resize and move gestures into their own hooks.
Refactor general gesture handling into its own hook.

## Legal

### DIFF
--- a/packages/dashboard/src/components/internalDashboard/gestures/index.spec.tsx
+++ b/packages/dashboard/src/components/internalDashboard/gestures/index.spec.tsx
@@ -1,0 +1,180 @@
+import React from 'react';
+import { renderHook, act } from '@testing-library/react-hooks';
+import { Provider } from 'react-redux';
+
+import { configureDashboardStore } from '../../../store';
+import { RecursivePartial } from '../../../types';
+import { DashboardState } from '../../../store/state';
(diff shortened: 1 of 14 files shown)


diehbria (approved):

packages/dashboard/src/components/internalDashboard/gestures/index.spec.tsx
```diff
@@ -0,0 +1,180 @@
...
+const TestProvider: React.FC<{
+  storeArgs?: RecursivePartial<DashboardState>;
+}> = ({ storeArgs, children }) => <Provider store={configureDashboardStore(storeArgs)}>{children}</Provider>;
+
+const widgetGestureAttribute = gestureable('widget');
+const resizeGestureAttribute = gestureable('resize');
+const anchorAttribute = anchorable('bottom');
+const idAttribute = idable('widget-1');
+
+const mockWidget = document.createElement('div');
```
diehbria, line 22: would be preferable to just construct these via JSX rather than imperatively across all tests

jmbuss: That is the way I initially wrote the test, but I was running into an issue where getAttribute which is important for determining the gesture was undefined.

PR #590, opened 2023-02-16
chore(dial): port over DialBase from SynchroCharts

diehbria:
## Overview
Ports over dial component from synchrocharts. Also refactors to remove calls to calculate bounding boxes and eliminate a lot of state mutations, in preference for a simpler, more straightforward implementation

Follow up PR will be made which will add icons, tests, etc. This is the first iteration, and does not expose the dial component publicly.

## Legal

### DIFF
--- a/packages/react-components/src/components/dial/dial.tsx
+++ b/packages/react-components/src/components/dial/dial.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import { DialBase } from './dialBase';
+import { TimeQuery, TimeSeriesData, TimeSeriesDataRequest, StyleSettingsMap, Viewport } from '@iot-app-kit/core';
+import { useTimeSeriesData } from '../../hooks/useTimeSeriesData';
+import { widgetPropertiesFromInputs } from '../../common/widgetPropertiesFromInputs';
+import { Annotations } from '../../common/thresholdTypes';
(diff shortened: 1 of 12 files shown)


NorbertNader (approved):

packages/react-components/src/components/dial/dial.tsx
```diff
@@ -0,0 +1,49 @@
...
+}) => {
+  const { dataStreams } = useTimeSeriesData({
+    viewport: passedInViewport,
+    query,
+    settings: { fetchMostRecentBeforeEnd: true },
+    styles,
+  });
+  const { viewport } = useViewport();
+
+  const utilizedViewport = passedInViewport || viewport; // explicitly passed in viewport overrides viewport group
```
jmbuss, line 31: Should an override be a part of the viewport hook? I imagine every component is going to have this line?

jmbuss (approved):

packages/react-components/src/components/dial/dial.tsx
diehbria: I'm not so sure, while your right it'll be a common pattern, it is possibly more intuitive for people to just handle this themselves then have to understand that a viewport passed into useViewport overrides the viewport and makes it so use viewport does nothing

PR #591, opened 2023-02-16
Dedupe Batch Requests

tjuranek:
## Overview
Before sending off a batch of requests, requests are filtered for uniqueness by the request information property.

## Legal


diehbria (changes requested):
awesome! just two comments:

1. update commit to utilize conventional commit
2. add test that would have previously failed that now passes

packages/source-iotsitewise/src/time-series-data/client/batchGetAggregatedPropertyDataPoints.ts
```diff
@@ -50,13 +51,15 @@ const sendRequest = ({
   // callback cache makes it convenient to capture request data in a closure.
   // the cache exposes methods that only require batch response entry as an argument.
   const callbackCache: BatchEntryCallbackCache = {};
-
   const batchSize = calculateNextBatchSize({ maxResults, dataPointsFetched });
 
+  // filter out duplicate requests
+  const uniqueRequests = uniqBy(batch, (entry) => entry.requestInformation);
```
diehbria, line 57: i think we may also want to omit a couple fields, such as `refId`, `cacheSettings`

PR #592, opened 2023-02-16
Testing Ground Example App

tjuranek:
## Overview
Creates a new /examples application that makes testing cross package development easier. By not requiring testing ground to be served by the stencil dev server we see reduced build times and more reliable reloading when dependent packages change.

## Legal

### DIFF
--- a/examples/testing-ground/src/index.tsx
+++ b/examples/testing-ground/src/index.tsx
@@ -1,0 +1,77 @@
+import { ScatterChart } from '@iot-app-kit/react-components';
+import { LineChart, WebglContext } from '@iot-app-kit/react-components';
+import { initialize } from '@iot-app-kit/source-iotsitewise';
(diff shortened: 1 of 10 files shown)


examples/testing-ground/src/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { environment } from './utils/environment';
+
+const { query } = initialize({
+  awsCredentials: {
+    accessKeyId: environment.AWS_ACCESS_KEY_ID,
+    secretAccessKey: environment.AWS_SECRET_ACCESS_KEY,
+    sessionToken: environment.AWS_SESSION_TOKEN,
+  },
+  awsRegion: 'us-west-2',
+  settings: {
```
diehbria, line 14: I think we should be able to omit this (if not, I'll add an issue to make it so we can)

PR #594, opened 2023-02-16
fix(dashboard): fix layering of widgets with selection and context menu

jmbuss:
## Overview
Add fix for layering bug where you can raise a group of widgets to overlay above the context menu, selection rectangle, and user selection rectangle. 

Adds tests for the context menu.

## Legal

### DIFF
--- a/packages/dashboard/src/components/contextMenu/index.test.tsx
+++ b/packages/dashboard/src/components/contextMenu/index.test.tsx
@@ -1,0 +1,322 @@
+import * as React from 'react';
+import { Provider } from 'react-redux';
+
+import { act } from 'react-dom/test-utils';
+import { screen } from '@testing-library/dom';
+import { render } from '@testing-library/react';
+import UserEvent from '@testing-library/user-event';
+
+import ContextMenu, { ContextMenuProps } from './index';
+import { DefaultDashboardMessages } from '~/messages';
+import { DASHBOARD_CONTAINER_ID } from '../grid/getDashboardPosition';
+import { configureDashboardStore } from '~/store';
+
+const renderContextMenu = (args: ContextMenuProps, open?: boolean) => {
+  const ret: { container: HTMLElement | null } = {
+    container: null,
+  };
+
+  act(() => {
+    const { container } = render(
+      <Provider store={configureDashboardStore()}>
+        <div id={DASHBOARD_CONTAINER_ID} style={{ height: '1000px', width: '1000px' }}>
+          <ContextMenu {...args} />
+        </div>
+      </Provider>
+    );
+    ret.container = container;
+  });
+
+  if (open) {
+    act(() => {
+      const ev = new MouseEvent('contextmenu', {
+        bubbles: true,
+        clientX: 10,
+        clientY: 10,
+      });
+      document.getElementById(DASHBOARD_CONTAINER_ID)?.dispatchEvent(ev);
+    });
+  }
+
+  return ret;
+};
+
+const clickOption = async (option: string) => {
+  const user = UserEvent.setup();
+  await act(async () => {
+    await user.click(screen.getByText(option));
+  });
+};
+
+describe('ContextMenu', () => {
+  it('can toggle', () => {
+    const args: ContextMenuProps = {
+      messageOverrides: DefaultDashboardMessages,
+      hasCopiedWidgets: false,
+      hasSelectedWidgets: false,
+      copyWidgets: () => {},
+      pasteWidgets: () => {},
+      deleteWidgets: () => {},
+      bringWidgetsToFront: () => {},
+      sendWidgetsToBack: () => {},
+    };
+
+    const { container } = renderContextMenu(args);
+    if (!container) {
+      throw new Error('Container not set');
+    }
+
+    expect(container.querySelector('.iot-context-menu')).toBeFalsy();
+
+    act(() => {
+      const ev = new MouseEvent('contextmenu', {
+        bubbles: true,
+        clientX: 10,
+        clientY: 10,
+      });
+      document.getElementById(DASHBOARD_CONTAINER_ID)?.dispatchEvent(ev);
+    });
+
+    expect(container.querySelector('.iot-context-menu')).toBeTruthy();
+  });
+
+  it('can copy', async () => {
+    const copyWidgets = jest.fn();
+
+    const args: ContextMenuProps = {
+      messageOverrides: DefaultDashboardMessages,
+      hasCopiedWidgets: false,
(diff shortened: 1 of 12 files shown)


packages/dashboard/src/components/contextMenu/index.test.tsx
```diff
@@ -0,0 +1,378 @@
...
+const clickOption = async (option: string) => {
+  const user = UserEvent.setup();
+  await act(async () => {
+    await user.click(screen.getByText(option));
+  });
+};
+
+describe('ContextMenu', () => {
+  it('can toggle', () => {
+    const copyWidgets = jest.fn();
```
NorbertNader, line 63: not: if we are not spying on these can we just use arrow functions and skip creating a const for these and inline then in the `args` like `copyWidgets: () => {}`.

I noticed this is copied for other tests so maybe worth just creating using `jest.fn()` for the one we want to spy on.

packages/dashboard/src/components/contextMenu/index.test.tsx
```diff
@@ -0,0 +1,378 @@
...
+import { configureDashboardStore } from '~/store';
+
+const renderContextMenu = (args: ContextMenuProps, open?: boolean) => {
+  const ret: { container: HTMLElement | null } = {
+    container: null,
+  };
+
+  act(() => {
+    const { container } = render(
+      <Provider store={configureDashboardStore()}>
```
NorbertNader, line 23: Why do we need to wrap this in the store provider and dnd provider?

packages/dashboard/src/components/contextMenu/index.test.tsx
```diff
@@ -0,0 +1,378 @@
+import * as React from 'react';
```
NorbertNader, line 1: These tests are only testing that a function gets called but don't actually test the functionality. From a users perspective I think we would benefit more from writing integration tests for these.

packages/dashboard/src/store/actions/bringToFront/index.ts
```diff
@@ -17,8 +18,16 @@ export const onBringWidgetsToFrontAction = (): BringWidgetsToFrontAction => ({
 export const bringWidgetsToFront = (state: DashboardState): DashboardState => {
   const widgets = state.dashboardConfiguration.widgets;
   const slectedWidgets = state.selectedWidgets;
+
+  const unselectedWidgets = xorBy(widgets, slectedWidgets, 'id');
```
NorbertNader, line 22: can't we just check that widgets length is equal to slectedWidgets length?

packages/dashboard/src/store/actions/bringToFront/index.ts
```diff
@@ -17,8 +18,16 @@ export const onBringWidgetsToFrontAction = (): BringWidgetsToFrontAction => ({
 export const bringWidgetsToFront = (state: DashboardState): DashboardState => {
   const widgets = state.dashboardConfiguration.widgets;
   const slectedWidgets = state.selectedWidgets;
```
NorbertNader, line 20: nit: typo in selectedWidgets

packages/dashboard/src/components/contextMenu/index.test.tsx
jmbuss: You're right, we don't, was copied in from another test.

packages/dashboard/src/components/contextMenu/index.test.tsx
jmbuss: Yes we can do that

packages/dashboard/src/store/actions/bringToFront/index.ts
jmbuss: ya, that would work too

packages/dashboard/src/components/contextMenu/index.test.tsx
jmbuss: I think integration tests would be useful, and we can certainly add those, but I disagree that these tests won't catch bugs that arise from the context menu. Also, I don't know if we would want to have to setup the entire dashboard just to ensure that the context menu has the right buttons and that those buttons are clickable under the right circumstances. I think a test like that would end up failing more because we changed the api of the dashboard and didn't update our test setup accordingly, than failing because we actually made a breaking change to the context menu.

I think what might be a better solution though is to refactor the context menu component to use the redux actions directly like I did with the keyboard shortcuts instead of passing them in as a function. Then I can move the jest mocks onto those redux actions and we can still test this component in isolation.

packages/dashboard/src/store/actions/bringToFront/index.ts
jmbuss: Oh actually we can't do that for this scenario. This list is also used to figure out the target z index for which selected widgets needs to be moved before / after. Previously we were just comparing against all widgets which would artificially inflate / defalte the zindex depending on the selection.

packages/dashboard/src/components/contextMenu/index.test.tsx
NorbertNader: I see your point. I agree with you integration tests that would just test that a button is on the page would not make sense. My point was to test that e.g. when we click on `copyWidgets` that we actually see the widget copied on the dashboard in the store.

NorbertNader (approved):

packages/dashboard/src/components/contextMenu/index.test.tsx
jmbuss: What do you think about doing this as an e2e test instead? I was thinking after we add playwright we could create a test which covers a bunch of these features. Thinking of a test like - drag a widget onto the dashboard, copy / paste with context menu, drag move it over, and delete with keyboard shortcut. That would cover gestures and actions in 1 shot.

NorbertNader: I love that idea.

packages/dashboard/src/store/actions/bringToFront/index.ts
NorbertNader: Oh I see, my bad. Thanks for clarifying!

packages/dashboard/src/store/actions/bringToFront/index.ts
jmbuss: fixed the typo :D

packages/dashboard/src/store/actions/bringToFront/index.ts
jmbuss: Haha I was like I thought there was a reason I did this! I forgot until I looked at the code again.

PR #595, opened 2023-02-17
feat(composer): Implement data overlay container

sheilaXu:
## Overview
Implement the container of data overlay component with simple text content rendering

overlay-container

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayComponent.tsx
+++ b/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayComponent.tsx
@@ -1,0 +1,32 @@
+import React, { ReactElement, useContext } from 'react';
+import { Html } from '@react-three/drei';
+
+import { ISceneNodeInternal } from '../../../store';
+import { sceneComposerIdContext } from '../../../common/sceneComposerIdContext';
+import { IDataOverlayComponentInternal } from '../../../store/internalInterfaces';
+
+import { DataOverlayContainer } from './DataOverlayContainer';
+
+export interface DataOverlayComponentProps {
+  node: ISceneNodeInternal;
+  component: IDataOverlayComponentInternal;
+}
+
+export const DataOverlayComponent = ({ node, component }: DataOverlayComponentProps): ReactElement => {
+  const sceneComposerId = useContext(sceneComposerIdContext);
+
+  return (
+    <group>
+      <Html
+        className='html-wrapper'
+        style={{
+          transform: 'translate3d(-50%,-100%,0)', // make the center of 3D transform the middle of bottom edge
+        }}
+      >
+        <sceneComposerIdContext.Provider value={sceneComposerId}>
+          <DataOverlayContainer node={node} component={component} />
+        </sceneComposerIdContext.Provider>
+      </Html>
+    </group>
+  );
+};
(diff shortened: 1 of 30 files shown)


packages/scene-composer/src/hooks/useCallbackWhenNotPanning.ts
```diff
@@ -16,7 +16,7 @@ const useCallbackWhenNotPanning = (callback, deps, acceptableDriftDistance = 1)
   const lastPointerDownLocation = useRef<[number, number] | null>(null);
 
   const onPointerUp = useCallback(
-    (e: ThreeEvent<MouseEvent>) => {
+    (e: ThreeEvent<MouseEvent> | PointerEvent) => {
```
jwills-jdubs, line 19: Not sure if it is necessary right now but should we use `ThreeEvent<PointerEvent>` vs `ThreeEvent<MouseEvent>`? Not your change just thinking out loud

jwills-jdubs (approved):

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { IDataOverlayComponentInternal } from '../../../store/internalInterfaces';
+
+import { DataOverlayContainer } from './DataOverlayContainer';
+
+export interface DataOverlayComponentProps {
+  node: ISceneNodeInternal;
+  component: IDataOverlayComponentInternal;
+}
+
+export function DataOverlayComponent({ node, component }: DataOverlayComponentProps): ReactElement {
```
mumanity, line 15: Why are you using the traditional `function` expression rather than implementing arrow functions? Are there plans to use scope as part of how these functions execute in the future? As of now, there are not any usages that would call for traditional functions so I'd encourage changing these to use arrows in order to clearly delineate the limitations.

mumanity (approved):

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayContainer.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { useCallback, useContext, useEffect, useRef, useState } from 'react';
+import * as awsui from '@awsui/design-tokens';
```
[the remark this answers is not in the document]
sheilaXu: will remove. forgot to remove it when I removed the inline styles. 
importing everything as `awsui` seems to be the pattern the awsui team recommended since this pattern is what they have in their development guideline

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayComponent.tsx
sheilaXu: will update. I was doing some other testing and didn't change it back

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayContainer.tsx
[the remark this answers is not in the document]
mumanity: Mass imports like that tend to lead to bloat & we have a lot of those in our package. It would probably benefit us to start getting more specific as to what we want. Not a major thing, but something to consider.

mumanity (approved):

PR #598, opened 2023-02-18
feat(dashboard): input widget async/error handling

NorbertNader:
## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/dashboard/src/components/widgets/primitives/input/index.tsx
```diff
@@ -1,17 +1,119 @@
...
+  const showSubmitAndOptions = options.length > 0 && !infoMessage && widget.assets?.[0]?.properties?.length;
   const isSelectedOptionValid = options.filter(({ label }) => selectedOption?.label === label).length > 0;
 
+  // SiteWise specific - in future we might want to add support for other data sources
+  const [propertyName, setPropertyName] = useState<string>();
+  const updateAssetPropertyValue = useUpdateAssetPropertyValue();
+  const describeAssetProperty = useDescribeAssetProperty();
+  const isWritable = useWritableSiteWiseProperty();
+
+  // For now we will only support one SiteWise Asset Property
```
diehbria, line 30: one of the great things about hooks is you don't need to write them all inlined in your component in the same way that you would for class based components (due to needing to have all the lifecycle methods contained in the class itself). can greatly increase readability by separating out the hook, to give more seperation between the visual component, and the hundred line hooks

packages/dashboard/testing/iotsitewiseSDK.ts
```diff
@@ -0,0 +1,57 @@
...
+export type mockSiteWiseSDKProps = {
+  describeAsset?: (input: DescribeAssetCommandInput) => Promise<DescribeAssetResponse>;
+  describeAssetProperty?: (input: DescribeAssetPropertyCommandInput) => Promise<DescribeAssetPropertyResponse>;
+  listTags?: (input: ListTagsForResourceCommandInput) => Promise<ListTagsForResourceResponse>;
+  putAssetPropertyValue?: (
+    input: BatchPutAssetPropertyValueCommandInput
+  ) => Promise<BatchPutAssetPropertyValueResponse>;
+};
+
+type inputProp =
```
diehbria, line 24: types should be capitalized

packages/dashboard/testing/iotsitewiseSDK.ts
```diff
@@ -0,0 +1,57 @@
...
+  ListTagsForResourceCommandInput,
+  ListTagsForResourceResponse,
+  BatchPutAssetPropertyValueCommandInput,
+  BatchPutAssetPropertyValueResponse,
+  IoTSiteWiseClient,
+} from '@aws-sdk/client-iotsitewise';
+
+const nonOverriddenMock = () => Promise.reject(new Error('Mock method not override.'));
+
+export type mockSiteWiseSDKProps = {
```
diehbria, line 15: types should be capitalized

packages/dashboard/testing/mocks.ts
```diff
@@ -147,3 +156,72 @@ export const dashboardConfig: DashboardConfiguration = {
...
+  id: '20d0c2bd-d4f7-4642-93ab-1aec642b645d',
+  name: 'LOGSTREAM',
+  notification: {
+    state: 'DISABLED',
+    topic:
+      '$aws/sitewise/asset-models/5ab18893-c34c-4b28-bd04-ca4489b8b8f5/assets/62d5e3ed-9f27-4fd7-aaea-1ba1e0a3a33b/properties/20d0c2bd-d4f7-4642-93ab-1aec642b645d',
+  },
+};
+
+export const NON_STRING_ASSET_PROPERTY = {
```
diehbria, line 172: should contain types

packages/dashboard/src/components/widgets/primitives/input/useUpdateAssetPropertyValue.ts
```diff
@@ -0,0 +1,38 @@
+import { BatchPutAssetPropertyValueCommand, BatchPutAssetPropertyValueResponse } from '@aws-sdk/client-iotsitewise';
+import { useContext } from 'react';
+import { ClientContext } from '~/components/dashboard/clientContext';
+
+const useUpdateAssetPropertyValue = () => {
+  const client = useContext(ClientContext);
+
+  return (
+    assetId: string,
+    propertyId: string,
```
diehbria, line 10: better to use named params so that the function is called as ({ assetId: 'windmi', ...}) instead of ('some string', 'a', 'b') which can make it easy to have the wrong params.

also in the future, this needs to support call via property alias. with your pattern, that would look like (undefined, undefined, propertyAlias) rather than just ({ propertAlias })

diehbria (changes requested):

packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.tsx
```diff
@@ -20,9 +20,19 @@ const InputSettings: FC<InputComponentProps> = ({ messageOverride }) => {
   const {
     sidePanel: { inputSettings },
   } = messageOverride;
-  const [options, setOptions] = useInputWidgetInput('options');
+  // TODO: Add type support for nested paths in useInput
+  // eslint-disable-next-line
+  // @ts-ignore
+  const useInputOptions = useInputWidgetInput('properties.options');
```
diehbria, line 26: a bit confusing to refer to the return of a hook as 'use___' as the use prefix is reserved for the hook itself, not it's return statement. The more standard way to writ this would be `const [options, setOptions] = usesInputWidgetInput...`

packages/dashboard/src/components/sidePanel/utils/index.ts
```diff
@@ -5,6 +5,11 @@ import { DashboardState } from '~/store/state';
 import { onUpdateWidgetsAction } from '~/store/actions';
 import { AppKitWidget, TextWidget, InputWidget, Widget } from '~/types';
 
+// // TODO: this should work for all widgets so we can access nested paths e.g. 'properties.options'
```
diehbria, line 8: i wonder if a better way to approach this is to allow a function which selects what it wants to alter, instead of a path.

packages/dashboard/src/components/sidePanel/sections/inputSettingsSection/index.tsx
NorbertNader: Good callout. That was my first approach yet with nested props that throws type errors. Let me try to find to find a better way of doing this.

packages/dashboard/src/components/sidePanel/utils/index.ts
NorbertNader: I think you are right. We'd have to refactor `useInput` to not use `lodash.get`

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: Definitely.

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
```diff
@@ -0,0 +1,63 @@
...
+    propertyAlias?: string;
+  }): Promise<{ writable: boolean; error: string | undefined }> => {
+    if (!client) return { writable: false, error: 'missing client' };
+    let asset;
+    let assetTags;
+
+    try {
+      asset = await describeAsset(client, assetId);
+    } catch (e) {
+      return { writable: false, error: (e as ErrorDetails).message };
```
tracy-french, line 37: Non-blocker: Don't cast errors. Use a type guard instead.

packages/dashboard/testing/iotsitewiseSDK.ts
```diff
@@ -0,0 +1,57 @@
...
+        case 'DescribeAssetPropertyCommand':
+          return describeAssetProperty(command.input);
+        case 'ListTagsForResourceCommand':
+          return listTags(command.input);
+        case 'BatchPutAssetPropertyValueCommand':
+          return putAssetPropertyValue(command.input);
+        default:
+          throw new Error(
+            `missing mock implementation for command name ${commandName}. Add a new command within the mock SiteWise SDK.`
+          );
```
tracy-french, line 54: Blocking: Why crash the test runner for this and not just let the tests fail?

packages/dashboard/src/components/widgets/widget.tsx
```diff
@@ -34,14 +36,41 @@ const WidgetComponent: React.FC<WidgetProps> = ({
...
+          const assetId = queryAssetsParam[0]?.assetId;
+          const propertyId = queryAssetsParam[0]?.properties?.[0].propertyId;
+          const { writable } = await isWritable({ assetId, propertyId });
+          if (writable) {
+            setAssets(queryAssetsParam);
+          }
+        } else {
+          setAssets(queryAssetsParam);
+        }
+        setError(null);
```
tracy-french, line 60: Blocking: Abstract into symbol. It's not clear what is happening here at first glace.

packages/dashboard/src/components/widgets/widget.tsx
```diff
@@ -34,14 +36,41 @@ const WidgetComponent: React.FC<WidgetProps> = ({
...
+        if (isInputWidget && monitor.isOver()) {
+          const assetId = queryAssetsParam[0]?.assetId;
+          const propertyId = queryAssetsParam[0]?.properties?.[0].propertyId;
+          const { error } = await isWritable({ assetId, propertyId });
+          if (error) {
+            setError(error);
+          }
+        } else {
+          setError(null);
+        }
```
tracy-french, line 73: Non-blocking: Abstract into symbol.

packages/dashboard/src/components/widgets/widget.spec.tsx
```diff
@@ -0,0 +1,89 @@
...
+  messageOverrides: DefaultDashboardMessages,
+};
+
+it('renders widget', () => {
+  generateWidget(defaultProps);
+
+  expect(screen.queryByText(textWidget.text)).toBeTruthy();
+});
+
+// TODO: backfill test for input widget
```
tracy-french, line 62: Blocking: How does someone uncomment these tests?

packages/dashboard/src/components/widgets/primitives/input/util/isValidQuery.ts
```diff
@@ -0,0 +1,7 @@
+import { AssetQuery } from '@iot-app-kit/core';
+
+// For now we will only support one SiteWise Asset Property
```
tracy-french, line 3: Non-blocking: Add more context. I'm not sure what this means.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
```diff
@@ -1,42 +1,105 @@
...
-  const options = widget.options.map(({ label }) => ({ label, value: label }));
+  const [showConfirmationModal, setShowConfirmationModal] = useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [infoMessage, setInfoMessage] = useState<string | null>(null);
+  const [successMessage, setSuccessMessage] = useState<string | null>(null);
+  const [isSending, setIsSending] = useState<boolean>(false);
+  const options = widget.properties.options.map(({ label }) => ({ label, value: label }));
   const [selectedOption, setSelectedOption] = useState<SelectProps.Option | null>(options[0]);
-  const disabled = !readOnly || options.length === 0;
+  const showSubmitAndOptions = options.length > 0 && !infoMessage && widget.assets?.[0]?.properties?.length;
```
tracy-french, line 20: Non-blocking: Decompose conditional.

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    try {
+      assetTags = await getTags(client, asset.assetArn);
+    } catch (e) {
+      return { writable: false, error: (e as ErrorDetails).message };
+    }
```
tracy-french, line 44: Non-blocking: Use a single try/catch.

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const hasTag = Object.keys(assetTags?.tags ?? {}).includes(WRITABLE_TAG);
+
+    // For now we will only support string properties
+    const isStringProperty = asset.assetProperties?.some(
+      ({ id, alias, dataType }) => (id === propertyId || alias === propertyAlias) && dataType === 'STRING'
+    );
+
+    if (!hasTag) {
+      return { writable: false, error: `Asset is not writable. Add ${WRITABLE_TAG} tag to asset.` };
```
tracy-french, line 54: Blocking: Customers do not know how to do this.

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
```diff
@@ -0,0 +1,63 @@
...
+  IoTSiteWiseClient,
+  ListTagsForResourceCommand,
+  ErrorDetails,
+} from '@aws-sdk/client-iotsitewise';
+import { useContext } from 'react';
+import { ClientContext } from '~/components/dashboard/clientContext';
+
+export const WRITABLE_TAG = 'iot-app-kit:writable';
+
+const useWritableSiteWiseProperty = () => {
```
tracy-french, line 12: Non-blocking: Decompose into sub-routines.

packages/dashboard/src/components/widgets/primitives/input/index.css
```diff
@@ -0,0 +1,9 @@
+.input-widget-container {
+  display: flex;
+  flex-direction: row;
+}
+
+.input-widget-options {
+  flex: 1;
+  margin-right: 20px;
```
tracy-french, line 8: Non-blocking: Remove magic number.

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
```diff
@@ -0,0 +1,63 @@
+import {
+  DescribeAssetCommand,
+  IoTSiteWiseClient,
+  ListTagsForResourceCommand,
+  ErrorDetails,
+} from '@aws-sdk/client-iotsitewise';
+import { useContext } from 'react';
+import { ClientContext } from '~/components/dashboard/clientContext';
+
+export const WRITABLE_TAG = 'iot-app-kit:writable';
```
tracy-french, line 10: Blocking: This is a 1-way door. We should not publicly release this without a principle engineer review and likely a security review.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
```diff
@@ -1,42 +1,105 @@
...
+        if (assetId && propertyId) {
+          const { error } = await isWritable({ assetId, propertyId });
+          if (error) {
+            setErrorMessage(error);
+          }
+        }
+      })();
+    }
+  }, [JSON.stringify(widget.assets), JSON.stringify(options)]);
+
```
tracy-french, line 51: Non-blocking: Reduce conditional branching.

packages/dashboard/src/components/widgets/primitives/input/util/isValidQuery.ts
```diff
@@ -0,0 +1,7 @@
...
+const isValidQuery = (assets: AssetQuery[]) =>
+  assets?.length === 1 && assets?.[0]?.properties?.length && assets?.[0]?.properties?.length === 1;
```
tracy-french, line 5: Non-blocking: Decompose conditional logic.

tracy-french (changes requested):

packages/dashboard/src/components/widgets/primitives/input/hooks/useDescribeAssetProperty.ts
```diff
@@ -0,0 +1,20 @@
...
+  const client = useContext(ClientContext);
+
+  return ({
+    assetId,
+    propertyId,
+  }: {
+    assetId: string;
+    propertyId: string;
+  }): Promise<DescribeAssetPropertyResponse | undefined> => {
+    if (!client) return Promise.resolve(undefined);
```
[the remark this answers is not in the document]
NorbertNader: We won't make the request. In the input widget this means we won't show the name of the property on the confirmation modal and display an error.

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
NorbertNader: Good point. We can create a predicate for this.

packages/dashboard/testing/iotsitewiseSDK.ts
NorbertNader: Good point. I like this error because it will be more descriptive and gives you a human readable explanation of what went wrong and what you can do to fix it.

packages/dashboard/src/components/widgets/widget.tsx
NorbertNader: Could you help me understand what you mean by abstract into symbol? This will just set the error state to null which means we won't show an error message.

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
NorbertNader: Could you help me undertand what you mean by sub-routines here?

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: Could you help me understand which part you are referring to and what your recommendation would be?

packages/dashboard/src/components/widgets/primitives/input/util/isValidQuery.ts
NorbertNader: Could you help me understand what your recommendation would be here?

packages/dashboard/src/components/widgets/primitives/input/hooks/useWritableSiteWiseProperty.ts
NorbertNader: Good callout. Why do you think this is a 1-way door? I'll reach out to app sec to kick off a review for this.

packages/dashboard/src/components/widgets/primitives/input/index.tsx
NorbertNader: Could you help me understand what your recommendation would be?

packages/dashboard/src/components/widgets/primitives/input/hooks/useUpdateAssetPropertyValue.ts
```diff
@@ -0,0 +1,45 @@
...
+    propertyId: string;
+    propertyAlias?: string;
+    value: string;
+  }): Promise<BatchPutAssetPropertyValueResponse | undefined> => {
+    if (!client) return Promise.resolve(undefined);
+    return client.send(
+      new BatchPutAssetPropertyValueCommand({
+        entries: [
+          {
+            entryId: 'inputWidget',
```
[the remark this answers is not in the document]
NorbertNader: I see your point. This is only used once though and I think creating a variable for this is a bit overkill.

packages/dashboard/src/components/widgets/primitives/input/index.css
NorbertNader: I see your point. We only use this once so I feel creating a variable for this might be overkill.

diehbria: Maybe we should use this instead? https://cloudscape.design/components/space-between/?tabId=playground

The dashboard should do it's best to adhere to the cloudscape design system - we should definitely avoid hard coded  one off spaces where we can

NorbertNader: Good point @diehbria. Unfortunately with space between the select will not stretch to the width of the widget.

NorbertNader: What do you say we use a css variable from cloudscape instead?

packages/dashboard/src/components/widgets/primitives/input/hooks/useUpdateAssetPropertyValue.ts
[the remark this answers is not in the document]
NorbertNader: Good point. I chose not create a variable because it’s only used once and not relevent to the code but required by the api. I’ll add a variable for this with comments.

packages/dashboard/src/components/widgets/primitives/input/index.css
diehbria: there must be some cloudscape way to put space inbetween an input and a button? this is a pretty basic use case and is the type of problem cloudscape is trying to solve with their component system.

when we hard code these values, it means the dashboard will not work with features such as 'compact' mode or other themes in the future. layout should utilize cloudscape components as much as possible, and then the design tokens if it's not possible.

diehbria (changes requested):

packages/dashboard/src/components/widgets/primitives/input/ConfirmationModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+    (async () => {
+      if (assetId && propertyId) {
+        try {
+          const property = await describeAssetProperty({ assetId, propertyId });
+          const name = property?.assetProperty?.name;
+          if (name) {
+            setPropertyName(name);
+          }
+        } catch (e) {
+          setErrorMessage((e as ErrorDetails).message ?? null);
```
diehbria, line 45: blocking: when error isn't as expected, it will set as having no error. should fall back to an error message such as 'unknown error'

diehbria (changes requested):

PR #604, opened 2023-02-21
fix(dashboard): remove tailing digits on widget dragging

square-li:
## Overview
uses the same `trimWidgetPosition` function from `moveWidgetAction` to trim position & size numbers.

also add tests for this behaviour.
## Legal

### DIFF
--- a/packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.spec.tsx
+++ b/packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.spec.tsx
@@ -7,9 +7,16 @@ import { DefaultDashboardMessages } from '~/messages';
 import React from 'react';
 import { BaseSettings } from './index';
 import { render } from '@testing-library/react';
+import createWrapper from '@cloudscape-design/components/test-utils/dom';
+import userEvent from '@testing-library/user-event';
+import { expect } from '@jest/globals';
 
 const widget: Widget = {
   ...MOCK_KPI_WIDGET,
+  x: 1.5,
+  y: 2,
+  width: 10.1,
+  height: 10.9,
 };
 
 const state: Partial<DashboardState> = {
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.tsx
```diff
@@ -1,9 +1,10 @@
 import React, { FC } from 'react';
 import { DashboardMessages } from '~/messages';
 import { useInput } from '../../utils';
-import { NonCancelableEventHandler } from '@cloudscape-design/components/internal/events';
```
square-li, line 4: remove using internal types

jmbuss (commented):
LGTM, what happens if someone enters a decimal into the input widget? do we allow that and then trim after they're done typing?

square-li:
> LGTM, what happens if someone enters a decimal into the input widget? do we allow that and then trim after they're done typing?
 
No we don't allow that. All decimal will be trimmed after typing. 
chrome-capture-2023-1-22 (1)

packages/dashboard/src/components/sidePanel/sections/baseSettingSection/index.spec.tsx
```diff
@@ -32,9 +39,28 @@ it('renders', () => {
...
+  expect(wrapper.findInput('[data-test-id="base-setting-height-input"]')?.findNativeInput().getElement().value).toBe(
+    '11'
+  );
+});
+
+it('rounds input on user input', async () => {
+  const wrapper = createWrapper(render(<TestBaseSettingSection />).baseElement);
+  const xInput = wrapper.findInput('[data-test-id="base-setting-x-input"]')!.findNativeInput().getElement();
+  await userEvent.type(xInput!, '2.4'); // current x is 2, this type appends 2 after it.
+  expect(xInput?.value).toBe('22'); // decimal digits on typing will be ignored not rounded.
```
jmbuss, line 65: shouldn't this be just 2? or would it end up being 24?

square-li: Current X for the widget is `2` and `type` event just append new chars after it.  So it is `2` -> `22.4` -> `22`

jmbuss: Oh I see, so the idea is you're attempting to change the x from 2 -> 22.4

I thought it was a scenario trying to change 2 -> 2.4

perfect lgtm

jmbuss (approved):

PR #609, opened 2023-02-24
chore(dashboard): toggle between view and edit mode

corteggiano:
## Overview
Add `readOnly` toggle to switch between edit and preview modes
Add `hasEditPermission` dashboard prop to specify if toggle should be displayed

output

## Legal

### DIFF
--- a/packages/dashboard/src/components/actions/index.tsx
+++ b/packages/dashboard/src/components/actions/index.tsx
@@ -1,26 +1,35 @@
 import React from 'react';
+import { useDispatch } from 'react-redux';
 
 import { Button } from '@cloudscape-design/components';
 
 import { DashboardMessages } from '~/messages';
 import { DashboardState, SaveableDashboard } from '~/store/state';
+import { onToggleReadOnly } from '~/store/actions';
 
 export type ActionsProps = {
-  onSave: (dashboard: SaveableDashboard) => void;
   messageOverrides: DashboardMessages;
   grid: DashboardState['grid'];
+  readOnly: boolean;
+  hasEditPermission: boolean;
   dashboardConfiguration: DashboardState['dashboardConfiguration'];
   assetsDescriptionMap: DashboardState['assetsDescriptionMap'];
+  onSave?: (dashboard: SaveableDashboard) => void;
 };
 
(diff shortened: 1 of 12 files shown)


packages/dashboard/src/components/actions/index.tsx
```diff
@@ -29,12 +38,21 @@ const Actions: React.FC<ActionsProps> = ({
...
   return (
     <div className='actions'>
       <h1 className='iot-dashboard-toolbar-title'>{messageOverrides.toolbar.actions.title}</h1>
-      <Button variant='primary' onClick={handleOnSave}>
+      {onSave && <Button variant='primary' onClick={handleOnSave}>
         {messageOverrides.toolbar.actions.save}
-      </Button>
+      </Button>}
+      {hasEditPermission && <Button variant='primary' onClick={handleOnReadOnly}>
+        {readOnly ? 'Edit mode' : 'Preview mode'}
```
jmbuss, line 54: What do you think about just using the words Edit / Preview?

jmbuss (commented):
Looks great! is it possible to style the actions to all be on the same line? I don't want to take more space away from the grid.

packages/dashboard/src/components/actions/index.tsx
jmbuss: I think this would help not take up too much space if we inline the buttons

packages/dashboard/src/components/internalDashboard/index.test.tsx
```diff
@@ -124,4 +129,86 @@ describe('InternalDashboard', () => {
...
+          </DndProvider>
+        </Provider>,
+        container
+      );
+    });
+
+    const actionsContainer = container.querySelector('.button-actions');
+    expect(actionsContainer).toBeFalsy();
+
+    const dashboard = container.querySelector('.iot-dashboard');
```
jmbuss, line 211: I tihnk this class wraps both the readonly and edit versions of the dashboard. I was using `.iot-dashboard-panes-area` for this type of check since that component is only used in the edit version, it wraps the resizeable panes

jmbuss (dismissed):

jmbuss (approved):

PR #617, opened 2023-03-01
feat(composer): data overlay editor & render markdown

sheilaXu:
## Overview
- Implement data overlay editor for text annotation
- use `react-markdown` to render markdown
- in order to keep bindings order unchanged during editing, updated data binding field in overlay component interface from 
```
    valueDataBindings: {
      [key: string]: {
        valueDataBinding?: ValueDataBinding;
      };
    };
```
to
```
    valueDataBindings: {
      bindingName: string;
      valueDataBinding?: ValueDataBinding;
    }[];
```

overlay-editor

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.spec.tsx
+++ b/packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.spec.tsx
@@ -1,0 +1,74 @@
+import React from 'react';
+import { act, render } from '@testing-library/react';
+import wrapper from '@awsui/components-react/test-utils/dom';
+
+import { KnownComponentType } from '../../../interfaces';
+import { IDataOverlayComponentInternal, ISceneNodeInternal, useStore } from '../../../store';
+import { Component } from '../../../models/SceneModels';
+import { mockProvider } from '../../../../tests/components/panels/scene-components/MockComponents';
+
+import { DataOverlayComponentEditor } from './DataOverlayComponentEditor';
+
+jest.mock('@awsui/components-react', () => ({
+  ...jest.requireActual('@awsui/components-react'),
+}));
+
+jest.mock('./data-overlay/DataBindingMapEditor', () => {
+  const originalModule = jest.requireActual('./data-overlay/DataBindingMapEditor');
+  return {
+    ...originalModule,
+    DataBindingMapEditor: (...props: unknown[]) => {
+      return <div data-testid='DataBindingMapEditor'>{JSON.stringify(props)}</div>;
+    },
+  };
+});
+
+describe('DataBindingMapEditor', () => {
+  const component: IDataOverlayComponentInternal = {
+    ref: 'comp-ref',
+    type: KnownComponentType.DataOverlay,
+    subType: Component.DataOverlaySubType.TextAnnotation,
+    dataRows: [
+      {
+        rowType: Component.DataOverlayRowType.Markdown,
+        content: 'markdown-content',
+      },
+    ],
+    valueDataBindings: [],
+  };
+  const node = {
+    ref: 'node-ref',
+  } as ISceneNodeInternal;
+  const updateComponentInternalMock = jest.fn();
+
+  const baseState = {
+    updateComponentInternal: updateComponentInternalMock,
+    getEditorConfig: jest.fn().mockReturnValue({ valueDataBindingProvider: mockProvider }),
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should add new binding by clicking add button', async () => {
+    useStore('default').setState(baseState);
+
+    const { container } = render(<DataOverlayComponentEditor node={node} component={component} />);
+    const polarisWrapper = wrapper(container);
+
+    const textarea = polarisWrapper.findTextarea();
+    expect(textarea).not.toBeNull();
+
+    const newContent = 'new content';
+    act(() => {
+      textarea!.setTextareaValue(newContent);
+    });
+
+    expect(updateComponentInternalMock).toBeCalledTimes(1);
+    expect(updateComponentInternalMock).toBeCalledWith(
+      node.ref,
+      { ref: component.ref, dataRows: [{ ...component.dataRows[0], content: newContent }] },
+      undefined,
+    );
+  });
+});
--- a/packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
+++ b/packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
@@ -1,0 +1,64 @@
+import React, { useCallback, useContext } from 'react';
+import { FormField, SpaceBetween, Textarea } from '@awsui/components-react';
+import { useIntl } from 'react-intl';
+
+import { IComponentEditorProps } from '../ComponentEditor';
+import { IDataOverlayComponentInternal, ISceneComponentInternal, useStore } from '../../../store';
+import { sceneComposerIdContext } from '../../../common/sceneComposerIdContext';
+import { Component } from '../../../models/SceneModels';
+
+import { DataBindingMapEditor } from './data-overlay/DataBindingMapEditor';
+
+export interface IDataOverlayComponentEditorProps extends IComponentEditorProps {
+  component: IDataOverlayComponentInternal;
+}
+
+export const DataOverlayComponentEditor: React.FC<IDataOverlayComponentEditorProps> = ({
+  node,
+  component,
+}: IDataOverlayComponentEditorProps) => {
+  const sceneComposerId = useContext(sceneComposerIdContext);
+  const updateComponentInternal = useStore(sceneComposerId)((state) => state.updateComponentInternal);
+  const valueDataBindingProvider = useStore(sceneComposerId)(
+    (state) => state.getEditorConfig().valueDataBindingProvider,
+  );
+  const { formatMessage } = useIntl();
+
+  const onUpdateCallback = useCallback(
+    (componentPartial: Partial<IDataOverlayComponentInternal>, replace?: boolean) => {
+      const componentPartialWithRef = { ref: component.ref, ...componentPartial };
+      updateComponentInternal(node.ref, componentPartialWithRef as ISceneComponentInternal, replace);
+    },
+    [node.ref, component.ref],
+  );
+
+  return (
+    <SpaceBetween size='s'>
+      <DataBindingMapEditor
+        valueDataBindingProvider={valueDataBindingProvider}
+        component={component}
+        onUpdateCallback={onUpdateCallback}
+      />
+
+      {component.dataRows.length > 0 &&
+        component.dataRows.map(
+          (row, index) =>
+            row.rowType === Component.DataOverlayRowType.Markdown && (
+              <FormField
+                key={index}
+                label={formatMessage({ defaultMessage: 'Markdown content', description: 'FormField label' })}
+              >
+                <Textarea
+                  value={row.content}
+                  onChange={(e) => {
+                    const newRows = [...component.dataRows];
+                    newRows[index] = { ...newRows[index], content: e.detail.value };
+                    onUpdateCallback({ dataRows: newRows });
+                  }}
+                />
+              </FormField>
+            ),
+        )}
+    </SpaceBetween>
+  );
+};
--- a/packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.spec.tsx
+++ b/packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.spec.tsx
@@ -1,0 +1,138 @@
+import React from 'react';
+import { act, render } from '@testing-library/react';
+import wrapper from '@awsui/components-react/test-utils/dom';
+
+import { mockProvider } from '../../../../../tests/components/panels/scene-components/MockComponents';
+import { IDataOverlayComponentInternal } from '../../../../store';
+import { Component } from '../../../../models/SceneModels';
+import { IValueDataBindingBuilderProps } from '../ValueDataBindingBuilder';
+
+import { DataBindingMapEditor } from './DataBindingMapEditor';
+
+jest.mock('@awsui/components-react', () => ({
+  ...jest.requireActual('@awsui/components-react'),
+}));
+
+jest.mock('../../../../utils/mathUtils', () => ({
+  generateUUID: jest.fn(() => 'random-uuid'),
+}));
+
+let builderOnChangeCb;
+jest.mock('../ValueDataBindingBuilder', () => {
+  const originalModule = jest.requireActual('../ValueDataBindingBuilder');
+  return {
+    ...originalModule,
+    ValueDataBindingBuilder: (props: IValueDataBindingBuilderProps) => {
+      builderOnChangeCb = props.onChange;
+      return <div data-testid='ValueDataBindingBuilder'>{JSON.stringify(props)}</div>;
+    },
+  };
+});
+
+describe('DataBindingMapEditor', () => {
+  const component = {
+    subType: Component.DataOverlaySubType.OverlayPanel,
+    valueDataBindings: [
+      {
+        bindingName: 'binding-1',
+        valueDataBinding: {
+          dataBindingContext: 'random-1',
+        },
+      },
+    ],
+  } as unknown as IDataOverlayComponentInternal;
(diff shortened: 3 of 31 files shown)


jwills-jdubs (approved):

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { KnownComponentType } from '../../../interfaces';
+import { IDataOverlayComponentInternal, ISceneNodeInternal, useStore } from '../../../store';
+import { Component } from '../../../models/SceneModels';
+import { mockProvider } from '../../../../tests/components/panels/scene-components/MockComponents';
+
+import { DataOverlayComponentEditor } from './DataOverlayComponentEditor';
+
+jest.mock('@awsui/components-react', () => ({
+  ...jest.requireActual('@awsui/components-react'),
+}));
```
TheEvilDev, line 14: We shouldn't be "unmocking" polaris, this is a temporary hack that allows old tests to continue functioning until they migrate off.

We shouldn't be including polaris functionality in our tests.

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const updateComponentInternalMock = jest.fn();
+
+  const baseState = {
+    updateComponentInternal: updateComponentInternalMock,
+    getEditorConfig: jest.fn().mockReturnValue({ valueDataBindingProvider: mockProvider }),
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
```
TheEvilDev, line 51: This shouldn't need to be explicitly called in most circumstances, it's a code smell that results from sharing objects between tests.

Each test should setup it's own behaviors that we care about validating (things we don't care can be mocked at the root, but we won't be making assertions on them). For example, a much cleaner way to test the update behavior in the component below is to isolate the update callback.

You can do this by mocking `useCallback` and grabbing the callback passed to it, and just executing it directly. You don't need to traverse the DOM at all or care about the DOM structure to validate it.

Then you add a second test that validates the DOM structure via a snpahsot, and we've effectively covered this components full set of responsibilities.

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const DataOverlayComponentEditor: React.FC<IDataOverlayComponentEditorProps> = ({
+  node,
+  component,
+}: IDataOverlayComponentEditorProps) => {
+  const sceneComposerId = useContext(sceneComposerIdContext);
+  const updateComponentInternal = useStore(sceneComposerId)((state) => state.updateComponentInternal);
+  const valueDataBindingProvider = useStore(sceneComposerId)(
+    (state) => state.getEditorConfig().valueDataBindingProvider,
+  );
+  const dataOverlayComponent = component as IDataOverlayComponentInternal;
```
TheEvilDev, line 23: You should define this type in the props interface, so you don't need to cast it here.

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
```diff
@@ -0,0 +1,63 @@
...
+        onUpdateCallback={onUpdateCallback}
+      />
+
+      {dataOverlayComponent.dataRows.length > 0 &&
+        dataOverlayComponent.dataRows.map(
+          (row, index) =>
+            row.rowType === Component.DataOverlayRowType.Markdown && (
+              <FormField
+                key={index}
+                label={intl.formatMessage({ defaultMessage: 'Markdown content', description: 'FormField label' })}
```
TheEvilDev, line 48: We should avoid `.` notation here, and just use object destructuring above:

```
const { formatMessage } = useIntl();
```

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.spec.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+jest.mock('@awsui/components-react', () => ({
+  ...jest.requireActual('@awsui/components-react'),
+}));
+
+jest.mock('../../../../utils/mathUtils', () => ({
+  generateUUID: jest.fn(() => 'random-uuid'),
+}));
+
+let builderOnChangeCb;
```
TheEvilDev, line 19: You don't need to do this. Instead in your mock code do:


```
return {
    ...originalModule,
    ValueDataBindingBuilder: jest.fn((props) => <div data-testid='ValueDataBindingBuilder'>{JSON.stringify(props)</div>);
    }
```

Then you just get it from the standard import:

```
import { ValueDataBindingBuilder as mockValueDataBindingBuilder } from '../ValueDataBindingBuilder'
```

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.spec.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+import { mockProvider } from '../../../../../tests/components/panels/scene-components/MockComponents';
+import { IDataOverlayComponentInternal } from '../../../../store';
+import { Component } from '../../../../models/SceneModels';
+
+import { DataBindingMapEditor } from './DataBindingMapEditor';
+
+jest.mock('@awsui/components-react', () => ({
+  ...jest.requireActual('@awsui/components-react'),
+}));
```
TheEvilDev, line 13: Don't do this, we should only rely on mocks.

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.spec.tsx
```diff
@@ -0,0 +1,160 @@
...
+      />,
+    );
+    const polarisWrapper = wrapper(container);
+
+    const nameInput = polarisWrapper.findInput('[data-test-id="binding-name-input"]');
+    expect(nameInput).not.toBeNull();
+
+    act(() => {
+      nameInput!.setInputValue('new-name');
+    });
```
TheEvilDev, line 153: Separate your responsibilities instead of writing unit tests like integration tests.

You should:
- Verify DOM Structure using a snapshot test. You may want to test a few different property settings depending on what you need
- Verify event handlers/callbacks/effects through mocking, and direct execution.

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.tsx
```diff
@@ -0,0 +1,114 @@
...
+  component: IDataOverlayComponentInternal;
+  onUpdateCallback: (componentPartial: Partial<IDataOverlayComponentInternal>, replace?: boolean | undefined) => void;
+}
+
+export const DataBindingMapEditor: React.FC<IDataBindingMapEditorProps> = ({
+  valueDataBindingProvider,
+  component,
+  onUpdateCallback,
+}) => {
+  const intl = useIntl();
```
TheEvilDev, line 22: Use destructuring for all these:

```
const { formatMessage } = useIntl();
```

And get rid of all the namespacing we have to do in the component.

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { ValueDataBindingBuilder } from '../ValueDataBindingBuilder';
+import { generateUUID } from '../../../../utils/mathUtils';
+
+interface IDataBindingMapEditorProps {
+  valueDataBindingProvider: IValueDataBindingProvider | undefined;
+  component: IDataOverlayComponentInternal;
+  onUpdateCallback: (componentPartial: Partial<IDataOverlayComponentInternal>, replace?: boolean | undefined) => void;
+}
+
+export const DataBindingMapEditor: React.FC<IDataBindingMapEditorProps> = ({
```
TheEvilDev, line 17: This component has too many responsibilities, and should be decomposed into smaller pieces. Use judgement, but pretty much every formfield/input combination can be it's own component responsible for it's own strings and behaviors, and it greatly simplifies testing and code structure.

Just apply single responsibility principle, each component should really only have one reason to change.

packages/scene-composer/src/components/panels/scene-components/data-overlay/__snapshots__/DataBindingMapEditorSnap.spec.tsx.snap
```diff
@@ -0,0 +1,379 @@
...
+    </div>
+    <div
+      data-mocked="Button"
+      data-test-id="add-binding-button"
+    >
+      Add data binding
+    </div>
+  </div>
+</div>
+`;
```
TheEvilDev, line 122: This is exactly my point from above, even with mocks, there are multiple fields with various data embedded, and 120 lines of DOM in the snapshot. This means this test can break for a lot of reasons, adding a new field, changing layout, etc. And not just one test (the snapshot), but all the tests will fail for these reasons.

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
```diff
@@ -12,15 +13,16 @@ export const DataOverlayDataRow = ({ rowData, overlayType }: DataOverlayDataRowP
...
+        <ReactMarkdown
+          skipHtml
           className={`markdown-row ${
             overlayType === Component.DataOverlaySubType.TextAnnotation ? 'annotation-row' : 'panel-row'
           }`}
+          linkTarget='_blank'
         >
           {row.content}
-        </p>
+        </ReactMarkdown>
```
TheEvilDev, line 25: We almost certainly want our own wrapper around this markdown component where we control the styles and structure. The data overlay shouldn't care about our markdown configuration, it should just provide "our markdown input".

packages/scene-composer/tests/components/panels/ComponentEditor.spec.tsx
```diff
@@ -47,6 +48,11 @@ describe('ComponentEditor renders correct component', () => {
     );
     expect(wrapper.find(MotionIndicatorComponentEditor).length).toBe(1);
 
+    wrapper = shallow(
```
TheEvilDev, line 51: Please refactor these tests to using our more modern standards, we should be actively removing our dependencies on enzyme everytime we go.

Also, since you changed this test, please move it to the appropriate location next to the system it's testing as we deprecate the tests folder. This should simplify all the import statements significantly.

TheEvilDev (changes requested):

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
sheilaXu: we will need to either cast here, or cast in `ComponentEditor` when passing into this component. it seems better to cast here so the component editor don't need to import all those interfaces

TheEvilDev: I'm not sure I follow. This components signature isn't enforcing the interface, we're just casting it internally. Is there a reason this isn't on the props?

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapEditor.tsx
TheEvilDev: Please break this one down into smaller pieces.

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
TheEvilDev: Please create a wrapper component just for the MarkdownInput component.

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
sheilaXu: I mean in [componentEditor](https://github.com/awslabs/iot-app-kit/blob/main/packages/scene-composer/src/components/panels/ComponentEditor.tsx#L40) when we pass `component` into the editor, the type of `component` is generic, so we either cast here, or cast before passing into the editor

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+jest.mock('@awsui/components-react', () => ({
+  ...jest.requireActual('@awsui/components-react'),
+}));
+
+jest.mock('./data-overlay/DataBindingMapEditor', () => {
+  const originalModule = jest.requireActual('./data-overlay/DataBindingMapEditor');
+  return {
+    ...originalModule,
+    DataBindingMapEditor: (...props: unknown[]) => {
```
divya-sea, line 20: Why the type is **unknown** here?

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <SpaceBetween size='s'>
+      <DataBindingMapEditor
+        valueDataBindingProvider={valueDataBindingProvider}
+        component={component}
+        onUpdateCallback={onUpdateCallback}
+      />
+
+      {component.dataRows.length > 0 &&
```
divya-sea, line 43: What will it display if the length < 0 on the panel?

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapNameEditor.tsx
```diff
@@ -0,0 +1,54 @@
...
+      newBindings[index] = { ...newBindings[index], bindingName: e.detail.value };
+      onUpdateCallback({ valueDataBindings: newBindings });
+    },
+    [valueDataBindings, onUpdateCallback],
+  );
+
+  return (
+    <FormField
+      errorText={bindingNameError(bindingName)}
+      label={formatMessage({ defaultMessage: 'Binding Name', description: 'FormField label' })}
```
divya-sea, line 49: Should not this be `Binding name`?

packages/scene-composer/src/components/wrappers/ReactMarkdownWrapper.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { FC } from 'react';
+import ReactMarkdown from 'react-markdown';
+
+import './ReactMarkdownWrapper.scss';
+
+export interface ReactMarkdownWrapperProps {
+  content: string;
+  className?: string;
+}
+export const ReactMarkdownWrapper: FC<ReactMarkdownWrapperProps> = ({ content, className }) => {
```
divya-sea, line 10: What's the significance of this component? Is this something can be reused in the future?

packages/scene-composer/src/components/panels/scene-components/data-overlay/DataBindingMapNameEditor.tsx
sheilaXu: this is to be consistent with the other labels for data binding, like "Entity Id", "Component Name" (see the screenshot in description). I guess the "Markdown content" should follow the same pattern.. I will update it in next PR since I will update that component again for next task

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.spec.tsx
sheilaXu: the type of data in it doesn't matter for this test file, so I just put unknown for now

packages/scene-composer/src/components/panels/scene-components/DataOverlayComponentEditor.tsx
sheilaXu: it will just display a close button

divya-sea (approved):

PR #619, opened 2023-03-01
feat(dashboard): add table component support

square-li:
## Overview
add `UpdateTableAssetsAction` action to create props that table widget needs.
image
 

## Legal

### DIFF
--- a/packages/dashboard/package.json
+++ b/packages/dashboard/package.json
@@ -99,6 +99,7 @@
     "@iot-app-kit/core": "^2.6.4",
     "@iot-app-kit/react-components": "^2.6.5",
     "@iot-app-kit/source-iotsitewise": "^2.6.4",
+    "@iot-app-kit/table": "^2.6.4",
     "@popperjs/core": "^2.11.6",
     "@synchro-charts/core": "7.2.0",
     "@synchro-charts/react": "^7.2.0",
--- a/packages/dashboard/src/store/actions/updateTableWidget/index.spec.tsx
+++ b/packages/dashboard/src/store/actions/updateTableWidget/index.spec.tsx
@@ -1,0 +1,154 @@
+import { AppKitWidget, Widget } from '~/types';
+import { DashboardState, initialState } from '~/store/state';
+import { MOCK_KPI_WIDGET } from '../../../../testing/mocks';
+import { onUpdateTableAssets, updateTableAssets } from '~/store/actions/updateTableWidget/index';
+import { AssetQuery } from '@iot-app-kit/core';
+import { DescribeAssetResponse } from '@aws-sdk/client-iotsitewise';
+import { toId } from '@iot-app-kit/source-iotsitewise';
+import { ColumnDefinition, Item } from '@iot-app-kit/table';
+
+const MOCK_TABLE_WIDGET: AppKitWidget = {
+  height: 0,
+  widgetId: 'mock-table-widget',
+  width: 0,
+  x: 0,
+  y: 0,
+  z: 0,
+  id: 'table-widget',
+  componentTag: 'iot-table',
+};
+const MOCK_ASSET_QUERY: AssetQuery[] = [
+  {
+    assetId: 'mock-asset-id-1',
+    properties: [
+      {
+        propertyId: 'mock-property-id-1',
+      },
+      {
+        propertyId: 'mock-property-id-2',
+      },
+    ],
+  },
+  {
+    assetId: 'mock-asset-id-2',
+    properties: [
+      {
+        propertyId: 'mock-property-id-2',
+      },
+      {
+        propertyId: 'mock-property-id-3',
+      },
+    ],
+  },
+];
+
+const MOCK_DESCRIPTION_MAP: DashboardState['assetsDescriptionMap'] = {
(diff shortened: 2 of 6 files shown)


packages/dashboard/src/store/actions/updateTableWidget/index.ts
```diff
@@ -0,0 +1,68 @@
...
+        },
+      };
+      if (!propertyNames.includes(name)) {
+        propertyNames.push(name);
+      }
+    });
+    return item;
+  });
+  const columnDefinitions: ColumnDefinition[] = propertyNames.map((name) => ({ key: name, header: name }));
+  const updateWidget: (widgets: Widget) => Widget = (w) => {
```
tjuranek, line 48: How is this passing? Also I don't think there's a need to be explicit on the type for this?

`const updateWidget = (widgets: Widget[]) => {}` should be fine here.

packages/dashboard/src/store/actions/updateTableWidget/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (w.id === widget.id) {
+      return {
+        ...w,
```
tjuranek, line 51: Nit comment, but I'd avoid single char variables names unless it's absolutely beneficial.

packages/dashboard/src/store/actions/updateTableWidget/index.spec.tsx
```diff
@@ -0,0 +1,154 @@
+import { AppKitWidget, Widget } from '~/types';
+import { DashboardState, initialState } from '~/store/state';
+import { MOCK_KPI_WIDGET } from '../../../../testing/mocks';
```
tjuranek, line 3: There a `~` way to import this?

packages/dashboard/src/store/actions/updateTableWidget/index.spec.tsx
```diff
@@ -0,0 +1,154 @@
...
+    updateTableAssets(
+      state,
+      onUpdateTableAssets({
+        widget: MOCK_KPI_WIDGET,
+        assetQuery: MOCK_ASSET_QUERY,
+      })
+    )
+  ).toEqual(state);
+});
+it('should create item and column definitions', () => {
```
tjuranek, line 140: Should there also be tests for deleting items or columns?

tjuranek (changes requested):

packages/dashboard/package.json
```diff
@@ -99,6 +99,7 @@
     "@iot-app-kit/core": "^2.6.4",
```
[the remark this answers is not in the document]
square-li: Didn't know that. Thanks! Adding it.

packages/dashboard/src/store/actions/updateTableWidget/index.spec.tsx
square-li: Updating

packages/dashboard/src/store/actions/updateTableWidget/index.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export const onUpdateTableAssets = (payload: UpdateTableAssetPayload): UpdateTableAssetsAction => ({
+  type: 'UPDATE_TABLE_ASSET',
+  payload,
+});
+
+export const updateTableAssets = (state: DashboardState, action: UpdateTableAssetsAction): DashboardState => {
+  const { assetQuery, widget } = action.payload;
+  if (widget.componentTag !== 'iot-table') return state;
```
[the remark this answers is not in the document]
square-li: Currently it is hardcoded [here](https://github.com/awslabs/iot-app-kit/blob/63bb953b55e3aa86122fa0b3958fe2742a28629e/packages/dashboard/src/types.ts) as
```
export const AppKitComponentTags = [
  'iot-bar-chart',
  'iot-kpi',
  'iot-line-chart',
  'iot-scatter-chart',
  'iot-status-grid',
  'iot-status-timeline',
  'iot-table',
] as const;
export type AppKitComponentTag = typeof AppKitComponentTags[number];
```
Typescript will warn if you provides a bad tag name.

packages/dashboard/src/store/actions/updateTableWidget/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  type: 'UPDATE_TABLE_ASSET';
+  payload: UpdateTableAssetPayload;
+}
+
+export const onUpdateTableAssets = (payload: UpdateTableAssetPayload): UpdateTableAssetsAction => ({
+  type: 'UPDATE_TABLE_ASSET',
+  payload,
+});
+
+export const updateTableAssets = (state: DashboardState, action: UpdateTableAssetsAction): DashboardState => {
```
jmbuss, line 23: I know this is the pattern we have been doing for other types of widgets with asset queries, but is there any way we can move this logic into the table component? I'm in the process of removing any store actions that don't apply to every widget. I don't want to add onto that work unless we absolutely have to.

square-li:
Closing this PR due to ongoing refectoring. Will open another one targeting to the latest structure.

PR #620, opened 2023-03-01
fix(core, source-iotsitewise)!: remove needed exports 

diehbria:
## Overview
The purpose of this PR is to reduce the overall exposed public API of IoT App Kit. No data-source specific code should be exported as part of core.

BREAKING CHANGE:

- Removes SiteWise specific typescript type definitions from being exported from @iot-app-kit/core
- Removes SiteWise alarm module from exported from @iot-app-kit/source-iotsitewise

## Legal

### DIFF
--- a/packages/core/src/data-module/TimeSeriesDataModule.spec.ts
+++ b/packages/core/src/data-module/TimeSeriesDataModule.spec.ts
@@ -1,7 +1,7 @@
 import flushPromises from 'flush-promises';
 import { DATA_STREAM, DATA_STREAM_INFO } from '../mockWidgetProperties';
-import { createMockSiteWiseDataSource } from '../__mocks__';
-import { DataSource, SiteWiseDataStreamQuery } from './types';
+import { createMockSiteWiseDataSource, MockSiteWiseQuery } from '../__mocks__';
+import { DataSource } from './types';
 import { DataPoint } from '@synchro-charts/core';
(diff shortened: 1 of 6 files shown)


packages/core/src/data-module/TimeSeriesDataModule.spec.ts
```diff
@@ -15,6 +15,7 @@ import Mock = jest.Mock;
 const { EMPTY_CACHE } = caching;
 
 const { propertyId: PROPERTY_ID, assetId: ASSET_ID } = toSiteWiseAssetProperty(DATA_STREAM.id);
+type SiteWiseDataStreamQuery = { assets: { assetId: string, properties: { refId?: string; propertyId: string }[] } [] } & DataStreamQuery;
```
jmbuss, line 18: is there no use for AssetQuery / PropertyQuery anywhere else internally? I know we don't want them to be a part of the public api, but would those still be useful?

diehbria: they are exposed within the source-iotsitewise package

jmbuss: ahh, and we couldn't use them here because it would lead to a circular reference then

diehbria: Also these tests aren't really testing the sitewise data source even though they look similar. If we were to invest more time into it I would make a simplified query format more detached from SiteWise

jmbuss (approved):

PR #658, opened 2023-03-07
core(react-components): setup ui tests for kpi and status

diehbria:
## Overview
- Adds integration tests for status and KPI
- Sets up github action to run playwright tests

## Legal


packages/react-components/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    // baseURL: 'http://localhost:3000',
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
+    baseURL: 'http://localhost:6007/',
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
```
jmbuss, line 47: Should we also configure firefox / safari to make sure any future styling changes work across all?

jmbuss (approved):

packages/react-components/playwright.config.ts
diehbria: will enable other browsers at a later time once we have a pipeline passing with screenshots generated

PR #659, opened 2023-03-07
build: Utilize npm cache to increase CI speed

diehbria:
Utilizing caching to increase CI speed: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

Also splits up the unit tests from E2E so that unit and E2E run in parellel. Because integration tests complete faster, this means we can add a lot more integration tests without impacting the CI completion time.

CI checks won't pass until we edit the repository rules to reflect the updates github action names

## Legal


tracy-french (approved):
LGTM!

PR #661, opened 2023-03-07
feat(dashboard): match Table component behaviours to current version 

square-li:
## Overview
- by default, Table will have 3 columns `Property` `Latest Value` and `Unit` only. So it matches current experiences on SWM.
- enable sorting by default
Screenshot from 2023-03-07 13-10-12

## Legal


packages/dashboard/src/customization/widgets/queryWidget/multiQueryWidget.tsx
```diff
@@ -23,15 +23,17 @@ const MultiQueryWidgetComponent: React.FC<QueryWidget> = ({ children, ...widget
...
+          widget.type === 'iot-table' && !(widget.properties as TableProperties).experimentalMode;
         const updatedWidget = {
           ...widget,
           properties: {
             ...widget.properties,
             queryConfig: {
               ...widget.properties.queryConfig,
               query: {
-                assets: mergeAssetQueries(widget.properties.queryConfig.query?.assets || [], asset),
+                assets: mergeAssetQueries(widget.properties.queryConfig.query?.assets || [], asset, dontMergeQueries),
```
jmbuss, line 36: can we just return current assets plus new asset here instead of adding this flag?

jmbuss (changes requested):

packages/dashboard/src/customization/widgets/types.ts
```diff
@@ -71,6 +72,13 @@ export type BarChartProperties = QueryProperties & {
 export type TableProperties = QueryProperties & {
   thresholdSettings?: ThresholdSettings;
   fontSettings?: ComplexFontSettings;
+  experimentalMode?: boolean;
```
jmbuss, line 75: do we need this property? this will be part of the public api

jmbuss (changes requested):

packages/dashboard/src/customization/widgets/queryWidget/multiQueryWidget.tsx
square-li: removed.

packages/dashboard/src/customization/widgets/types.ts
square-li: removed this flag

jmbuss (approved):

diehbria (changes requested):
looks good! just fix the commit history (rebase instead of merge commit)

packages/dashboard/src/customization/widgets/queryWidget/multiQueryWidget.tsx
```diff
@@ -23,15 +23,17 @@ const MultiQueryWidgetComponent: React.FC<QueryWidget> = ({ children, ...widget
...
         const updatedWidget = {
           ...widget,
           properties: {
             ...widget.properties,
             queryConfig: {
               ...widget.properties.queryConfig,
               query: {
-                assets: mergeAssetQueries(widget.properties.queryConfig.query?.assets || [], asset),
+                assets:
+                  widget.type === 'iot-table'
```
diehbria, line 34: should not have widget specific logic here

square-li:
replaced by another PR https://github.com/awslabs/iot-app-kit/pull/695

PR #664, opened 2023-03-08
fix(composer): fix error with rendering 3D tiles

sheilaXu:
## Overview
- fix error with rendering 3D tiles
- fix local npm install issue

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,14 +10,12 @@
     "node": ">=16.0.0",
     "npm": ">=8.7.0"
   },
-  "workspaces": {
-    "packages": [
-      "examples/*",
-      "packages/*"
-    ]
-  },
+  "workspaces": [
+    "examples/*",
+    "packages/*"
+  ],
   "scripts": {
-    "bootstrap": "npm install -ws && npm run build",
+    "bootstrap": "npm install && npm run build",
     "start": "cd packages/components && npm start",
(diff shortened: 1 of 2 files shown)


diehbria (approved):

package.json
```diff
@@ -10,14 +10,12 @@
...
-      "packages/*"
-    ]
-  },
+  "workspaces": [
+    "examples/*",
+    "packages/*"
+  ],
   "scripts": {
-    "bootstrap": "npm install -ws && npm run build",
+    "bootstrap": "npm install && npm run build",
```
TheEvilDev, line 18: Why is this reverted?

TheEvilDev (approved):

package.json
sheilaXu: `bootstrap` is expected to be the single command we need to run to build everything, so I'm having local build issue when it has the `-ws` but the first `npm install` is not added to bootstrap. but without -ws, I see the build works without issue, and composer storybook can run as well. so we probably don't need the flag anymore. 

I also tried to remove the r3f dependency in the root package.json, and it will cause module not found error, which indicates the root dependencies are also installed

PR #681, opened 2023-03-09
feat(dashboard): add onUpdateWidget middleware for widget handling da…

square-li:
…shboard action by themselves.

## Overview
Add `onUpdateWidget` function to `WidgetRegistrationOptions`. When any action with `widgets` filed and `applyUpdateWidget` flag is dispatched, `onUpdateWidget` will be called. This allows users to handle widget updates in a customized way.
For example, if KPI widget registered as 
```
export const kpiPlugin: DashboardPlugin = {
  install: ({ registerWidget }) => {
    registerWidget<KPIWidget>('iot-kpi', {
      ....
      onUpdateWidget: (widget) => {
        console.log(widget);
        return widget;
      },
    });
  },
};
```
It would log widget states to console on wiget state changes.
chrome-capture-2023-2-9

## Legal

### DIFF
--- a/packages/dashboard/src/customization/widgetPropertiesGeneratorMap.ts
+++ b/packages/dashboard/src/customization/widgetPropertiesGeneratorMap.ts
@@ -1,4 +1,4 @@
-import { AnyWidget } from '../types';
+import { AnyWidget } from '~/types';
 
 /**
  * map of widget type to a generator func to create properties when this widget is dropped into the grid
@@ -11,5 +11,6 @@ export const WidgetPropertiesGeneratorMap: {
     //eslint-disable-next-line
     properties?: () => Record<any, any>;
     initialSize?: Pick<AnyWidget, 'height' | 'width'>;
+    onUpdateWidget?: <W extends AnyWidget>(widget: W) => W;
   };
 } = {};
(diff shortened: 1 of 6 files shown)


packages/dashboard/src/customization/widgetPropertiesGeneratorMap.ts
```diff
@@ -11,5 +12,6 @@ export const WidgetPropertiesGeneratorMap: {
     //eslint-disable-next-line
     properties?: () => Record<any, any>;
     initialSize?: Pick<AnyWidget, 'height' | 'width'>;
+    onUpdateWidget?: (action: DashboardAction) => DashboardAction;
```
jmbuss, line 15: What if we just had a function which takes a widget type and returns a widget type. I don't think a user of the api should have to care about redux

jmbuss (changes requested):

packages/dashboard/src/customization/widgetPropertiesGeneratorMap.ts
jmbuss: This was my idea with the useWidgetActions hook. The only thing it does right now is give you a function to update widget, you just need to call it with your updated widget and it handles dispatching the correct store action.

square-li: updated

PR #698, opened 2023-03-14
feat(TableComponent): replace table annotations API with thresholds

diehbria:
## Overview
Port table component directly into react with an updated API to provide thresholds directly instead of needing to wrap thresholds into the `Annotations` type i.e. `{ y: thresholds }`.

## Summary
    * deprecate table package, consolidate in react-components
    * update table documentation
    * update typescript rules to be stricter
    
    BREAKING CHANGES:
    * remove annotations API to be replaced with thresholds for a more concise API

## Legal


packages/react-components/global.d.ts
```diff
@@ -0,0 +1,5 @@
+import 'jest-extended';
+
+declare global {
+  const IS_REACT_ACT_ENVIRONMENT: boolean;
```
square-li, line 4: Nit: From https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html & https://github.com/testing-library/react-testing-library/issues/509#issuecomment-917989936, we might not need this flag anymore if we have `@testing-library/react` after `13.0.0-alpha.1`. And we are using `^14.0.0` in all packages.

square-li (approved):

packages/react-components/global.d.ts
diehbria: will remove in next PR

PR #700, opened 2023-03-14
chore(Components): expose unstable KPI and unstable status; remove un…

diehbria:
expose unstable KPI and unstable status; remove unnecessary e2e install script

## Legal

### DIFF
--- a/packages/react-components/src/index.ts
+++ b/packages/react-components/src/index.ts
@@ -18,5 +18,7 @@ export {
 export { useViewport } from './hooks/useViewport';
 export { TimeSync } from './components/time-sync';
 
+export { Kpi as UnstableKpi } from './components/kpi/kpi';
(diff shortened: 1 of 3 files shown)


square-li (approved):

packages/react-components/src/index.ts
```diff
@@ -18,5 +18,7 @@ export {
 export { useViewport } from './hooks/useViewport';
 export { TimeSync } from './components/time-sync';
 
+export { Kpi as UnstableKpi } from './components/kpi/kpi';
```
jmbuss, line 21: I think we might need to export the styles as well? I think there is a css file for both of these.

diehbria: your correct, need to alter the build system and figure out how we want to build the styles hopefully in a way that doesn't require additional style imports

PR #712, opened 2023-03-15
ci: run playwright tests in CI, setup KPI and Status tests

diehbria:
The primary goal of this PR is to include e2e tests in the CI for react-components, which can later be used by all packages leveraging playwright and storybook.

* Improve default mergings in KPI and status widgets to allow undefined props passed in to not overwrite the defaults
* Exclude tests in build outputs for packages
* remove unused uuid dependence on testing util package
* correctly set iotsitewise client as dependency for core (instead of dev-dep)
* regenerate status and KPI snapshots with correct images
* re-factor react-components storybook structure to more organized and provide a better DX in utilization of storybook
* update e2e github action to utilize latest action releases
* utilize ubuntu-latest in CI actions

## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/react-components/playwright.config.ts
```diff
@@ -19,7 +19,7 @@ export default defineConfig({
      * For example in `await expect(locator).toHaveText();`
      */
     timeout: 5000,
-    toMatchSnapshot: { maxDiffPixels: 200 },
+    toMatchSnapshot: { maxDiffPixels: 50 },
```
tracy-french, line 22: Nice!

packages/react-components/src/components/kpi/kpi.tsx
```diff
@@ -5,7 +5,8 @@ import { useViewport } from '../../hooks/useViewport';
 import { widgetPropertiesFromInputs } from '../../common/widgetPropertiesFromInputs';
 import type { Annotations, StyleSettingsMap, Viewport, TimeSeriesDataQuery } from '@iot-app-kit/core';
 import type { KPISettings } from './types';
-import { DEFAULT_KPI_SETTINGS } from './constants';
+
+const DEFAULT_VIEWPORT: Viewport = { duration: '10m' };
```
tracy-french, line 9: Non-blocking question: Why is the KPI not using a shared constant for default viewport which other visualizations would also use?

packages/react-components/src/components/kpi/types.ts
```diff
@@ -1,7 +1,7 @@
 import type { WidgetSettings } from '../../common/dataTypes';
 
 export type KPIProperties = WidgetSettings & {
-  settings: KPISettings;
+  settings?: Partial<KPISettings>;
 };
```
tracy-french, line 5: Non-blocking question: Why do KPI settings need to be made optional? Why not set them as optional from the start?

packages/react-components/stories/dial/dialBase.stories.tsx
```diff
@@ -18,7 +21,9 @@ export default {
   },
 } as ComponentMeta<typeof DialBase>;
 
-export const Main: ComponentStory<typeof DialBase> = ({ yMin, yMax, showName, showUnit, ...args }) => (
+type StoryInputs = DialSettings & { alarmPoint?: DataPoint; propertyPoint?: DataPoint };
```
tracy-french, line 24: Non-blocking feedback: The strict TS eslint settings you plan to introduce will prefer you use `interface` when typing records.

tracy-french: This might be useful to you https://github.com/sindresorhus/type-fest/blob/main/source/spread.d.ts.

tracy-french (approved):
LGTM. No blocking feedback.

packages/react-components/src/components/kpi/kpi.tsx
diehbria: done!

packages/react-components/src/components/kpi/types.ts
diehbria: good catch, i was not quite doing what I intended on doing - slightly revised but kept the partial

tracy-french: Nice! Thank you!

tracy-french (approved):

PR #732, opened 2023-03-16
feat(dashboard): kpi and status widget empty states

tjuranek:
## Overview
Adds empty states for the KPI and status widgets on the dashboard.

Screenshot 2023-03-16 at 9 58 36 AM

## Legal

### DIFF
--- a/packages/dashboard/src/customization/widgets/kpi/component.css
+++ b/packages/dashboard/src/customization/widgets/kpi/component.css
@@ -1,0 +1,14 @@
+.kpi-widget-empty-state {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+}
+
+.kpi-widget-empty-state-message-container {
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  justify-content: center;
+  text-align: center;
+}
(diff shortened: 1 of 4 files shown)


packages/dashboard/src/customization/widgets/kpi/component.css
```diff
@@ -0,0 +1,16 @@
...
+  flex-direction: column;
+  height: 100%;
+  padding: 0 2em;
+  position: relative;
+  justify-content: center;
+  text-align: center;
+}
+
+.kpi-widget-empty-state-title {
+  position: absolute;
```
jmbuss, line 13: can we use some flex or grid to handle the positioning of the title? I think this will end up overlaying on the no properties or alarms text if you size the widget small enough. I think it's ok if the text overflows the widget bounds too, the other charts behave that way too.

packages/dashboard/src/customization/widgets/status/component.css
```diff
@@ -0,0 +1,16 @@
+.status-widget-empty-state {
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  padding: 0 2em;
```
jmbuss, line 6: Can we wrap the widget in a box to handle the padding? I want to defer all of the spacing styles to cloudscape so it's consistent.

packages/dashboard/src/customization/widgets/kpi/component.css
jmbuss: I was thinking maybe use SpaceBetween (which spans the whole widget) and wrap the title and the body text, and left align the title and then center align the body?

jmbuss (approved):

PR #737, opened 2023-03-16
chore(Components): refactor react components to utilize styled compon…

diehbria:
This PR serves 3 main purposes:
- update AWS SDKs to be consistent across the monorepo
- Utilize styled components in react-components, remvoe css from the build process
- Get a single working snapshot test in dashboard to ensure no breakages

Version management:
- set release-please to always increment patch version. this is done because `main` is currently set to release as alpha. This will be removed once were ready for production v4.
- update some versions that weren't correctly in sync due to not being included in the release manifest 

Dependency related:
- update dashboard dependencies
- Update SDK to utilize consistent, specific version. Not allowing a range due to the availability of other patch versions not existing within internal amzn build systems

Other Dashboard changes
* set a larger default size for the dashboard size-panes

Video Player changes
* remove unused videojs css export on video-player. This was not being bundled correctly or exposed to users of the video player in a documented way. needs to be revisted (@sheilaXu for visibility)

## Legal


packages/react-components/src/components/shared-components/StatusIcon/StatusIcon.tsx
```diff
@@ -1,15 +1,20 @@
...
-import './StatusIcon.css';
+import styled from 'styled-components';
 import type { StatusIconType } from '@iot-app-kit/core';
 import { STATUS_ICON_TYPE } from '@iot-app-kit/core';
 import { getIcons } from '../../../common/iconUtils';
 
+const Icon = styled.div`
+  position: relative;
+  margin-right: 3px;
+  top: 2px;
```
jmbuss, line 10: should we setup a variables object to keep some standardized spacing? or we can use cloudscape and just use their variables in JS?

diehbria: Yes we should revisit this styling - I was looking to tackle that in a follow up and this PR focus on just porting over to styled components to allow it to be imported with the existing build system that doesn't handle `css` files correctly

jmbuss: cool works for me!

jmbuss (approved):

PR #739, opened 2023-03-17
fix(dashboard): fix resizing issue

### LINKED ISSUE #573: Dashboard: resize bug
*Resize bug*

1. add 2 widgets
2. select both widgets
3. resize them down to 0 height and width
4. resize them up to some larger height and width
    1. the widgets are now both full size.
    2. (works with any amount of widgets in selection)

square-li:
## Overview
Switched to a new model to calcuate resizing. Now resizing is based on the changes of selection boxes to ensure group resizing doesn't change relative sizes&positions of widgets in the group.

chrome-capture-2023-2-17

Also renamed a few helper functions to reflex its purposes.

fix #573 

## Legal


packages/dashboard/src/util/resizeSelectionBox.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  if (anchor.includes('left')) {
+    // don't allow the selection box to be smaller than the minimum width
+    if (newRect.width - vector.x >= MIN_WIDTH) {
+      newRect.x += vector.x;
+      newRect.width -= vector.x;
+    }
+  }
+  if (anchor.includes('right')) {
+    newRect.width += vector.x;
```
jmbuss, line 35: should we do a min width check here too? The vector can be negative

square-li: width/height checking is at end of the function by `rectWithinMin`

square-li: We only need to check if x/y are moved here

packages/dashboard/src/util/resizeSelectionBox.ts
```diff
@@ -0,0 +1,42 @@
+import type { Position, Rect } from '~/types';
+import type { Anchor } from '~/store/actions';
+
+const MIN_WIDTH = 2;
```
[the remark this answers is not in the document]
jmbuss: ahh ok cool. didn't know if there was some edge case that we were handling that required a specific size

packages/dashboard/src/util/resizeSelectionBox.ts
jmbuss: so the specific checks on left and top are there to prevent the x value from pushing?

packages/dashboard/src/util/resizeSelectionBox.ts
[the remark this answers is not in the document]
square-li: I think it could give more spaces between anchors so they don't get stacked

packages/dashboard/src/util/resizeSelectionBox.ts
square-li: Yes. So if widget doesn't get resized on width or height, don't move the origin point (x,y).

square-li: we only need to check moving from top or left, since only these two directions modify x/y.

jmbuss: gotcha, perfect!

jmbuss (approved):

diehbria (approved):

PR #767, opened 2023-03-21
feat(composer): data binding variable support for data overlay

sheilaXu:
## Overview
Implement the data binding variable support for data overlay.
User can input `${binding-name}` in overlay markdown/string content to display the latest property history value of that configured data binding

variable-editing
variable-viewing

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
+++ b/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
@@ -1,22 +1,80 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useCallback, useEffect, useMemo, useState } from 'react';
 
+import { useSceneComposerId } from '../../../common/sceneComposerIdContext';
+import { ITwinMakerEntityDataBindingContext } from '../../../interfaces';
 import { Component } from '../../../models/SceneModels';
+import { useStore } from '../../../store';
+import { dataBindingValuesProvider } from '../../../utils/dataBindingUtils';
+import { replaceBindingVariables } from '../../../utils/dataBindingVariableUtils';
 import { ReactMarkdownWrapper } from '../../wrappers/ReactMarkdownWrapper';
 import './styles.scss';
 
 export interface DataOverlayDataRowProps {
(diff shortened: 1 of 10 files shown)


packages/scene-composer/src/utils/dataBindingUtils.ts
```diff
@@ -78,7 +78,7 @@ export const dataBindingValuesProvider = (
   dataInput?: IDataInput,
   dataBinding?: IValueDataBinding,
   dataBindingTemplate?: IDataBindingTemplate,
-): Record<string, any> => {
+): Record<string, unknown> => {
```
diehbria, line 81: thank you! 🙏 lets celebrate by decrementing the max warnings down one in the package.json :-)

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
```diff
@@ -1,22 +1,80 @@
...
+  rowData,
+  overlayType,
+  valueDataBindings,
+}: DataOverlayDataRowProps): ReactElement => {
+  const sceneComposerId = useSceneComposerId();
+  const dataInput = useStore(sceneComposerId)((state) => state.dataInput);
+  const dataBindingTemplate = useStore(sceneComposerId)((state) => state.dataBindingTemplate);
+  const isEditing = useStore(sceneComposerId)((state) => state.isEditing());
+
+  const [stringContent, setStringContent] = useState<string>('');
```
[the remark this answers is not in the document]
sheilaXu: this component is not just markdown row, but could be any row type, so we can't just initialize it assuming it's markdown

packages/scene-composer/src/utils/dataBindingUtils.ts
sheilaXu: that will need more effort to look through the whole data input flow typings to make sure this type is correct. the return type will be something like this at least
 ```
Record<string, number | boolean | string | number[] | boolean[] | string[]>
```
but the type of `values` filed in `dataInput` is currently `any[]`, and the value types in the return type should be derived from it.

PR #777, opened 2023-03-21
feat(TimeSync): add a lastUsedBy string to the useViewport hook

corteggiano:
## Overview
Add a lastUpdatedBy string which will be used to gracefully re-render all the charts if a user scrolls or moves the charts. This string will be passed to synchro-charts components so they know when to ignore viewport updates.

## Legal


packages/react-components/src/components/time-sync/time-sync.spec.tsx
```diff
@@ -113,3 +113,17 @@ it('provides autogenerated viewport group name', () => {
     </TimeSync>
   );
 });
+
+it('returns no lastUpdateBy value', () => {
+  const ViewportUser = () => {
+    const { lastUpdatedBy } = useViewport();
+    expect(lastUpdatedBy).toBeNull;
```
diehbria, line 120: `toBeNull()` (should execute the assertion call

diehbria (approved):

PR #809, opened 2023-03-23
feat(TwinMakerTools): add automatic workspace creation in deploy + misc bug fixes, add unit tests

gsechkin-aws:
## Overview
Added new workspace creation feature to tmdt, merged previously reviewed bug fixes in deploy.ts, init.ts, and component-type.ts, and added a few unit tests.

## Verifying Changes

npm test
_and_ 
npm run build -> manually testing the individual ./tmdt_local init, deploy, nuke commands

## Legal


packages/tools-iottwinmaker/README.md
```diff
@@ -4,7 +4,7 @@ This package contains functionality for the AWS IoT TwinMaker Development Tools
 (https://docs.aws.amazon.com/iot-twinmaker/latest/guide/what-is-twinmaker.html) project management. 
 
 
-TMDT can snapshot an existing TwinMaker workspace (init), deploy this snapshot into any other existing workspace (deploy), and easily delete all entities, component types, and scenes from a workspace (nuke).
+TMDT can snapshot an existing TwinMaker workspace (init), deploy this snapshot into any other existing workspace or automatically create a new workspace, S3bucket, and workspace role to deploy into (deploy), and easily delete all entities, component types, and scenes from a workspace (nuke).
```
matwld, line 7: Not an issue for this pr but we should probably break out the commands into their own sections with at least the required parameters explained

packages/tools-iottwinmaker/src/lib/utils.ts
```diff
@@ -31,4 +32,54 @@ async function verifyWorkspaceExists(workspaceId: string) {
...
+  return result;
+}
+
+/**
+ * Helper function for retrying workspace creation during role propogation
+ * @param func function to retry
+ * @param attempts number of attempts
+ * @returns generic promise for the passed in function return value
+ */
+async function retryFunction<T>(func: () => Promise<T>, attempts: number) {
```
matwld, line 68: The documentation on this needs to be generalized or rename the function. If this is just to retry creating a workspace then it doesn't need to take a function, it can just take workspace id and try with existing credentials. If we want a generalized retryFunction method then it should not report that it is waiting for role creation when it receives a validation exception. This method should either be specific or general, but not both

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
+import {
+  CreateWorkspaceCommandOutput,
+  GetWorkspaceCommandOutput,
+  ResourceNotFoundException,
+} from '@aws-sdk/client-iottwinmaker';
+
```
matwld, line 6: Nit: line breaks between import statements

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
...
+  const roleArn: string = createRoleResponse.Role.Arn;
+  console.log(`Created Role: ${roleArn}.`);
+  const createPolicyResponse: CreatePolicyCommandOutput = await aws().iam.createPolicy({
+    PolicyName: `IoTTwinMaker${roleType}Policy-${uniqueId}`,
+    PolicyDocument: rolePolicy,
+    Description: `${roleType} Policy generated by AWS IoT TwinMaker Getting Started Script`,
+  });
+
+  const policyArn: string | undefined = createPolicyResponse.Policy?.Arn;
```
matwld, line 52: Above you throw an error if the role is not created successfully, should follow same pattern here

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
...
+    accountId,
+    accountArn,
+    region,
+    workspaceS3BucketArn: s3BucketArn,
+  };
+
+  const workspaceRolePolicy: string = replaceTemplateVars(JSON.stringify(workspaceRolePolicyTemplate), policyParams);
+  const workspaceRoleArn: string = await createRoleAndPolicy(
+    uniqueId,
+    'WorkspaceRole',
```
matwld, line 110: Need a more specific name for the policy than this, maybe we can follow the standard twinmaker pattern of:
`twinmaker-workspace-<workspaceId>-<accountId>-<region>`

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
...
+    JSON.stringify(WORKSPACE_DASHBOARD_ROLE_ASSUME_POLICY),
+    policyParams
+  );
+  const workspaceDashboardRolePolicy: string = replaceTemplateVars(
+    JSON.stringify(workspaceDashboardRolePolicyTemplate),
+    policyParams
+  );
+  const workspaceDashboardRoleArn: string = await createRoleAndPolicy(
+    uniqueId,
+    'WorkspaceDashboardRole',
```
matwld, line 141: Same as above comment, need more specific name, at least include workspaceId

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
...
+  return {
+    workspaceArn,
+    workspaceS3BucketArn: s3BucketArn,
+    workspaceRoleArn,
+    workspaceDashboardRoleArn,
+  };
+}
+
+/**
+ * Wrapper function for workspace creation
```
matwld, line 155: Nit: I don't think this is a wrapper function, documentation could just be:
`Creates a workspace if it does not already exist`

packages/tools-iottwinmaker/src/lib/policy/workspace-dashboard-role-policy.ts
```diff
@@ -0,0 +1,28 @@
+export const workspaceDashboardRolePolicyTemplate = {
```
johnnyw-aws, line 1: To confirm: does this match the latest roles we generate in the console? I think the one I recently generated there also had SiteWise and Lambda permissions?

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
...
+ * Helper function during workspace creation to create S3 bucket
+ * @param uniqueId identifier for bucket used as suffix in the name
+ * @returns promise of the role S3 bucket name and arn
+ */
+async function createWorkspaceS3Bucket(uniqueId: string) {
+  const s3BucketName = `iot-twinmaker-workspace-bucket-${uniqueId}`;
+  console.log('Creating S3 Bucket for the TwinMaker Workspace...');
+  const createBucketResponse: CreateBucketCommandOutput = await aws().s3.createBucket({ Bucket: s3BucketName });
+  console.log(`Workspace S3 Bucket created: ${createBucketResponse.Location}`);
+  console.log('Configuring CORS policy...');
```
johnnyw-aws, line 72: UX question: does this need a corresponding message to indicate it completed (similar to the "creating s3 bucket ...")

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
...
+      role: workspaceRoleArn,
+      s3Location: s3BucketArn,
+    });
+    const workspaceArn: string | undefined = createWorkspaceResponse.arn;
+    if (!workspaceArn) {
+      throw new Error('Failed to create Workspace');
+    }
+    console.log(`Created Workspace: ${workspaceArn}`);
+    return workspaceArn;
+  }, 10);
```
johnnyw-aws, line 127: later versions of the AWS SDK may support this feature of "waiters" (intended to make it easier to wait for a resource to transition to some state [[1]](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/IoTTwinMaker/Waiters.html) [[2]](https://aws.amazon.com/blogs/developer/waiters-in-modular-aws-sdk-for-javascript/) not sure if they already exist in the client we're using but may be something to check out if it saves us from having to build our own retry logic

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -0,0 +1,182 @@
...
+  });
+  return { s3BucketName, s3BucketArn: `arn:aws:s3:::${s3BucketName}` };
+}
+
+/**
+ * Driver function for workspace creation
+ * @param workspaceId workspace-id to be created
+ * @returns promise of workspace arn, s3 arn, role arn, dashboardrole arn
+ */
+async function prepareWorkspace(workspaceId: string) {
```
johnnyw-aws, line 94: does the creation logic match what we do in the console? curious if we can reuse some logic there

e.g. I believe in console the created bucket has access logging enabled as well as object versioning?

packages/tools-iottwinmaker/src/lib/workspace.ts
johnnyw-aws: +1, similar to other comment, I think ideally the TMDT create workspace should result in the same as if it was done using the console

packages/tools-iottwinmaker/src/commands/test-constants.ts
```diff
@@ -74,6 +75,48 @@ export const componentType1Definition = {
   description: getComponentType1Resp.description,
 };
 
+export const componentTypeWithInheritedProperty = {
```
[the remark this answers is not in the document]
diehbria: yes we'll want to keep all things typed :+1:

packages/tools-iottwinmaker/src/commands/deploy.ts
```diff
@@ -84,11 +110,24 @@ export const handler = async (argv: Arguments<Options>) => {
...
             acc[key] = value;
+          } else if ('defaultValue' in value) {
+            acc[key] = { defaultValue: value['defaultValue'] };
           }
           return acc;
         }, {} as { [key: string]: object });
         componentTypeDefinition['propertyDefinitions'] = filtered_property_definitions;
       }
+      // remove inherited functions
+      const componentTypeFunctions = componentTypeDefinition['functions'] as object;
```
diehbria, line 121: `as object` is almost the same as saying `as any`. there must be something we know more about it than it being any object?

packages/tools-iottwinmaker/src/lib/workspace.ts
gsechkin-aws: I'm not sure I understand this one, does the current "Workspace S3 Bucket created" not indicate it completed?

packages/tools-iottwinmaker/src/lib/workspace.ts
gsechkin-aws: Interesting, this may be something to look into more but I simplified the retry function as per Matt's comments above, and I think it is fairly straightforward for our use case for now. Not entirely sure if we are using V2 or V3 like linked above

packages/tools-iottwinmaker/src/commands/test-constants.ts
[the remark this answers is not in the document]
gsechkin-aws: I fixed most of the missing types in this file, but unfortunately there are some that don't belong to a specific "GetComponentTypeCommandOutput" or something simple like that because they have modified fields to fit tmdt's needs (differs from API types).

I could explicitly declare the full type definition e.g:

type componentTypeWithInheritedProperty: {
    componentTypeId: string | undefined;
    description: string | undefined;
    propertyDefinitions: {
        testProperty: {
            isInherited: boolean;
        };
    };
}

However that same information is visible when you hover over the name in TS, and this type would only be used 1 time, is this still necessary in these specific cases?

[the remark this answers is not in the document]
diehbria: Yes makes it even more important. precisely what are the differences? There's utility functions in typescript that can be used to slightly modify an existing type, so you will not need to fully re-declare the type.

see https://www.typescriptlang.org/docs/handbook/utility-types.html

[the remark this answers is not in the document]
gsechkin-aws: Got it, will update!

packages/tools-iottwinmaker/src/commands/deploy.ts
gsechkin-aws: fixed same as above, thanks!

packages/tools-iottwinmaker/src/lib/workspace.ts
gsechkin-aws: makes sense, updated

packages/tools-iottwinmaker/src/lib/workspace.ts
gsechkin-aws: I notice the created bucket has those 2 things disabled, is there a way for my to enable that using the SDK?

I checked this documentation: (https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-overview.html) and don't seem to see information about this.

In the TS file, our input options include this in the createBucket request:
```
/**
     * <p>The canned ACL to apply to the bucket.</p>
     */
    ACL?: BucketCannedACL | string;
    /**
     * <p>The name of the bucket to create.</p>
     */
    Bucket: string | undefined;
    /**
     * <p>The configuration information for the bucket.</p>
     */
    CreateBucketConfiguration?: CreateBucketConfiguration;
    /**
     * <p>Allows grantee the read, write, read ACP, and write ACP permissions on the
     *          bucket.</p>
     */
    GrantFullControl?: string;
    /**
     * <p>Allows grantee to list the objects in the bucket.</p>
     */
    GrantRead?: string;
    /**
     * <p>Allows grantee to read the bucket ACL.</p>
     */
    GrantReadACP?: string;
    /**
     * <p>Allows grantee to create new objects in the bucket.</p>
     *          <p>For the bucket and object owners of existing objects, also allows deletions and overwrites of those objects.</p>
     */
    GrantWrite?: string;
    /**
     * <p>Allows grantee to write the ACL for the applicable bucket.</p>
     */
    GrantWriteACP?: string;
    /**
     * <p>Specifies whether you want S3 Object Lock to be enabled for the new bucket.</p>
     */
    ObjectLockEnabledForBucket?: boolean;
    /**
     * <p>The container element for object ownership for a bucket's ownership controls.</p>
     *          <p>BucketOwnerPreferred - Objects uploaded to the bucket change ownership to the bucket
     *          owner if the objects are uploaded with the <code>bucket-owner-full-control</code> canned
     *          ACL.</p>
     *          <p>ObjectWriter - The uploading account will own the object if the object is uploaded with
     *          the <code>bucket-owner-full-control</code> canned ACL.</p>
     *          <p>BucketOwnerEnforced - Access control lists (ACLs) are disabled and no longer affect permissions.
     *          The bucket owner automatically owns and has full control over every object in the bucket. The bucket only
     *          accepts PUT requests that don't specify an ACL or bucket owner full control
     *          ACLs, such as the <code>bucket-owner-full-control</code> canned
     *          ACL or an equivalent form of this ACL expressed in the XML format.</p>
     */
    ObjectOwnership?: ObjectOwnership | string;
```

johnnyw-aws: it might be a separate API: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketVersioning.html

but would recommend following up with other folks on the team to see how the console does it

packages/tools-iottwinmaker/src/lib/workspace.ts
johnnyw-aws: sorry, I meant I see a message like `Configuring CORS policy...` but didn't see one indicating to the customer this activity completed

packages/tools-iottwinmaker/src/lib/policy/workspace-dashboard-role-policy.ts
gsechkin-aws: Yes, it depends on which permissions you select in the console, I'll update it to the "include all" instead of "no video permissions" as that seems more robust and can be changed later if needed?

johnnyw-aws: (can discuss offline) Sorry, can you clarify what you mean by "include all" ? Security will want us to ensure the permissions are least-privilege

packages/tools-iottwinmaker/src/lib/utils.ts
```diff
@@ -31,4 +31,48 @@ async function verifyWorkspaceExists(workspaceId: string) {
...
+    ['eu-central-1', 'FRA'],
+    ['us-iso-east-1', 'DCA'],
+    ['ap-northeast-2', 'ICN'],
+    ['ap-south-1', 'BOM'],
+    ['us-east-2', 'CMH'],
+    ['cn-northwest-1', 'ZHY'],
+    ['eu-west-2', 'LHR'],
+    ['ca-central-1', 'YUL'],
+    ['us-isob-east-1', 'LCK'],
+    ['eu-west-3', 'CDG'],
```
matwld, line 73: Maybe we only need the regions twinmaker is available in? Otherwise could confuse customers

matwld (approved):

gsechkin-aws:
> What's the state of this PR?

Waiting Johnny's approval, will ping again today.

matwld (approved):

matwld (approved):

PR #810, opened 2023-03-23
fix(composer): black screen on Matterport scene

mukeshsahay:
## Overview
- Fixed the black screen issue with Matterport Scene
When navigating in the scene with Matterport space the floor turns black before moving. This is due to the ground plane mesh inserted by WebGLCanvasManager which writes to the depth buffer but is otherwise invisible. Due to quirks of render order sorting, sometimes it gets rendered before the Matterport model mesh, which causes portions below the ground plane to be discarded by depth testing.

Setting its renderOrder to a value greater than zero i.e. 1 so it’s never rendered before the Matterport mesh when scene is using a Matterport space otherwise leave the default render order.

- Also, updated to not show the 3rd Party Resources section in Viewer mode

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/WebGLCanvasManager.tsx
+++ b/packages/scene-composer/src/components/WebGLCanvasManager.tsx
@@ -40,6 +40,7 @@ export const WebGLCanvasManager: React.FC = () => {
   const { gl } = useThree();
   const domRef = useRef<HTMLElement>(gl.domElement.parentElement);
   const environmentPreset = getSceneProperty(KnownSceneProperty.EnvironmentPreset);
+  const matterportModelId = getSceneProperty(KnownSceneProperty.MatterportModelId);
   const rootNodeRefs = document.rootNodeRefs;
   const [startingPointerPosition, setStartingPointerPosition] = useState<THREE.Vector2>(new THREE.Vector2());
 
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/components/WebGLCanvasManager.tsx
```diff
@@ -54,6 +56,14 @@ export const WebGLCanvasManager: React.FC = () => {
     }
   }, [environmentPreset]);
 
+  useEffect(() => {
```
diehbria, line 59: it is best to avoid using `useEffect` where possible, please read: https://react.dev/learn/you-might-not-need-an-effect

For example in this case, it would actually be better in your render function to just write:

`const groundMeshOrder = matterportModelId ? 1 : undefined`

diehbria (changes requested):

packages/scene-composer/src/components/WebGLCanvasManager.tsx
sheilaXu: seems more like a `useMemo` case, and also simple enough to just add to the rendering part

diehbria: I don't think `useMemo` is appropriate here - caching won't save us anything on a simple computation like this unless i'm missing something

diehbria (approved):

PR #814, opened 2023-03-24
feat(Matterport): update scene based on Matterport settings

mukeshsahay:
## Overview
- Handle model id and connection name change and update view to enable/disable Matterport viewer
- Corrected the dependency for Sync Tags button
- Fixed some warnings in code (touched files)

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -64,15 +66,27 @@ export const MatterportIntegration: React.FC = () => {
     return () => unsubscribe(onUpdated);
   }, []);
 
-  const onConnectionNameChange = useCallback(async (event) => {
+  const enableMatterportViewer = useMemo(() => {
```
diehbria, line 69: `useMemo` is not necessary for simple operations and adds complexity. It's completely fine to just rune this computation on every render and won't have any impact on performance. should avoid useMemo unless its needed

diehbria (changes requested):

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -88,20 +102,28 @@ export const MatterportIntegration: React.FC = () => {
...
+      );
       setDirty(false);
+      setMatterportViewer(
+        matterportModelId.current !== undefined &&
+          matterportModelId.current !== '' &&
+          selectedConnectionName !== OPTIONS_PLACEHOLDER_VALUE,
+      );
     }
-  }, [matterportModelId]);
+  }, [matterportModelId.current]);
```
diehbria, line 126: why does matter port model id need to be a ref? Refs are typically used for things that you _don't_ want to trigger re-renders, but it looks like we do want it to trigger a re-render? Using lots of hooks and refs, and dirty flags complicates things and should be avoided if possible

https://react.dev/learn/referencing-values-with-refs

diehbria:
overall it feels like the logic around matterport is very complex - lots of side effects, dirty flags, etc.

the goal of react is to try to keep things declarative, one way data flow, but its difficult to follow this logic.

There must be a way to simplify? This type of logic won't scale and is hard to follow

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -22,7 +22,8 @@ export const MatterportIntegration: React.FC = () => {
   const setSceneProperty = useStore(sceneComposerId)((state) => state.setSceneProperty);
 
   const [dirty, setDirty] = useState(false);
-  const [matterportModelId, setMatterportModelId] = useState(getSceneProperty(KnownSceneProperty.MatterportModelId));
+  const matterportModelId = useRef(getSceneProperty(KnownSceneProperty.MatterportModelId));
```
diehbria, line 25: whys it a ref? we set it as a ref, and then force re-renders anyway with dirty flags?

diehbria:
another high level comment. Avoid state. If something can just be written as a function on existing props and state to get the value, _do that instead_.

State is complicated. we are creating lots of edge cases and complicated data flow by taking things that don't need to be state, and turning it into state.

avoid state, avoid useEffect, avoid using refs - try to write it in the simplest, most straight forward way and only reach outside of typical react data flow when absolutely necessary

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -88,20 +102,28 @@ export const MatterportIntegration: React.FC = () => {
   const onMatterportModelIdChange = useCallback(
     (event) => {
       const value = event.detail.value;
-      if (value !== matterportModelId) {
-        setMatterportModelId(value);
+      if (value !== matterportModelId.current) {
+        matterportModelId.current = value;
         setDirty(true);
```
diehbria, line 107: i dont understand why we have all these dirty flags - react knows when to re-render based on updates to state and manages it efficiently. It feels like were trying to step outside of react and write imperative code

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -88,20 +102,28 @@ export const MatterportIntegration: React.FC = () => {
...
       }
     },
-    [setMatterportModelId, setDirty],
+    [setDirty],
   );
 
   useEffect(() => {
```
diehbria, line 113: i'm not convinced we need an effect to solve this. why do we need to use effect to update these?

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
mukeshsahay: Removed the dirty flag. As per the new changes, loss of focus from the input control will be used to update the value in scene file.

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
mukeshsahay: useState was not working for me at the time when I was working on it. I wanted to verify the checks on changes, so created a PR with these changes. Have reverted it to useState and now it is working as expected.

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
mukeshsahay: Not using useRef anymore

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
mukeshsahay: Removed useMemo

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -64,15 +66,17 @@ export const MatterportIntegration: React.FC = () => {
...
-    }
-  }, []);
+  const onConnectionNameChange = useCallback(
+    async (event) => {
+      const newConnection = event.detail.selectedOption.value;
+      if (twinMakerSceneMetadataModule) {
+        const updatedSceneInfo = await getUpdatedSceneInfoForConnection(newConnection, twinMakerSceneMetadataModule);
+        await twinMakerSceneMetadataModule.updateSceneInfo(updatedSceneInfo);
+      }
+      setSelectedConnectionName(newConnection);
```
sheilaXu, line 76: will the UI be delayed to refresh with the new selection if you set state after the update API?

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -85,23 +89,28 @@ export const MatterportIntegration: React.FC = () => {
...
+    }
+    setFocusInput(false);
+  }, [matterportModelId, setFocusInput]);
+
+  const onMatterportModelIdChange = useCallback((event) => {
+    const value = event.detail.value;
+    if (value !== matterportModelId) {
+      setMatterportModelId(value);
+    }
+  }, []);
```
sheilaXu, line 104: you will need to add matterportModelId as dependency, otherwise it will always compares with the original value

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
```diff
@@ -85,23 +89,28 @@ export const MatterportIntegration: React.FC = () => {
...
+
+  const enableMatterportViewer =
+    matterportModelId !== undefined && matterportModelId !== '' && selectedConnectionName !== OPTIONS_PLACEHOLDER_VALUE;
```
sheilaXu, line 107: nit: `matterportModelId !== undefined && matterportModelId !== ''` is essentially equivalent to `!isEmpty(matterportModelId)`

packages/scene-composer/src/store/slices/ViewOptionStateSlice.ts
```diff
@@ -31,4 +30,10 @@ export const createViewOptionStateSlice = (
       draft.lastOperation = 'setTagSettings';
     });
   },
+  setMatterportViewer: (enabled: boolean) => {
```
sheilaXu, line 33: something like `setMatterportViewerEnabled`? setMatterportViewer sounds like it's setting a viewer..

sheilaXu (changes requested):

packages/scene-composer/src/store/slices/ViewOptionStateSlice.ts
mukeshsahay: Done

mukeshsahay: Done

packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -241,18 +242,24 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
...
           }
         })
         .catch((error) => {
           setLoadSceneError(error || new Error('Failed to get scene details'));
         });
+    } else {
+      setUpdatedExternalLibraryConfig({ ...externalLibraryConfig });
     }
-  }, [externalLibraryConfig, matterportModelId, sceneMetadataModule]);
+  }, [enableMatterportViewer, externalLibraryConfig, matterportModelId, sceneMetadataModule]);
```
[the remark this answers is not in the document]
mukeshsahay: Updated the if condition

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
mukeshsahay: Done

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
mukeshsahay: added matterportModelId as dependency

packages/scene-composer/src/components/panels/scene-settings/MatterportIntegration.tsx
mukeshsahay: No delay observed. Expected to save first before updating the state variable as it will trigger the loading of viewer

sheilaXu (approved):

PR #815, opened 2023-03-24
feat(react-comp): sync viewports using @iot-app-kit/charts pckg

corteggiano:
## Overview
pass through the `setViewport` function (from the `useViewports` hook) into the @iot-app-kit/charts package, which will set the viewport on Gesture and will prevent re render if `lastUpdatedBy` is set to be `chart-gesture`

viewport

## Legal

### DIFF
--- a/packages/react-components/src/components/bar-chart/barChart.tsx
+++ b/packages/react-components/src/components/bar-chart/barChart.tsx
@@ -1,8 +1,7 @@
 import React from 'react';
 import { StyleSettingsMap, Threshold, TimeSeriesDataQuery, Viewport } from '@iot-app-kit/core';
-import { BarChart as BarChartBase } from '@iot-app-kit-visualizations/react';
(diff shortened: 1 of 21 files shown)


packages/react-components/src/components/bar-chart/barChart.tsx
```diff
@@ -34,16 +35,23 @@ export const BarChart = ({
     },
     styles,
   });
-  const { viewport } = useViewport();
+  const { viewport, setViewport, group, lastUpdatedBy } = useViewport();
   const allThresholds = [...queryThresholds, ...thresholds];
 
   const utilizedViewport = passedInViewport || viewport || DEFAULT_VIEWPORT; // explicitly passed in viewport overrides viewport group
 
+  useEffect(() => {
```
diehbria, line 43: blocking: don't useEffect here to add global listeners - react has a great built in way to handle this use case https://react.dev/learn/responding-to-events

diehbria (changes requested):

PR #824, opened 2023-03-27
feat(TwinMakerTools): add tools-iottwinmaker to npm publish workflow

matwld:
## Overview
Add tools-iottwinmaker (`tmdt`) to app-kit npm publishing workflow.
Also:
- update package version to 4.0.0
- replace references to tmdk with tmdt
- update README
- update package-lock and .gitignore

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayComponent.tsx
+++ b/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayComponent.tsx
@@ -18,7 +18,7 @@ export const DataOverlayComponent = ({ node, component }: DataOverlayComponentPr
   return (
     <group>
       <Html
-        className='html-wrapper'
+        className='tm-html-wrapper'
         // TODO: add position after finding proper way to always display overlay right above tag
         // position={position}
         style={{
--- a/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/__tests__/DataOverlayComponentSnap.spec.tsx
+++ b/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/__tests__/DataOverlayComponentSnap.spec.tsx
@@ -51,7 +51,7 @@ describe('DataOverlayComponent', () => {
     const { container } = render(
       <DataOverlayComponent node={mockNode as ISceneNodeInternal} component={mockComponent} />,
     );
-    expect(container.getElementsByClassName('html-wrapper').length).toBe(1);
+    expect(container.getElementsByClassName('tm-html-wrapper').length).toBe(1);
     expect(container).toMatchSnapshot();
   });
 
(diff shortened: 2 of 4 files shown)


johnnyw-aws:
Just wanted to confirm my understanding of the customer experience:

- after this change, there will be an page on https://www.npmjs.com/ for TMDT?
- what will the content on that page be?
- and I should be able to `npm install -g` the package and then I have the `tmdt` executable in my environment?
- is there a version number associated with the package? looks like it's [3.0](https://github.com/awslabs/iot-app-kit/blob/main/packages/tools-iottwinmaker/package.json#L4) now? Should it be 1.0? also want to see if we can align the user-agent value before releasing to npm [[1]](https://github.com/awslabs/iot-app-kit/blob/main/packages/tools-iottwinmaker/src/lib/aws-clients.ts#L20) ?
- to confirm: after this there's no way to "undo" the release right?

matwld:
> Just wanted to confirm my understanding of the customer experience:
> 
>     * after this change, there will be an page on https://www.npmjs.com/ for TMDT?
> 
>     * what will the content on that page be?
> 
>     * and I should be able to `npm install -g` the package and then I have the `tmdt` executable in my environment?
> 
>     * is there a version number associated with the package? looks like it's [3.0](https://github.com/awslabs/iot-app-kit/blob/main/packages/tools-iottwinmaker/package.json#L4) now? Should it be 1.0? also want to see if we can align the user-agent value before releasing to npm [[1]](https://github.com/awslabs/iot-app-kit/blob/main/packages/tools-iottwinmaker/src/lib/aws-clients.ts#L20) ?
> 
>     * to confirm: after this there's no way to "undo" the release right?

* I believe there will be a page for tmdt under @iot-app-kit/tools-iottwinmaker. I assume the readme at tmdt's root will be used for the npm package.
* Yes, npm install -g from the tools-iottwinmaker root will currently install the tmdt executable globally
* The version number in the package.json is left over from our set up, should I change it to 1?
* For the user-agent value, I believe we can probably remove the versioning from tmdt definition but I will discuss offline with Johnny.
* I do not believe there is a way to "undo" this release.

johnnyw-aws:
Thanks Matt

> * I believe there will be a page for tmdt under @iot-app-kit/tools-iottwinmaker. I assume the readme at tmdt's root will be used for the npm package.

Can we change it to a generic or empty page for now? Looking over it again, the README for the project needs to be cleaned up, am wondering if we can decouple that from first npm release (or if needs to be cleaned up first?)

e.g. it currently refers to the private repo and the sample commands should remove references to our development to avoid confusing customers ("CookieFactory0601" and "SyncB") Also am wondering if the `tmdt_local` instructions should be part of a developer-only page instead of the landing README.

> * The version number in the package.json is left over from our set up, should I change it to 1?

Since it's alpha, is it appropriate to use a v0 version number? e.g. 0.1.0?

> * For the user-agent value, I believe we can probably remove the versioning from tmdt definition but I will discuss offline with Johnny.

Sorry, which versioning definition are you referring to? 

- we're using the user-agent one for usage tracking
- we should keep the version in the tmdt.json file (have we renamed it from tmdk.json?)  for compatibility-tracking
- I thought the one in the package.json gets reflected in npm?

PR #843, opened 2023-03-29
feat(composer): visibility toggles for overlays in settings panel

sheilaXu:
## Overview
enable control the visibility of all overlay panels and annotations from settings panel

overlay-toggle

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/package.json
```diff
@@ -46,7 +46,7 @@
     "convert-svg": "npx @svgr/cli --out-dir src/assets/auto-gen/icons/ --typescript --index-template tools/index-template.js -- src/assets/icons/",
     "release": "run-s compile copy-assets",
     "copy-assets": "copyfiles -e \"**/*.tsx\" -e \"**/*.ts\" -e \"**/*.snap\" -e \"**/*.js\" -e \"**/*.jsx\" -e \"**/*.json\" \"src/**/*\" dist/",
-    "lint": "eslint . --max-warnings=825",
+    "lint": "eslint . --max-warnings=730",
```
diehbria, line 49: yay!

packages/scene-composer/src/utils/componentSettingsUtils.spec.ts
```diff
@@ -11,6 +11,12 @@ describe('componentSettingsUtils', () => {
       expect(tagSettings).toEqual(DEFAULT_TAG_GLOBAL_SETTINGS);
     });
 
+    it('should return default overlay settings', () => {
+      const state = { getSceneProperty: jest.fn() } as any;
```
diehbria, line 15: no any here - that way if the expected input changes thru a refactor, this can be discovered via typescript error

sheilaXu: oops, that's why the warning count is larger than my early update lol.. I'll fix it

packages/scene-composer/src/components/panels/scene-settings/ComponentVisibilityToggle.tsx
```diff
@@ -0,0 +1,64 @@
...
+}) => {
+  const sceneComposerId = useContext(sceneComposerIdContext);
+  const toggleComponentVisibility = useViewOptionState(sceneComposerId).toggleComponentVisibility;
+  const componentVisible = useViewOptionState(sceneComposerId).componentVisibilities[componentType];
+  const getComponentRefByType = useStore(sceneComposerId)((state) => state.getComponentRefByType);
+  const getSceneNodeByRef = useStore(sceneComposerId)((state) => state.getSceneNodeByRef);
+  const componentNodeMap = useStore(sceneComposerId)((state) => state.document.componentNodeMap);
+  const { formatMessage } = useIntl();
+
+  const hasComponent = useMemo(() => {
```
divya-sea, line 30: What this method signifies?

packages/scene-composer/src/components/panels/scene-settings/OverlayPanelVisibilityToggle.tsx
```diff
@@ -0,0 +1,62 @@
...
+  );
+
+  const setSceneProperty = useStore(sceneComposerId)((state) => state.setSceneProperty);
+  const getSceneProperty = useStore(sceneComposerId)((state) => state.getSceneProperty);
+  const isViewing = useStore(sceneComposerId)((state) => state.isViewing());
+
+  // When the document settings is changed from other places (e.g. undo/redo),
+  // sync the view options visibility value with document settings.
+  useEffect(() => {
+    if (!isViewing && !!overlayPanelVisible !== !!documentSettings.overlayPanelVisible) {
```
divya-sea, line 30: Too many checks in one condition. May be you can simply this logic.

packages/scene-composer/src/components/panels/scene-settings/ComponentVisibilityToggle.tsx
mumanity: It's returning the component based on whether it's a  `Component.DataOverlaySubType` or `KnownComponentType`.

packages/scene-composer/src/components/panels/scene-settings/ComponentVisibilityToggle.tsx
```diff
@@ -0,0 +1,64 @@
...
+            ) as IDataOverlayComponentInternal
+          ).subType === componentType,
+      );
+    } else {
+      return Object.keys(getComponentRefByType(componentType as KnownComponentType)).length > 0;
+    }
+  }, [componentNodeMap, getComponentRefByType]);
+
+  return (
+    <React.Fragment>
```
mumanity, line 48: nit-pick: Prefer to only use the `<React.Fragment>` syntax when you have to add an attribute to the tag, otherwise using `<>` is sufficient & improves readability.

packages/scene-composer/src/components/panels/scene-settings/OverlayPanelVisibilityToggle.tsx
mumanity: There's only two checks here. What are you referencing for this being too many?

packages/scene-composer/src/components/panels/SettingsPanel.tsx
```diff
@@ -113,12 +136,13 @@ export const SettingsPanel: React.FC<SettingsPanelProps> = ({ valueDataBindingPr
...
       )}
 
-      {tagResizeEnabled && (
+      {(tagResizeEnabled || overlayEnabled) && (
         <ExpandableInfoSection
           title={intl.formatMessage({ description: 'ExpandableInfoSection Title', defaultMessage: 'Tag Settings' })}
           defaultExpanded={false}
         >
-          <SceneTagSettingsEditor />
+          {tagResizeEnabled && <SceneTagSettingsEditor />}
```
divya-sea, line 144: Component name is bit confusing to me as we have a separate section on the setting which says **Scene Settings** . We have another section called Tag Settings. Can we call this component name as Tag Settings?

mumanity (approved):

packages/scene-composer/src/components/panels/scene-settings/OverlayPanelVisibilityToggle.tsx
divya-sea: 1. !isViewing
2. !!overlayPanelVisible
3. !!documentSettings.overlayPanelVisible
4. !!overlayPanelVisible !== !!documentSettings.overlayPanelVisible
5. !isViewing && !!overlayPanelVisible !== !!documentSettings.overlayPanelVisible

packages/scene-composer/src/components/panels/SettingsPanel.tsx
sheilaXu: I'd rather think the Scene Settings is a bit too broad because essentially every settings in the settings panel is a scene setting that impacts the whole scene. and SceneTagSettings describes that this setting is for all tags in the scene, not just one tag, which I think is ok

packages/scene-composer/src/components/panels/scene-settings/ComponentVisibilityToggle.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Component } from '../../../models/SceneModels';
+import { findComponentByType } from '../../../utils/nodeUtils';
+
+export interface ComponentVisibilityToggleProps {
+  componentType: KnownComponentType | Component.DataOverlaySubType;
+  label: string;
+  onChange?: (newValue: boolean) => void;
+}
+
+export const ComponentVisibilityToggle: React.FC<ComponentVisibilityToggleProps> = ({
```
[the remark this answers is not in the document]
sheilaXu: mostly the toggle will only be on the scene's settings page since it can only handle visibility of all instances of the components of the specified type, but not for one instance, so I'll just place it here. we can move it later if other panels want it

packages/scene-composer/src/components/panels/scene-settings/ComponentVisibilityToggle.tsx
sheilaXu: this is to check whether the component exists. when there is no instance of the specified component type, the toggle will be disabled.

packages/scene-composer/src/components/panels/scene-settings/ComponentVisibilityToggle.tsx
sheilaXu: sure. I'll update in my next PR

packages/scene-composer/src/components/panels/scene-settings/OverlayPanelVisibilityToggle.tsx
sheilaXu: 2 & 3 are just to convert nullable booleans to boolean type because for this code, `undefined` and `false` are the same. so there are just 2 conditions here

divya-sea: I am not talking about conditions but talking about converting these values or checking the state of the each values part of these two conditions. It would be good if the logics are easy to read.

PR #847, opened 2023-03-29
feat(dashboard): assign default color based on cloudscape design

square-li:
## Overview
- default color are now based on [cloudscape data visualization colors](https://cloudscape.design/foundation/visual-foundation/data-vis-colors/)
chrome-capture-2023-2-29 (1)
- default refId are now based on both `assetId` and `propertyId` instead of just `propertyId` to avoid refId collision. Changing one property's color will not change other color of other properties with same `propertyId`.
chrome-capture-2023-2-29 (2)

- default colors are assigned based on index of property in all properties instead of index in asset query. So properties from different assets can be colored correctly.

- move assign default color logic and `colorPalette` to `core-util` package.

## Legal


packages/core-util/src/sdks/colorPalette.ts
```diff
@@ -0,0 +1,19 @@
+// color palette copied from https://cloudscape.design/foundation/visual-foundation/data-vis-colors/
+export const colorPalette = [
+  '#688ae8',
```
jmbuss, line 3: is it possible to just take their JS variables?

square-li: No we can't. I tried to import their variables. But they export variables like 
``` 
export var colorChartsPaletteCategorical1 = "var(--color-charts-palette-categorical-1-bonb16, #688ae8)";
```
which won't work with our charts.  So we have to use hard coded strings.

packages/dashboard/src/customization/widgets/utils/assignDefaultStyleSettings.ts
```diff
@@ -1,14 +1,20 @@
...
 
 const assignDefaultRefId = (siteWiseAssetQuery: SiteWiseAssetQuery) => ({
   assets: siteWiseAssetQuery.assets.map(({ properties, ...others }) => ({
     ...others,
     properties: properties.map((propertyQuery) => ({
       ...propertyQuery,
-      refId: propertyQuery.refId || propertyQuery.propertyId,
+      refId:
+        propertyQuery.refId ||
+        toId({
```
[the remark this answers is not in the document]
square-li: good point! updating now

packages/core-util/src/sdks/colorPalette.ts
jmbuss: ahhh dang, makes sense!

diehbria (approved):

PR #857, opened 2023-03-30
fix(ReactComponents): configure table, kpi and status to always fetch raw

diehbria:
## Overview
Table, KPI and status do not currently support displaying non raw data, but were trying to fetch non-raw data.

Updated configuration to default to requesting raw data

fixes https://github.com/awslabs/iot-app-kit/issues/854

## Legal

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "configuration/*"
   ],
   "scripts": {
-    "bootstrap": "npm install --workspaces --include-workspace-root && npm run build",
+    "install-ws": "npm install --workspaces --include-workspace-root",
+    "bootstrap": "npm run install-ws && npm run build",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -16,7 +16,8 @@
     "configuration/*"
   ],
   "scripts": {
-    "bootstrap": "npm install --workspaces --include-workspace-root && npm run build",
+    "install-ws": "npm install --workspaces --include-workspace-root",
+    "bootstrap": "npm run install-ws && npm run build",
```
square-li, line 20: Do we need a space between install and -ws?

square-li: 👍

square-li (approved):

PR #908, opened 2023-04-06
fix(composer): Hide camera view for matterport

mukeshsahay:
## Overview
Hide camera view functionality for a Matterport enabled scene

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


mukeshsahay:
On-hold for now.

packages/scene-composer/src/components/panels/TopBar.tsx
```diff
@@ -39,7 +41,7 @@ export const TopBar: FC = () => {
   }, [nodeMap]);
 
   const hasCameraView = cameraItems.length > 0;
-  const showTopBar = hasCameraView;
+  const showTopBar = hasCameraView && !matterportSdk;
```
sheilaXu, line 44: logically, you probably want to do 
```
  const showCameraView = cameraItems.length > 0 && !matterportSdk;
  const showTopBar = showCameraView;
```
since you are hiding the camera view specifically, not the whole top bar

packages/scene-composer/src/components/toolbars/floatingToolbar/items/AddObjectMenu.tsx
```diff
@@ -113,49 +114,50 @@ export const AddObjectMenu = () => {
...
+      {
+        uuid: ObjectTypes.MotionIndicator,
+      },
+    ];
+    if (!matterportSdk) {
+      menuItems.push({
+        uuid: ObjectTypes.ViewCamera,
+        feature: { name: COMPOSER_FEATURES.CameraView },
+      });
+    }
```
sheilaXu, line 158: you will change the position of this item by pushing as the last element. can we just disable it or need to delete it totally?

mukeshsahay: I wanted to remove this option completely, that's why this code. Is there any issue with change in position?

packages/scene-composer/src/components/panels/TopBar.tsx
mukeshsahay: Both are same as per the current implementation. I will add back the camera view feature once Matterport SDK has fix for this issue.

PR #818, opened 2023-03-24
fix(dashboard): height and width validation

tjuranek:
## Overview
Adds validation to the height and width fields on the side bar.

## Legal
This project is available under the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0.html).

PR #910, opened 2023-04-06
fix(dashboard): base settings validation

tjuranek:
## Overview
Republish of https://github.com/awslabs/iot-app-kit/pull/818

Validates the height and width fields with react-hook-form. Previous comments suggested using the `FormField` component in cloudscape. After talking with the designer, the current markup is what's desired as Cloudscape doesn't really support our use case.

React-hook-form doesn't really work well with the conditional dispatches I'm doing in `onChange` handlers, which is why there's some manual setValue and trigger calls.

## Legal


tracy-french:
I'll be taking over this PR and getting it merged.

PR #917, opened 2023-04-06
fix(composer): Set environment preset (lighting) to neutral on scene …

mumanity:
## Overview

- Sets the `Environment Preset` to `neutral` when initializing a scene for the first time. 
- Implements accessibility fix for Environment Preset settings. 
- Removes assignment of `undefined`.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/Environment.tsx
+++ b/packages/scene-composer/src/components/three-fiber/Environment.tsx
@@ -6,10 +6,18 @@ import NuetralHDR from '../../assets/hdri/Neutral_sm.hdr';
 import DirectionalHDR from '../../assets/hdri/Directional_sm.hdr';
 import ChromaticHDR from '../../assets/hdri/Chromatic_sm.hdr';
 
-export const presets = {
-  neutral: NuetralHDR as string,
-  directional: DirectionalHDR as string,
-  chromatic: ChromaticHDR as string,
+type Presets = {
+  noPreset: string;
+  neutral: string;
+  directional: string;
+  chromatic: string;
+};
+
+export const presets: Presets = {
+  noPreset: '',
+  neutral: NuetralHDR,
+  directional: DirectionalHDR,
+  chromatic: ChromaticHDR,
 };
 
 interface EnvironmentProps {
--- a/packages/scene-composer/src/components/panels/SettingsPanel.tsx
+++ b/packages/scene-composer/src/components/panels/SettingsPanel.tsx
@@ -6,7 +6,13 @@ import useLifecycleLogging from '../../logger/react-logger/hooks/useLifecycleLog
 import { presets } from '../three-fiber/Environment';
 import { sceneComposerIdContext } from '../../common/sceneComposerIdContext';
 import { useStore } from '../../store';
-import { COMPOSER_FEATURES, IValueDataBindingProvider, KnownComponentType, KnownSceneProperty } from '../../interfaces';
+import {
+  COMPOSER_FEATURES,
+  EnvironmentPreset,
+  IValueDataBindingProvider,
+  KnownComponentType,
+  KnownSceneProperty,
+} from '../../interfaces';
 import { pascalCase } from '../../utils/stringUtils';
 import { getGlobalSettings } from '../../common/GlobalSettings';
 import { Component } from '../../models/SceneModels';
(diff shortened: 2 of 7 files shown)


packages/scene-composer/src/components/WebGLCanvasManager.tsx
```diff
@@ -39,7 +39,8 @@ export const WebGLCanvasManager: React.FC = () => {
   const appendSceneNode = useStore(sceneComposerId)((state) => state.appendSceneNode);
   const { gl } = useThree();
   const domRef = useRef<HTMLElement>(gl.domElement.parentElement);
-  const environmentPreset = getSceneProperty(KnownSceneProperty.EnvironmentPreset);
+  const defaultLighting: string | undefined = useViewOptionState(sceneComposerId).defaultLighting;
```
tracy-french, line 42: Nit: I think it’s useful to have a ‘Maybe’ utility type available in the repo to handle non-existence. Then you can create some super useful type predicates.

```typescript
type Just<T> = T;
type Nothing = null | undefined;
type Maybe<T> = Just<T> | Nothing;

function isJust<T>(maybe: Maybe<T>): maybe is Just<T> {
  return maybe != null;
}

function isNothing<T>(maybe: Maybe<T>): maybe is Nothing {
  return !isJust(maybe);
}
```

This makes null checks a bit cleaner and can improve type safety, depending on the usage.

packages/scene-composer/src/store/slices/ViewOptionStateSlice.ts
```diff
@@ -10,10 +10,12 @@ export interface IViewOptionStateSlice {
   }>;
   tagSettings?: ITagSettings;
   enableMatterportViewer?: boolean;
+  defaultLighting?: string;
```
tracy-french, line 13: Is this truly a string that can be anything or should it be a literal string union (e.g., ‘neutral’ | ‘something’ | undefined) with a defined set of values?

packages/scene-composer/src/components/three-fiber/Environment.tsx
```diff
@@ -7,6 +7,7 @@ import DirectionalHDR from '../../assets/hdri/Directional_sm.hdr';
 import ChromaticHDR from '../../assets/hdri/Chromatic_sm.hdr';
 
 export const presets = {
+  noPreset: '' as string,
```
tracy-french, line 10: Not blocking and I see the pattern below of this, but I believe there should be a way to achieve the same with casting.

packages/scene-composer/src/components/WebGLCanvasManager.tsx
mumanity: I like the concept. The only pushback that I'd have is that I would like to see us start using `undefined` as it is meant to work in JavaScript - "Hey! Nothing was assigned to this." The problem with grouping it with `null` is that there is the `nullable` concept where we would intentionally assign `null` to a variable in place of a primitive, object, etc... By keeping these seprate, we're able to appropriately determine if a variable initialized without a value (`undefined`) or was intentionally empty (`null`). 

This was left with the `string` | `undefined` in order to improve readability that sometimes (initial scene creation) this value is truly `undefined`. 

Let me know if that doesn't make sense or I misunderstood the feedback. :)

packages/scene-composer/src/store/slices/ViewOptionStateSlice.ts
mumanity: If the parameter is passed, it will be a `string`.

packages/scene-composer/src/components/three-fiber/Environment.tsx
mumanity: I'll take a look & see what I can do here. :)

packages/scene-composer/src/store/slices/ViewOptionStateSlice.ts
mumanity: ... aaaand that just sank in. I can set it to use an `enum`.

packages/scene-composer/src/components/WebGLCanvasManager.tsx
tracy-french: Nope, I agree. I don't prefer to use null either. It's simply for the `Nothing` type itself that it includes null, as you can't control things like a library integration.

packages/scene-composer/src/components/three-fiber/Environment.tsx
mumanity: Fixed.

packages/scene-composer/src/store/slices/ViewOptionStateSlice.ts
mumanity: Fixed!

tracy-french:
@mumanity Could you add a comment here on the next steps for this change? Thank you! :)

mumanity:
> 

Updated the description with changes.

tracy-french (approved):
Looks great! Thank you for your work!

PR #924, opened 2023-04-07
feat(Composer): Add overlay for matterport tags

mukeshsahay:
## Overview
- Add data overlay for matterport tags with label and description (match with Matterport portal).
- Enabled overlay feature flag for Matterport story

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/hooks/useMatterportTags.ts
```diff
@@ -2,9 +2,32 @@ import { useCallback } from 'react';
...
-import { IAnchorComponentInternal, ISceneNodeInternal, useStore } from '../store';
+import { IAnchorComponentInternal, IDataOverlayComponentInternal, ISceneNodeInternal, useStore } from '../store';
 import { MattertagItem, TagItem } from '../utils/matterportTagUtils';
 import { RecursivePartial } from '../utils/typeUtils';
+import { Component } from '../models/SceneModels';
+import { generateUUID } from '../utils/mathUtils';
+
+const getContentForOverlayComponent = (label: string, description: string): string => {
+  return `#### **${label}**  
+${description}`;
```
sheilaXu, line 13: this indentation looks weird.. can we just add `\n` to make new line?

mukeshsahay: This is the preferred way from Markdown guideline.

packages/scene-composer/src/hooks/useMatterportTags.ts
```diff
@@ -27,6 +50,9 @@ const addTag = (
       matterportId: id, //mattertag uses item.sid and tag uses item.id so we just us the collection key for both
     },
   } as ISceneNode;
+
+  const dataoverlayComponent = getNewDataOverlayComponent(item);
+  node.components?.push(dataoverlayComponent);
```
sheilaXu, line 55: can we create the overlay before creating the node and directly adding to the components at line43? having this part separately will easily get missed

packages/scene-composer/src/hooks/useMatterportTags.ts
sheilaXu: also is that possible for the item to not have label and/or description? so we may only have one line or even no overlay at all?

mukeshsahay: Label will always be there.

mukeshsahay: Label is similar to the Tag name.

packages/scene-composer/src/hooks/useMatterportTags.ts
mukeshsahay: Done

packages/scene-composer/src/hooks/useMatterportTags.ts
mukeshsahay: Added a comment in code

sheilaXu (approved):

PR #932, opened 2023-04-10
feat(TwinMakerTools): added 2 optional flags to nuke for deleting s3 bucket and TM workspace

gsechkin-aws:
## Overview
Nuke extension for supporting s3 bucket + contents, IAM role, and workspace deletion. Updated README.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/tools-iottwinmaker/src/commands/nuke.ts
```diff
@@ -24,27 +28,61 @@ export const builder: CommandBuilder<Options> = (yargs) =>
     'workspace-id': {
       type: 'string',
       require: true,
-      description: 'Specify the ID of the Workspace to delete.',
+      description: 'Specify the ID of the Workspace to delete from.',
+    },
+    'delete-workspace': {
+      type: 'boolean',
+      require: false,
```
matwld, line 35: Can we set the default to false for these flags? That makes it obvious to developers what the default behavior is

packages/tools-iottwinmaker/src/commands/nuke.spec.ts
```diff
@@ -185,3 +195,53 @@ it('deletes all twinmaker resources when given a populated workspace', async ()
     sceneId: oneSceneListScenesResp.sceneSummaries![0].sceneId,
   });
 });
+
+it('deletes all twinmaker resources, s3 bucket, IAM role, workspace when given delete-workspace and delete-s3-bucket flags', async () => {
```
matwld, line 199: Does command still delete IAM role? If not please rename test case

packages/tools-iottwinmaker/src/commands/nuke.ts
```diff
@@ -57,11 +95,33 @@ export const handler = async (argv: Arguments<Options>) => {
...
+        // first retrieve s3 and role information before workspace deletion
+        const { s3Location, role } = await aws().tm.getWorkspace({ workspaceId });
+        // s3Location format: arn:aws:s3:::{s3BucketName}
+        const s3BucketName = s3Location?.split(':').pop();
+        // role format: arn:aws:iam::accountID:role/{roleName}
+        const roleName = role?.split(/:|\//).pop();
+        const deleteWorkspaceResp: DeleteWorkspaceCommandOutput | undefined = await deleteWorkspace(workspaceId);
+
+        // only proceed if workspace deletion was successful
+        if (deleteWorkspaceResp != undefined) {
```
matwld, line 110: Nit: we can use:
`if (deleteWorkspaceResp)`

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -318,5 +326,142 @@ async function createWorkspaceIfNotExists(workspaceId: string) {
...
+ * @param workspaceId TM workspaceID
+ * @returns DeleteWorkspaceCommandOutput if successful
+ */
+async function deleteWorkspace(workspaceId: string): Promise<DeleteWorkspaceCommandOutput | undefined> {
+  try {
+    const deleteWorkspaceResp: DeleteWorkspaceCommandOutput = await aws().tm.deleteWorkspace({ workspaceId });
+    console.log(`deleted workspace: ${workspaceId}`);
+    return deleteWorkspaceResp;
+  } catch (err) {
+    console.log(`Unable to delete workspace ${workspaceId}:\n ${err}`);
```
matwld, line 410: Should we just throw an error here? If customer stated to delete workspace and it fails we should make that obvious

packages/tools-iottwinmaker/README.md
```diff
@@ -62,7 +74,12 @@ tmdt init --region us-east-1 --workspace-id CookieFactory0601 --out /tmp/testpro
 
 ___
 
-## Deploy a tmdt project to another workspace
+## TMDT Deploy
+
+### Deploy Parameters (required)
+- *--region*: Specify the AWS region to deploy to
+- *--workspace-id*: Specify the ID of the Workspace to deploy to
+- *--dir*: SSpecify the project location, directory for tmdt.json file
```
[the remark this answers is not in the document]
gsechkin-aws: good catch, will update

packages/tools-iottwinmaker/README.md
```diff
@@ -90,6 +117,18 @@ For example:
...
+Another example:
+
+The following will delete all entities, component types, scenes, workspace s3 bucket, and TwinMaker workspace itself.
+
+**Warning:** This command is destructive. Please take caution before running this command.
+```
+tmdt nuke --region us-east-1 --workspace-id SyncB --delete-workspace --delete-s3-bucket
+```
+**Additional notes:** 
+1. *--delete-s3-bucket* implies *--delete-workspace*; you will not be able to delete a workspace s3 bucket without deleting the workspace itself (however you may delete the workspace without deleting the s3 bucket)
```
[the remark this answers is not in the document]
gsechkin-aws: That actually is the case. If a user attempts to set delete-s3-bucket without workspace, the command will not run bc I set "implies()" in yargs command builder, it will prompt user "missing dependency delete-s3-bucket -> delete-workspace"

packages/tools-iottwinmaker/src/commands/nuke.ts
gsechkin-aws: I'm not sure what this means, do you mean 
`default: false`?
By default (the flag is not included at all since it is not required), it is false?
Currently an example of intended behavior is: "tmdt init --region <region> --workspace-id <ws> --delete-workspace". Do you want to explicitly set "--delete-workspace true" instead? Not sure if that is any different.

packages/tools-iottwinmaker/src/commands/nuke.spec.ts
gsechkin-aws: Will update!

packages/tools-iottwinmaker/README.md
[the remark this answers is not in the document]
matwld: Awesome, well designed

packages/tools-iottwinmaker/src/commands/nuke.ts
matwld: Yes I mean explicitly state it. The current behavior is correct but not clear to developers

matwld (approved):

packages/tools-iottwinmaker/README.md
```diff
@@ -3,24 +3,29 @@
...
 
+You can follow [AWS IoT TwinMaker Getting Started](https://github.com/aws-samples/aws-iot-twinmaker-samples) to setup a sample TwinMaker workspace.
 
-TMDT can snapshot an existing TwinMaker workspace (init), deploy this snapshot into any other existing workspace or automatically create a new workspace, S3bucket, and workspace role to deploy into (deploy), and easily delete all entities, component types, and scenes from a workspace (nuke).
 
-You can follow [AWS IoT TwinMaker Getting Started](https://github.com/aws-samples/aws-iot-twinmaker-samples) to setup a sample TwinMaker workspace.
+TMDT has 3 core commands:
+1. TMDT init - create a TMDT snapshot an existing TwinMaker workspace into a specifified directory
+2. TMDT deploy - deploy a TMDT snapshot into any other existing workspace or automatically create a new workspace, S3bucket, and workspace role to deploy into
+3. TMDT nuke - delete all entities, component types, and scenes from a workspace, with additional flags for deleting workspace s3 bucket + contents, and TwinMaker workspace
```
johnnyw-aws, line 12: shall we rename `nuke` to `destroy`?

packages/tools-iottwinmaker/README.md
```diff
@@ -48,7 +53,14 @@ tmdt -h
...
+## TMDT Init
+
+### Init Parameters (required)
+- *--region*: Specify the AWS region of the Workspace to bootstrap the project from
+- *--workspace-id*: Specify the ID of the Workspace to bootstrap the project from
+- *--out*: Specify the directory to initialize a project in
+
+Use this command to bootstrap a TMDT project from an existing workspace.
 
 The following will initialize a tmdt project at the specified directory with a `tmdt.json` file
```
johnnyw-aws, line 66: "CookieFactory0601" and "SyncB" should be replaced with something more generic, maybe just "YourWorkspace" ?

Same for "testproj"

packages/tools-iottwinmaker/src/commands/nuke.ts
```diff
@@ -24,27 +28,63 @@ export const builder: CommandBuilder<Options> = (yargs) =>
...
+    const nukeResponse = await prompt({
       type: 'text',
       name: 'confirmation',
       message: `Are you sure you wish to delete all entities, component types, and scenes from IoT TwinMaker workspace [${workspaceId}] in region [${region}] ? (Y/n)`,
     });
+    if (deleteS3Flag) {
+      const promptResp = await prompt({
+        type: 'text',
+        name: 'confirmation',
+        message: `Are you sure you wish to delete the workspace [${workspaceId}], its associated workspace S3 bucket, and any potential logging bucket ? (Y/n)`,
```
johnnyw-aws, line 69: Is it easy to also output some information about the bucket? e.g. what files are in it and how many, maybe something like:

```
workspace S3 bucket to be deleted: <bucket name>
contents to be deleted:
file1.json
file2.json
file3.json
<...up to 10 files...>
...with <N> files total, total bucket size: <M>

Are you sure you wish to delete the workspace [${workspaceId}], its associated workspace S3 bucket, and any potential logging bucket ?
```

packages/tools-iottwinmaker/README.md
```diff
@@ -74,13 +91,23 @@ For example:
...
+- *--workspace-id*: Specify the ID of the Workspace to delete from
 
-Warning: This command is destructive. Please take caution before running this command.
+### Nuke Parameters (optional flags)
+- *--delete-workspace*: Specify if TM workspace should also be deleted
+- *--delete-s3-bucket*: Specify if workspace s3 Bucket, its contents, and any associated logging bucket should be deleted
+
+The following will delete all entities, component types, and scenes from a workspace so that the workspace can be deleted. By default (no optional flags below selected), this will not delete workspace s3 bucket, IAM role, nor the TwinMaker workspace itself.
+
+**Warning:** This command is destructive. Please take caution before running this command.
```
johnnyw-aws, line 110: Is it possible to make it default to a `dry-run-mode`? i.e. they can run the script and it will give the prompts etc. but when they select "Y" it will only print out what would be deleted without actually performing the deletion

Only if they run with `--non-dry-run-mode` will it behave as it is currently

packages/tools-iottwinmaker/src/lib/workspace.ts
johnnyw-aws: What was the discrepancy in verify? Looks like it catches to give custom message but does throw the error back out? https://github.com/awslabs/iot-app-kit/blob/main/packages/tools-iottwinmaker/src/lib/utils.ts#L19

If we just catch and log will the script return a success (e.g. `0`) error code (which doesn't seem right?)

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -318,5 +326,142 @@ async function createWorkspaceIfNotExists(workspaceId: string) {
...
+    isTruncated = listObjectsResp.IsTruncated;
+    NextContinuationToken = listObjectsResp.NextContinuationToken;
+    if (listObjectsResp.Contents != undefined) {
+      for (const obj of listObjectsResp.Contents) {
+        await aws().s3.deleteObject({ Bucket: s3BucketName, Key: obj['Key'] });
+        console.log(`deleted S3 Object: ${obj['Key']}`);
+      }
+    }
+  }
+  // Delete all versions in bucket if bucket has versioning
```
johnnyw-aws, line 436: Just want to confirm: for AWS CLI there are high-level commands to delete the whole bucket (e.g. with `rm` [[1]](https://docs.aws.amazon.com/cli/latest/userguide/cli-services-s3-commands.html)) I'm guessing they're not available with the SDK?

johnnyw-aws (commented):
Thanks for getting this feature in! Adding some comments so far

packages/tools-iottwinmaker/src/lib/workspace.ts
gsechkin-aws: There is no discrepancy right now (sorry for confusion), we just had a similar change in that function (I was told in a JS course that it's always best practice to catch any known potential errors such as API calls). But now that you mention it, returning the 0 error code doesn't make sense, you're right I will update.

packages/tools-iottwinmaker/README.md
gsechkin-aws: Yep, will update

packages/tools-iottwinmaker/README.md
gsechkin-aws: makes sense

packages/tools-iottwinmaker/README.md
gsechkin-aws: Yes we can do this, although the easiest way I think to do this is passing in a "dry-run" boolean as an extra parameter into every single delete helper function, and reworking all the functions to call get instead of delete if that boolean is true.

Also, do you mean dry run is enabled by default for EVERY combination of nuke (regardless of what delete-workspace or delete-s3-bucket flags are there), so that it warns about entities, component types, scenes? We actually don't display which entities get deleted in "non dry run" but rather just the parent and print "waiting for recursion" so may need to consider how dry run works on entities?

Also, this change might take up to a dev day to implement and test properly, how high priority is this, compared to instead listing the s3 bucket contents before deletion like the above comment (and doesn't this conflict with that idea if we add both since they're both a form of letting the user know what gets deleted?)

packages/tools-iottwinmaker/src/commands/nuke.ts
gsechkin-aws: Great idea! I'll look into all the get API options for S3 tmr morning and experiment with how easy this would be

packages/tools-iottwinmaker/src/lib/workspace.ts
gsechkin-aws: Correct as far as I have seen. Following rough steps outlined here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-bucket.html
(this version is in Java though but same process w/ SDK)

packages/tools-iottwinmaker/src/commands/nuke.ts
gsechkin-aws: Update: there is no easy way to list the total number of files in an s3 bucket, especially using SDK: [Stack Overflow].(https://stackoverflow.com/questions/2862617/how-can-i-tell-how-many-objects-ive-stored-in-an-s3-bucket#:~:text=If%20you%20use%20the%20s3cmd,it%20to%20a%20text%20file.&text=Then%20in%20linux%20you%20can,(1%20line%20per%20object).&text=The%20%2Dr%20in%20the%20command,for%20sub%2Dfolders%20as%20well.&text=3%20notes%20on%20this.) 
I can list the first few (like 10 as you mentioned) files, but that may not be indicative if we don't have the total #?
Unfortunately SDK does not have the cli command "aws s3 ls "s3://MyBucket" --summarize", similar to how there is no easy rm command mentioned below.

johnnyw-aws: >  the solution it seems is using ListObjectsV2, but that is limited to the first 1,000 objects

Could we do an interim approach: e.g. list the first 10 then say the count of remaining if less than 1k otherwise say something like "1000+ objects"

fyi: am ok iterating on it too in next revision if may take more additional time

packages/tools-iottwinmaker/README.md
johnnyw-aws: > do you mean dry run is enabled by default for EVERY combination of nuke

I'm not sure I understood the combinations aspect: I meant it as a global option. If `non-dry-run` isn't passed in, no delete operations occur regardless of what combination of delete operations are requested

> but rather just the parent and print "waiting for recursion" so may need to consider how dry run works on entities?

good point, I think we can maintain the existing approach in dry-run: basically just print that we would do recursive deletion on [entity name] -- could maybe also do something similar for the s3 one and call list-entities a few times to get a count (or print a cap of like 1000+ entities again)

> how high priority is this

I personally would put it fairly high since delete is a non-recoverable operation so would bias towards "safe default behavior" but am open to getting other opinions, especially if may take some time to implement

packages/tools-iottwinmaker/src/lib/workspace.ts
```diff
@@ -448,19 +447,21 @@ async function deleteS3Bucket(s3BucketName: string | undefined) {
     isTruncated = listObjectVersionResp.IsTruncated;
     if (listObjectVersionResp.Versions != undefined) {
       for (const version of listObjectVersionResp.Versions) {
-        await aws().s3.deleteObject({ Bucket: s3BucketName, Key: version['Key'], VersionId: version['VersionId'] });
+        if (nonDryRun)
+          await aws().s3.deleteObject({ Bucket: s3BucketName, Key: version['Key'], VersionId: version['VersionId'] });
```
johnnyw-aws, line 451: fyi: I noticed some inconsistencies in the formatting, some places we have the `if (nonDryRun)` on single line and other places is next line with indent, let's make them consistent?

~~to clarify other comment: I think in areas like this we can have an `else` that is like: "would delete s3://<bucketname>/<key> if non-dry-run"~~ edit: actually I guess this would end up having to be plumbed through all the messages which may overcomplicate things, maybe let's just add a message at the start indicating that "all delete messages below are not actually performing delete operations"

gsechkin-aws: True, those inconsistencies are due to line length sometimes forcing an indent by ESLint. I can make it more consistent by adding braces around all if statements. I tried to be consistent by having no braces because in all cases it is a single command within the if, but perhaps that's takes away from readability.

As for second comment, I agree it might be confusing; I added a log "\nDry Run: None of the following resources will be deleted. Use --nonDryRun flag for real deletion.\n" to hopefully clarify, but open to suggestions on how to make it more clear!

johnnyw-aws (approved):

PR #946, opened 2023-04-11
fix(composer): Restores loading indicator for scene-composer

mumanity:
## Overview
Bug fix to restore loading indicator in scene composer. 

Additionally included minor fix to remove double-border appearing on action icons in the Scene Hierarchy panel. 

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/Tree/tree.scss
+++ b/packages/scene-composer/src/components/Tree/tree.scss
@@ -15,7 +15,6 @@ $active-color: #ec7211;
   background: $active-bg;
 }
 @mixin selected-action {
-  @include selected;
   background: $active-bg-gradient;
   margin-top: -1px;
   margin-bottom: -1px;
--- a/packages/scene-composer/src/layouts/SceneLayout/SceneLayout.tsx
+++ b/packages/scene-composer/src/layouts/SceneLayout/SceneLayout.tsx
@@ -47,52 +47,44 @@ const R3FWrapper = (props: {
   children?: any;
   sceneLoaded?: boolean;
 }) => {
+  const { children, sceneLoaded, enableMatterport, matterportConfig } = props;
   const sceneComposerId = useContext(sceneComposerIdContext);
   const ContextBridge = useContextBridge(LoggingContext, sceneComposerIdContext, ThemeContext);
+  const loadMatterPort = sceneLoaded && enableMatterport && matterportConfig?.modelId;
 
-  if (!props.sceneLoaded) {
-    return null;
-  }
-
-  if (props.enableMatterport && props.matterportConfig?.modelId) {
-    return (
-      <MatterportViewer
-        assetBase={props.matterportConfig?.assetBase}
-        m={props.matterportConfig?.modelId}
-        applicationKey={props.matterportConfig?.applicationKey}
-        connect-auth={props.matterportConfig?.accessToken}
-        connect-provider='iot-twinmaker'
-        onReady={(matterportSdk: MpSdk) => {
-          // propagate this out elsewhere if you wish to useMatterportSdk() interface
-          // to control this viewer from other non-3d ui
-          // <MpSdkContext.Provider value={matterportSdk}>
-          //  <Other2DComponents/>
-          // <MpSdkContext.Provider/>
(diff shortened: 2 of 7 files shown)


packages/scene-composer/src/components/Tree/tree.scss
```diff
@@ -15,7 +15,6 @@ $active-color: #ec7211;
   background: $active-bg;
 }
 @mixin selected-action {
-  @include selected;
```
TheEvilDev, line 18: What's this for?

TheEvilDev (dismissed):

mukeshsahay (approved):

packages/scene-composer/src/layouts/SceneLayout/SceneLayout.tsx
```diff
@@ -47,52 +47,46 @@ const R3FWrapper = (props: {
...
-        play={1}
-        mls={1}
-      >
-        <ContextBridge>
-          <Suspense fallback={null}>{props.children}</Suspense>
-        </ContextBridge>
-      </MatterportViewer>
-    );
-  } else {
-    setMatterportSdk(sceneComposerId, undefined);
```
sheilaXu, line 87: this line is removed? we probably won't be able to go back to standard scene without this line

sheilaXu (changes requested):
confirm if removing `setMatterportSdk(sceneComposerId, undefined);` will cause regression

packages/scene-composer/src/layouts/SceneLayout/SceneLayout.tsx
mukeshsahay: Good catch. Need to add this line back.

packages/scene-composer/src/components/Tree/tree.scss
mumanity: It's a CSS `mixin` that was adding the matching border for the `.action` class. Looks like there was an update from a dependency that modified the sizing so currently there's a nested border showing above & below the action buttons on hover/selected state. Removing the `mixin` to resolve the issue.

packages/scene-composer/src/layouts/SceneLayout/SceneLayout.tsx
mumanity: This line was creating a console error within React.  Does it not initialize as `undefined`? 

The Block that this was included in was not rendering anything from what I could tell. Once the value comes in for `loadMatterPort` then it will trigger a re-render & return the condition that instead passes `matterportSdk`. 

Can one of you show me how to test for this scenario? 
"we probably won't be able to go back to standard scene without this line"

I'd like to make sure that I'm fully testing this.

sheilaXu: actually I forgot to comment that line during later testing.. and I can see issue by
1. create a scene with camera view and matterport setup, observe the cameras dropdown on the top bar
2. load the matterport view and observe the dropdown is removed
3. remove the matterport model id from setting, the scene goes back to standard one, but the cameras dropdown is not coming back

mukeshsahay: @mumanity and I worked on this and have added a fix.

mukeshsahay (approved):

sheilaXu (approved):

packages/scene-composer/src/layouts/SceneLayout/SceneLayout.tsx
```diff
@@ -102,15 +94,14 @@ interface SceneLayoutProps {
   showMessageModal: boolean;
   externalLibraryConfig?: ExternalLibraryConfig;
 }
-
 const SceneLayout: FC<SceneLayoutProps> = ({
   isViewing,
+  onPointerMissed,
```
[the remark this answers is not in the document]
mumanity: Not sure how that got added back. Just pushed the removal of it. Thanks for catching!

sheilaXu (approved):

PR #948, opened 2023-04-12
fix(composer): fix ViewCursorWidget for 3D Tiles

hwandersman:
fix ViewCursorWidget for 3D Tiles and made cursor size dynamic based on camera distance

## Overview
When a user wants to add a tag, motion indicator, or model to the scene with enhanced editing, the mouse becomes a cursor with a projected SVG to show where the widget/object will be placed in relation to the surface of another object. A raycast from the camera to the mouse is used to calculate the intersection distance, position, and orientation for an object in the scene.

For regular 3D models this works fine, but the raycast for 3D Tiles did not work. After investigating I found that the `raycast` method for the 3D Tiles meshes in the scene is not called from the mesh object but from the TilesGroup which is a parent of all meshes in the tileset.

Updated the logic for traversing the scene and calculating the intersection of rays to objects. The list of objects that will call `raycast` are just any TilesGroup node and any non-tileset and non-gizmo meshes.

The size of the projected SVG is a static size which is difficult to see for large objects and a camera that is far away. It is too small and looks like the feature is broken. With this change I update the SVG scale based on the distance of the closest intersection.

## Verifying Changes
Verified in storybook with 3D Tilesets I have in my personal AWS account.

1. Scene with small 3D Tiles object and regular gltf object
2. Scene with large (in scale to the scene) 3D Tiles object and regular gltf object
3. Scene with small gltf object:

https://user-images.githubusercontent.com/87385528/231552392-a69d3b85-36ae-4727-a64b-f2acc63dd235.mov

4. Scene with 3D Tiles:

https://user-images.githubusercontent.com/87385528/231552458-cb8c1ff7-1b1e-4a67-b10f-6ff079c1ea44.mov

Test:
1. Select to add a tag
2. Hover over 3D Tiles object and gltf object to see the cursor
6. Zoom in and out and pan with the camera to see the cursor remain a visible and somewhat consistent size
7. Click on the 3D Tiles to place the tag

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
@@ -1,12 +1,17 @@
 import React, { useCallback, useContext, useEffect, useMemo, useRef } from 'react';
-import { useFrame, useThree } from '@react-three/fiber';
-import { Mesh as THREEMesh, Object3D as THREEObject3D, Vector3 as THREEVector3 } from 'three';
+import { useFrame, useLoader, useThree } from '@react-three/fiber';
+import { Mesh as THREEMesh, Object3D as THREEObject3D, Vector3 as THREEVector3, Vector3 } from 'three';
+import { SVGLoader } from 'three-stdlib';
+import { TilesGroup } from '3d-tiles-renderer';
 
-import { convertSvgToMesh } from '../../../../utils/svgUtils';
+import { convertSvgToMesh, getDataUri } from '../../../../utils/svgUtils';
 import { getIntersectionTransform } from '../../../../utils/raycastUtils';
 import { sceneComposerIdContext } from '../../../../common/sceneComposerIdContext';
 import { useEditorState } from '../../../../store';
 import { ViewCursorEditSvgString, ViewCursorMoveSvgString } from '../../../../assets/svgs';
+
+export const INIT_SVG_SCALE = 0.003;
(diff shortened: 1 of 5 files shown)


packages/scene-composer/src/augmentations/components/three-fiber/viewpoint/ViewCursorWidget.tsx
```diff
@@ -1,23 +1,26 @@
...
+import { TilesGroup } from '3d-tiles-renderer';
 
-import { convertSvgToMesh } from '../../../../utils/svgUtils';
+import { convertSvgToMesh, getDataUri } from '../../../../utils/svgUtils';
 import { getIntersectionTransform } from '../../../../utils/raycastUtils';
 import { sceneComposerIdContext } from '../../../../common/sceneComposerIdContext';
 import { useEditorState } from '../../../../store';
 import { ViewCursorEditSvgString, ViewCursorMoveSvgString } from '../../../../assets/svgs';
 
+export const INIT_SVG_SCALE = 0.003;
```
sheilaXu, line 13: is this going to be huge for smaller scene?

hwandersman: No it's always around the same size. I added a video in the description to show that.

sheilaXu (dismissed):

sheilaXu (approved):

PR #970, opened 2023-04-14
fix(react-components): useTimeSeriesData hook works when number of queries changes

corteggiano:
## Overview
React error was happening when the number of queries passed into useEffect changed
`The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.`

Now useEffect will check against a concatenated string of all the queries so it will always have a length of 1

## Legal


tracy-french (approved):
Nothing makes me happier than seeing a tiny change with magnitude more lines of tests. Thank you!

PR #1011, opened 2023-04-25
fix(CrashOnRemount): scene-composer doesn't crash when remounted now

TheEvilDev:
## Overview
This fixes a bug in the SceneComposer component when anything would cause the component to be unmounted and remounted, it would crash and require a page refresh.

There's some complicated DOM binding in the OrbitControls bit that was losing the dom element when it was destroyed and recreated, fixing this issue so it was correctly bound to a component resolved the issue.

## Verifying Changes

I added a manual verification test (mounting the entire scene composer in a unit test is a bit crazy), you can see in the storybook output under "scene-viewer" tests. There's a toggle switch that will mount and unmount the component as you toggle it.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/controls/OrbitControls.tsx
+++ b/packages/scene-composer/src/components/three-fiber/controls/OrbitControls.tsx
@@ -47,7 +47,7 @@ export const OrbitControls = forwardRef<OrbitControlsImpl, OrbitControlsProps>(
         if (onChange) onChange(e);
       };
 
-      controls.connect(explDomElement!);
+      controls.connect(explDomElement || gl.domElement);
       controls.addEventListener('change', callback);
 
       if (onStart) controls.addEventListener('start', onStart);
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/components/three-fiber/controls/OrbitControls.tsx
```diff
@@ -47,7 +47,7 @@ export const OrbitControls = forwardRef<OrbitControlsImpl, OrbitControlsProps>(
         if (onChange) onChange(e);
       };
 
-      controls.connect(explDomElement!);
+      controls.connect(explDomElement || gl.domElement);
```
TheEvilDev, line 50: This is the magic that fixes the issue.

tracy-french: :chefs-kiss:

tracy-french (approved):
Nice fix! Well done @TheEvilDev!

PR #1012, opened 2023-04-25
fix(dashboard): remove box-intersect dependency

jmbuss:
## Overview
Refactor of code to remove the box-intersect dependency. Rename the isContained util to overlaps to be more specific about the functionality. The isContained function returned true in an overlap existed between the 2 rectangles as opposed to returning true if a was fully contained by b or vice versa. All of the same tests for isContained remain and are now run under the overlaps.spec file.

## Legal


square-li (approved):

chejimmy (approved):
🙏 for the quick turnaround

PR #1025, opened 2023-04-28
feat(Knowledge Graph Component): graph data visualizer

MO-Elmu:
## Overview
This is the first PR for the TwinMaker Knowledge Graph AppKit Component. Code here is laying the ground for graph visualization react component that uses Cytoscape library. The code also leverages a lot of classes from the CookieFactory graph visualization component by Chris Bolen. Code has been added to and modified to become more generic and fit the 3D KG component needs. 

## Verifying Changes
npm run build and Code also verified in a local App. The storybook for this component and unit tests will come in the following PRs. We have been given the green light to relax unit tests to rush out this feature.

Screenshot 2023-04-27 at 9 29 22 PM

## Legal


packages/react-components/src/graph-visualizer/constants.ts
```diff
@@ -0,0 +1,43 @@
+export const GRAPH_COLORS = {
```
sheilaXu, line 1: should we import the colors from awsui design tokens so we can get light & dark mode support?

MO-Elmu: Yes, I am switching to that. its coming in the next PR.

packages/react-components/src/graph-visualizer/svg.ts
```diff
@@ -0,0 +1,185 @@
...
+  borderColor,
+  borderWidth,
+  shape,
+  size,
+}: Pick<NodeStyleProps, 'backgroundColor' | 'borderColor' | 'borderWidth'> & {
+  shape?: NodeShape;
+  size: number;
+}) {
+  switch (shape) {
+    case 'diamond':
```
haweston, line 26: Is the cytoscape NodeShape type have enums that we can do the case matches on instead of strings? If they're using a union string type then this is fine as is.

packages/react-components/src/graph-visualizer/graph.ts
```diff
@@ -0,0 +1,523 @@
+/* eslint-disable prefer-const */
```
haweston, line 1: This file was large enough github didn't want to load it for the diff review so it's probably worth seeing if we can break it up.

packages/react-components/src/graph-visualizer/graph.ts
```diff
@@ -0,0 +1,523 @@
...
+};
+
+const NODE_SELECTED_STYLE_PROPS: NodeStyleProps = {
+  ...NODE_DEFAULT_STYLE_PROPS,
+  backgroundImage: getSelectedSvg,
+  borderColor: GRAPH_COLORS.WHITE,
+  borderWidth: 6,
+  labelTextColor: GRAPH_COLORS.WHITE,
+};
+
```
haweston, line 91: I think all these prop functions can be split to a separate file.

packages/react-components/src/graph-visualizer/cytoGraph.tsx
```diff
@@ -0,0 +1,105 @@
...
+    if (graph) {
+      if (props.edgeData) {
+        graph.setGraphData(getElementsDefinition([...props.nodeData.values()], [...props.edgeData.values()]));
+      } else {
+        graph.setGraphData(getElementsDefinition([...props.nodeData.values()]));
+      }
+      graph.resize();
+      graph.center();
+    }
+  }, [graph]);
```
haweston, line 84: This may need props.nodeData and props.edgeData as dependencies for if the props change like in a case where multiple queries are run in the same session.

MO-Elmu: great catch, I wouldn’t have noticed it till I started integrating queries. Fixed.

packages/react-components/src/graph-visualizer/graph.ts
MO-Elmu: Addressed, moved to a separate defaultProps file.

packages/react-components/src/graph-visualizer/graph.ts
MO-Elmu: Done,

packages/react-components/src/graph-visualizer/svg.ts
MO-Elmu: They are using union string type,

packages/react-components/src/graph-visualizer/graph.ts
MO-Elmu: I am always in favor of code decomposition.

haweston (approved):
For next PR I might break up graph.ts a bit further, but this is fine for now:
-put getElementsDefinition in it's own elements file maybe? I could see maybe just keeping it in graph, but it's not call in graph any.
-put the Style functions at the bottom in their own file together.

PR #1045, opened 2023-05-05
feat(Knowledge Graph Comp): Knowledge Graph DataSource

MO-Elmu:
## Overview
This PR adds TwinMaker Knowledge Graph DataSource for the source-iottwinmaker package. The datasource rely on the ExecuteQueryCommand API.

## Legal

### DIFF
--- a/packages/source-iottwinmaker/src/initialize.ts
+++ b/packages/source-iottwinmaker/src/initialize.ts
@@ -16,6 +16,7 @@ import {
 } from './aws-sdks';
 import { S3SceneLoader } from './scene-module/S3SceneLoader';
 import { SceneMetadataModule } from './scene-module/SceneMetadataModule';
+import { KGDataModule } from './knowledgeGraph-module/KGDataModule';
 import { VideoDataImpl } from './video-data/VideoData';
 import { VideoDataProps } from './types';
 import { TwinMakerDataStreamQuery, TwinMakerQuery } from './time-series-data/types';
@@ -153,6 +154,7 @@ export const initialize = (
     s3SceneLoader: (sceneId: string) => new S3SceneLoader({ workspaceId, sceneId, twinMakerClient, s3Client }),
     sceneMetadataModule: (sceneId: string) =>
       new SceneMetadataModule({ workspaceId, sceneId, twinMakerClient, secretsManagerClient }),
+    kGDatamodule: () => new KGDataModule({ workspaceId, twinMakerClient }),
     videoData: (videoDataProps: VideoDataProps) =>
       new VideoDataImpl({
         workspaceId: workspaceId,
--- a/packages/source-iottwinmaker/src/knowledgeGraph-module/KGDataModule.ts
+++ b/packages/source-iottwinmaker/src/knowledgeGraph-module/KGDataModule.ts
@@ -1,0 +1,51 @@
+import {
+  ExecuteQueryCommand,
+  ExecuteQueryCommandOutput,
+  ExecuteQueryCommandInput,
+  IoTTwinMakerClient,
+  ColumnDescription,
+  Row,
+} from '@aws-sdk/client-iottwinmaker';
+import { KGDataModuleInput, executeQueryParams } from './types';
+import { TwinMakerKGQueryDataModule } from '../types';
+
+export class KGDataModule implements TwinMakerKGQueryDataModule {
+  private twinMakerClient: IoTTwinMakerClient;
+  private workspaceId: string;
+
+  constructor(input: KGDataModuleInput) {
+    this.twinMakerClient = input.twinMakerClient;
+    this.workspaceId = input.workspaceId;
+  }
+
+  executeQuery = async ({
+    queryStatement,
+    resultsPerPage,
+    maxPagesCount,
+  }: executeQueryParams): Promise<ExecuteQueryCommandOutput> => {
+    const rows: Row[] = [];
+    let columnDescriptions: ColumnDescription[] = [];
+    let $metadata = {};
+    let pageCounter = 0;
+    const commandInput: ExecuteQueryCommandInput = {
+      workspaceId: this.workspaceId,
+      queryStatement: queryStatement,
+      maxResults: resultsPerPage,
+    };
+    do {
+      const response: ExecuteQueryCommandOutput = await this.twinMakerClient.send(
+        new ExecuteQueryCommand(commandInput)
+      );
+      pageCounter = pageCounter + 1;
+      if (response.rows) {
+        rows.push(...response.rows);
+      }
+      if (response.columnDescriptions && columnDescriptions.length === 0) {
+        columnDescriptions = response.columnDescriptions;
+      }
+      if (response.$metadata && Object.keys($metadata).length === 0) $metadata = response.$metadata;
+      commandInput.nextToken = response.nextToken;
+    } while (maxPagesCount ? commandInput.nextToken && pageCounter < maxPagesCount : commandInput.nextToken);
+    return { rows, columnDescriptions, $metadata };
+  };
+}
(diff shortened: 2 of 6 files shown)


packages/source-iottwinmaker/src/initialize.ts
```diff
@@ -153,6 +154,7 @@ export const initialize = (
     s3SceneLoader: (sceneId: string) => new S3SceneLoader({ workspaceId, sceneId, twinMakerClient, s3Client }),
     sceneMetadataModule: (sceneId: string) =>
       new SceneMetadataModule({ workspaceId, sceneId, twinMakerClient, secretsManagerClient }),
+    kGDatamodule: (queryStatement: string) => new KGDataModule({ workspaceId, twinMakerClient, queryStatement }),
```
sheilaXu, line 157: should the `queryStatement` be a parameter of the executeQuery function so when the user want to execute multiple different queries, they will do 
```
const kg = initialize(...).kGDatamodule();
...
kg.executeQuery('SELECT A....')
kg.executeQuery('SELECT B....')
kg.executeQuery('SELECT C....')
```
instead of 
```
const ds = initialize(...)
...
ds.kGDatamodule('SELECT A....').executeQuery()
ds.kGDatamodule('SELECT B....').executeQuery()
ds.kGDatamodule('SELECT C....').executeQuery()
```

MO-Elmu: I was debating the same thing, I think you are right your suggestion is better so we can we have one instance of the class. Will fix and re-commit.

packages/source-iottwinmaker/src/knowledgeGraph-module/KGDataModule.ts
```diff
@@ -0,0 +1,48 @@
...
+      );
+      if (response.rows) {
+        rows.push(...response.rows);
+      }
+      if (response.columnDescriptions && columnDescriptions.length === 0) {
+        columnDescriptions = response.columnDescriptions;
+      }
+      if (response.$metadata && Object.keys($metadata).length === 0) $metadata = response.$metadata;
+      commandInput.nextToken = response.nextToken;
+    } while (this.maxResultsCount ? commandInput.nextToken && this.maxResultsCount : commandInput.nextToken);
```
haweston, line 45: It doesn't look like we're submitting maxResults as an API parameter or checking it's value to see if it's exceed manually?

MO-Elmu: The maxResults here is to allow user to control large results and the intention is not related to the maxResult API parameter as the results are paginated and this one refer to the max results to be returned per page. The ExecuteQuery Request maxResult param has default 25 and max 250 which the number of results per page. That means if a use sets maxResultsCount=10 will still get 25 results as this is the minimum default according to pagination.

MO-Elmu: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-iottwinmaker/interfaces/executequerycommandinput.html#maxresults

MO-Elmu: so the you will always get multiples of 25 by default if you set maxResultsCount=26 you will still get 50 results. and Thats fine and explained well in the documentation. just to avoid the case when you have 10,000 results this gives you the control to round up to the closest multiple of 25.

packages/source-iottwinmaker/src/knowledgeGraph-module/KGDataModule.ts
```diff
@@ -0,0 +1,44 @@
...
+      );
+      if (response.rows) {
+        rows.push(...response.rows);
+      }
+      if (response.columnDescriptions && columnDescriptions.length === 0) {
+        columnDescriptions = response.columnDescriptions;
+      }
+      if (response.$metadata && Object.keys($metadata).length === 0) $metadata = response.$metadata;
+      commandInput.nextToken = response.nextToken;
+    } while (maxResultsCount ? commandInput.nextToken && maxResultsCount : commandInput.nextToken);
```
haweston, line 41: I think this line reduces to just:

} while (commandInput.nextToken);

which means maxResultsCount can be dropped as a parameter as it has no impact?

sheilaXu: +1

sheilaXu:
please also make sure your PR only have one commit

packages/source-iottwinmaker/src/knowledgeGraph-module/KGDataModule.ts
MO-Elmu: oops what I meant is rows.length < maxResultsCount
 (maxResultsCount ? commandInput.nextToken && rows.length < maxResultsCount : commandInput.nextToken)
I fixed it.

packages/source-iottwinmaker/src/knowledgeGraph-module/KGDataModule.ts
```diff
@@ -0,0 +1,44 @@
...
+export class KGDataModule implements TwinMakerKGQueryDataModule {
+  private twinMakerClient: IoTTwinMakerClient;
+  private workspaceId: string;
+
+  constructor(input: KGDataModuleInput) {
+    this.twinMakerClient = input.twinMakerClient;
+    this.workspaceId = input.workspaceId;
+  }
+
+  executeQuery = async (queryStatement: string, maxResultsCount?: number): Promise<ExecuteQueryCommandOutput> => {
```
sheilaXu, line 21: should `maxResultsCount` be `maxPagesCount` instead? otherwise we should update the code to not return more than `maxResultsCount` number of rows

MO-Elmu: addressed.

packages/source-iottwinmaker/src/knowledgeGraph-module/KGDataModule.ts
```diff
@@ -0,0 +1,58 @@
...
+        columnDescriptions = response.columnDescriptions;
+      }
+      if (response.$metadata && Object.keys($metadata).length === 0) $metadata = response.$metadata;
+      commandInput.nextToken = response.nextToken;
+    } while (
+      maxPagesCount
+        ? commandInput.nextToken &&
+          (resultsPerPage && resultsPerPage > 0
+            ? rows.length / resultsPerPage < maxPagesCount
+            : rows.length / MAX_RESULTS_PER_PAGE < maxPagesCount)
```
sheilaXu, line 53: I think you should have a counter for the number of pages received instead of using the rows to calculate. 

this calculation relies on the assumptions that for each previous pages, the backend will always return the max number of rows, and also the default max result is always 25 and won't change in the future. These assumptions make this calculation easy to break.

MO-Elmu: Hmm, These assumptions are based on API contracts I feel it is a stretch to account for if API contract is changed. The backend should always return the max number of rows except for the last page if there is no more data and this case is guarded by the fact there will be no next token. However I like the idea of page counter here is how the code will look like let me know 

do {
      const response: ExecuteQueryCommandOutput = await this.twinMakerClient.send(
        new ExecuteQueryCommand(commandInput)
      );
      pageCounter = pageCounter + 1;
      if (response.rows) {
        rows.push(...response.rows);
      }
      if (response.columnDescriptions && columnDescriptions.length === 0) {
        columnDescriptions = response.columnDescriptions;
      }
      if (response.$metadata && Object.keys($metadata).length === 0) $metadata = response.$metadata;
      commandInput.nextToken = response.nextToken;
    } while (maxPagesCount ? commandInput.nextToken && pageCounter < maxPagesCount : commandInput.nextToken);
    return { rows, columnDescriptions, $metadata };
  };

sheilaXu: yes

PR #1052, opened 2023-05-05
fix(scene composer): collapse scene panels on scene load

mumanity:
## Overview

- Sets panels to be automatically collapsed when user opens scene in scene composer (view & edit modes) 
- Consolidates `LeftPanel` / `RightPanel` into single reusable component
- Removes set `400px` of panels to instead be a `max-width: 400px` while maintaining smaller sizes based on size of viewport
- Removes styled component overhead in place of `.scss` files for better performance
- Refactor of `TabbedPanelContainer` to functional component with updated props
- Converted `ScenePanel` fold behavior to be managed by CSS classes

https://user-images.githubusercontent.com/6610619/236550131-643b019c-7485-4a63-8dc2-721d1d11f983.mov

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/layouts/SceneLayout/components/FoldableContainer.scss
```diff
@@ -0,0 +1,26 @@
+// @use '~/@awsui-design-tokens/polaris.scss';
+
+.tm-wrapper-left,
+.tm-wrapper-right {
+  display: flex;
+  align-items: stretch;
+  height: 100%;
+}
+
+.tm-wrapper-left { flex-direction: row;}
```
tracy-french, line 10: Looks like you formatter might be off? Referring to the `;}` instead of `; }`.

packages/scene-composer/src/layouts/SceneLayout/components/FoldableContainer.tsx
```diff
@@ -1,64 +1,41 @@
...
-  const [fold, setFold] = useState(false);
+const FoldableContainer: React.FC<FoldableContainerProps> = ({
+  direction,
+  children,
+  open,
+  setIsOpen,
+}: FoldableContainerProps) => {
+  const [fold, setFold] = useState(!open);
+  const isLeft = direction === 'Left';
+  const isRight = direction === 'Right';
```
tracy-french, line 22: Nit: A small improvement could be to define `isRight` as `!isLeft`, assuming there will never be more than two directions.

packages/scene-composer/src/layouts/SceneLayout/components/ScenePanel.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+    expect(container).toMatchSnapshot();
+  });
+  it('should render the right panel as closed by default', () => {
+    const rightArgs = {
+      direction: Direction.Right,
+      panels: { [1]: <div>TestRightPanel</div> },
+    };
+    const container = render(<ScenePanel {...rightArgs} />);
+    expect(container).toMatchSnapshot();
+  });
```
tracy-french, line 26: Instead of relying on snapshots, is it possible to assert on the presence of the `open` class name?

tracy-french (commented):
Awesome change and great refactor @mumanity! Thank you for your contribution! I'm good with approving your work after the lint warning limit is brought down to passing (looks like it went over by 3 - 680 vs 683) and after you check out if you can revise those tests to not rely on snapshots and instead assert on class name (or maybe visibility, which would be best).

packages/scene-composer/src/layouts/SceneLayout/components/FoldableContainer.scss
mumanity: Great catch! Maybe this is why the linter is failing. :/

mumanity: Ah... hadn't read the overall comment yet.

packages/scene-composer/src/layouts/SceneLayout/components/ScenePanel.spec.tsx
mumanity: Fixed!

tracy-french (approved):
Thank you @mumanity!

packages/scene-composer/src/layouts/SceneLayout/components/FoldableContainer.scss
```diff
@@ -0,0 +1,26 @@
+// @use '~/@awsui-design-tokens/polaris.scss';
```
sheilaXu, line 1: should be `@use '@awsui/design-tokens/index.scss' as awsui;`
then use in code like `awsui.$color-border-divider-default`

packages/scene-composer/src/interfaces/sceneViewer.ts
```diff
@@ -9,6 +11,11 @@ export interface DracoDecoderConfig {
   path?: string;
 }
 
+export interface PanelType {
+  direction: typeof Direction.Left | typeof Direction.Right;
+  panels: Record<string, JSX.Element>;
+}
```
sheilaXu, line 17: this file is only meant for the public interfaces for SceneViewer component. the panelType is an internal component interface that should be in the other types file in the components folder

PR #1116, opened 2023-05-09
feat(dashboard): add grid settings configuration ux

jmbuss:
## Overview
Adds a toggle in actions that will show a dashboard settings modal. This modal can be used to configure the grid settings of the dashboard. The settings toggle is only visible in edit mode as it may affect the widget positions.

grid-settings

## Legal

### DIFF
--- a/packages/dashboard/src/components/actions/settings.tsx
+++ b/packages/dashboard/src/components/actions/settings.tsx
@@ -1,0 +1,61 @@
+import React from 'react';
+
+import { Box, Checkbox, Modal, NonCancelableCustomEvent, SpaceBetween } from '@cloudscape-design/components';
+import { BaseChangeDetail } from '@cloudscape-design/components/input/interfaces';
+
+import LabeledInput from '../util/labeledInput';
+import { useGridSettings } from './useGridSettings';
+
+// Should never return NaN
+const numberFromDetail = (event: NonCancelableCustomEvent<BaseChangeDetail>) => parseInt(event.detail.value) || 0;
+
+export type DashboardSettingsProps = {
+  onClose: () => void;
+  isVisible: boolean;
+};
+
+const DashboardSettings: React.FC<DashboardSettingsProps> = ({ onClose, isVisible }) => {
+  const {
+    rows,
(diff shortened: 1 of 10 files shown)


packages/dashboard/src/components/actions/settings.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+
+import { Box, Checkbox, Modal, NonCancelableCustomEvent, SpaceBetween } from '@cloudscape-design/components';
```
tracy-french, line 3: Nit: You should use default exports on Cloudscape imports to enable treeshaking.

packages/dashboard/src/components/actions/settings.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { BaseChangeDetail } from '@cloudscape-design/components/input/interfaces';
+
+import LabeledInput from '../util/labeledInput';
+import { useGridSettings } from './useGridSettings';
+
+// Should never return NaN
+const numberFromDetail = (event: NonCancelableCustomEvent<BaseChangeDetail>) => parseInt(event.detail.value) || 0;
```
tracy-french, line 10: Nit: I believe linters are starting to mark this is an error and recommending `??` used instead of short-circuiting.

tracy-french (approved):
Great contribution! Your change is super clean and easy to follow!

Just left two nit comments. I imagine named imports are probably already used, so not a blocker, but I would recommend taking a look at making a global improvement to the codebase by switching them to default imports. Might get some perf gains!

packages/dashboard/src/components/actions/settings.tsx
jmbuss: I'll add a story to the backlog to sweep through all of these!

packages/dashboard/src/components/actions/settings.tsx
jmbuss: I don't know if that would work in this scenario since NaN is not null / undefined? But I know it's falsey :D

PR #1207, opened 2023-05-16
feat(3dkg): added 3dkg changes for entity data binding

divya-sea:
## Overview
This PR contains following changes,
* Logic to removed components and properties from data binding.
* Logic On the onSelelctionChangeEvent, take entityId from the data binding and provide it to the additionalDatacomponent variable.
* Unit tests.

## Verifying Changes
* Tested on console.
Screen Shot 2023-05-16 at 11 12 16 AM
* npm run test

## Legal

### DIFF
--- a/packages/scene-composer/src/utils/dataBindingTemplateUtils.ts
+++ b/packages/scene-composer/src/utils/dataBindingTemplateUtils.ts
@@ -1,17 +1,18 @@
-import { cloneDeep } from 'lodash';
+import { cloneDeep, pick } from 'lodash';
 
-import { RootState } from '../store';
+import {
+  DEFAULT_DATA_BINDING_TEMPLATE_COMPONENT_NAME,
+  DEFAULT_DATA_BINDING_TEMPLATE_ENTITY_ID,
+  DataBindingLabelKeys,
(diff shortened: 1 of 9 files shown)


packages/scene-composer/src/utils/dataBindingTemplateUtils.ts
```diff
@@ -96,3 +96,14 @@ export function applyDataBindingTemplate(
...
+
+/**
+ * We extract entityId from the data binding object.
+ * @param dataBinding we send data binding object
+ * @returns
+ */
+export const extractEntityId = (dataBinding: IValueDataBinding): string => {
+  const contextData = dataBinding.dataBindingContext ?? {};
+  const bindingKeys = Object.keys(contextData);
+  return contextData[bindingKeys[0]];
```
haweston, line 108: Since we can't control the order that the keys are in a binding and they can have more than just 1 key (like component, property we might might to do something more like https://github.com/awslabs/iot-app-kit/blob/main/packages/scene-composer/src/utils/dataBindingUtils.ts#L38 so it's more reliable and be reused with tags etc.

mukeshsahay (approved):

PR #1226, opened 2023-05-17
fix(scene composer): modifying auto collapse to only apply to viewer

mumanity:
## Overview
Updating default collapse behavior to only apply to `SceneViewer` left panel.

https://github.com/awslabs/iot-app-kit/assets/6610619/6b5014ed-a786-43eb-a55c-c59bda65fbcf

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/layouts/SceneLayout/components/ScenePanel.tsx
+++ b/packages/scene-composer/src/layouts/SceneLayout/components/ScenePanel.tsx
@@ -8,15 +8,15 @@ import './ScenePanel.scss';
 interface PanelType {
   direction: typeof Direction.Left | typeof Direction.Right;
   panels: Record<string, JSX.Element>;
+  collapse?: boolean;
 }
 
 const ScenePanel = (props: PanelType) => {
-  const { direction, panels } = props;
-  const [isOpen, setIsOpen] = useState<boolean>(false);
-  const ref = useRef<React.ReactFragment>(null);
+  const { direction, collapse = false, panels } = props;
+  const [isOpen, setIsOpen] = useState<boolean>(!collapse);
 
   return (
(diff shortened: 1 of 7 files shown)


packages/scene-composer/src/layouts/SceneLayout/components/ScenePanel.tsx
```diff
@@ -8,15 +8,16 @@ import './ScenePanel.scss';
 interface PanelType {
   direction: typeof Direction.Left | typeof Direction.Right;
   panels: Record<string, JSX.Element>;
+  isViewing?: boolean;
```
mukeshsahay, line 11: No need to pass `isViewing` as a prop. This can be determined using useStore hook.

`const sceneComposerId = useSceneComposerId();`
` const isViewing = useStore(sceneComposerId)((state) => state.isViewing());`

mukeshsahay (changes requested):

packages/scene-composer/src/layouts/SceneLayout/components/ScenePanel.tsx
mumanity: It's a good suggestion, however, the value for `isViewing` coming from `useStore` gets modified on multiple re-renders & doesn't behave as you would expect. The prop being passed works as intended & does not create any additional dependency that didn't already exist. This shouldn't be a blocker.

mumanity: Per sync with @mukeshsahay we've modified the prop from `isViewing` to `collapse` to improve readability & maintain the flexibility of the component. :D

mukeshsahay (approved):

PR #1239, opened 2023-05-18
fix(composer): click events for tag and overlay

sheilaXu:
## Overview
Fix the issues:
- click on Tag in 3D scene doesn't select the Tag
- mouse down somewhere, move the camera, release mouse over the overlay or annotation, the camera continues tracking with the mouse until clicking again

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
@@ -178,8 +178,6 @@ export function AsyncLoadedAnchorWidget({
               ],
             });
           }
-
-          event.stopPropagation();
         }
       }
     },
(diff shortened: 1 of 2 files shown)


divya-sea (commented):
Why did we change from onPointerUp to onClickContainer? How did we determine onClickContainer is the right method use? Anywhere else do we need this change?

sheilaXu:
> Why did we change from onPointerUp to onClickContainer? How did we determine onClickContainer is the right method use? Anywhere else do we need this change?

onPointerUp/Down is used by orbitControl to move camera which is causing conflicts with the events on overlay. the two issues mentioned have been verified to be fixed

mumanity (approved):

PR #1244, opened 2023-05-19
feat(scene composer): entity data binding UI changes and unit tests

divya-sea:
## Overview
This PR contains following changes,
UI changes for entity data binding.
Rearranged the imports.
Unit tests

## Verifying Changes
* npm run test
* added unit tests

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/StateManager.tsx
+++ b/packages/scene-composer/src/components/StateManager.tsx
@@ -40,7 +40,7 @@ import {
   useViewOptionState,
 } from '../store';
 import { getCameraSettings } from '../utils/cameraUtils';
-import { applyDataBindingTemplate, extractEntityId } from '../utils/dataBindingTemplateUtils';
+import { applyDataBindingTemplate } from '../utils/dataBindingTemplateUtils';
 import { combineTimeSeriesData, convertDataStreamsToDataInput } from '../utils/dataStreamUtils';
 import { findComponentByType } from '../utils/nodeUtils';
 import sceneDocumentSnapshotCreator from '../utils/sceneDocumentSnapshotCreator';
@@ -180,9 +180,9 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
       // will always have only one entity data.
       if (entityBindingComponent) {
         additionalComponentData?.push({
-          dataBindingContext: !entityBindingComponent?.valueDataBindings?.[0].valueDataBinding?.dataBindingContext
+          dataBindingContext: !entityBindingComponent?.valueDataBinding?.dataBindingContext
             ? undefined
-            : extractEntityId(entityBindingComponent?.valueDataBindings?.[0].valueDataBinding),
+            : entityBindingComponent?.valueDataBinding?.dataBindingContext,
         });
       }
       onSelectionChanged({
--- a/packages/scene-composer/src/components/panels/AddComponentMenu.tsx
+++ b/packages/scene-composer/src/components/panels/AddComponentMenu.tsx
@@ -1,17 +1,17 @@
-import React, { useCallback, useContext, useMemo } from 'react';
-import { defineMessages, MessageDescriptor, useIntl } from 'react-intl';
-import * as THREE from 'three';
 import { IconProps } from '@awsui/components-react';
+import React, { useCallback, useContext, useMemo } from 'react';
+import { MessageDescriptor, defineMessages, useIntl } from 'react-intl';
+import * as THREE from 'three';
 
-import { ToolbarItemOptionRaw, ToolbarItemOptions } from '../toolbars/common/types';
-import { sceneComposerIdContext } from '../../common/sceneComposerIdContext';
-import { IDataOverlayComponentInternal, useStore } from '../../store';
-import { findComponentByType } from '../../utils/nodeUtils';
-import { COMPOSER_FEATURES, KnownComponentType } from '../../interfaces';
-import { ToolbarItem } from '../toolbars/common/ToolbarItem';
 import { getGlobalSettings } from '../../common/GlobalSettings';
+import { sceneComposerIdContext } from '../../common/sceneComposerIdContext';
+import { COMPOSER_FEATURES, KnownComponentType } from '../../interfaces';
 import { Component } from '../../models/SceneModels';
+import { IDataOverlayComponentInternal, useStore } from '../../store';
 import { IDataBindingComponentInternal } from '../../store/internalInterfaces';
+import { findComponentByType } from '../../utils/nodeUtils';
+import { ToolbarItem } from '../toolbars/common/ToolbarItem';
+import { ToolbarItemOptionRaw, ToolbarItemOptions } from '../toolbars/common/types';
 
 interface AddComponentMenuProps {
   onSelect?: (selectedObject: ObjectTypes) => void;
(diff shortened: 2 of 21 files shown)


packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -157,7 +167,16 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
           },
         ];
       }
-
+      // Add entityID info part of additional component data
+      // We assumed IDataBindingMap will have only one mapping as data binding
+      // will always have only one entity data.
+      if (entityBindingComponent) {
+        additionalComponentData?.push({
```
sheilaXu, line 174: so if it's an empty node with just the entity binding component, we will not return entity binding data? I don't think this is expected behavior, and we should always return entity binding data when the selected node has this component

packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -157,7 +167,16 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
...
+          dataBindingContext: !entityBindingComponent?.valueDataBindings?.[0].valueDataBinding?.dataBindingContext
+            ? undefined
+            : extractEntityId(entityBindingComponent?.valueDataBindings?.[0].valueDataBinding),
```
sheilaXu, line 177: I think data returned here should be consistent with the actual entity binding type - if the entity binding data format support multiple, then we should return array here even though the array may only have 1 element. or we make the binding data format be single binding only, then we just return that. 

otherwise, if in the future we can having multiple entity binding for some reason, we will need to change the return value type here

packages/scene-composer/src/components/panels/AddComponentMenu.tsx
```diff
@@ -76,13 +76,15 @@ export const AddComponentMenu: React.FC<AddComponentMenuProps> = ({ onSelect })
...
       : [];
-    const addDataBindingItem = dataBindingComponentEnabled
-      ? [
-          {
-            uuid: ObjectTypes.DataBinding,
-          },
-        ]
-      : [];
+    const addDataBindingItem =
+      dataBindingComponentEnabled && !findComponentByType(selectedSceneNode, KnownComponentType.DataBinding)
```
sheilaXu, line 80: remove `&& !findComponentByType(selectedSceneNode, KnownComponentType.DataBinding)`?

packages/scene-composer/src/components/panels/ComponentEditMenu.tsx
```diff
@@ -21,6 +21,8 @@ interface ComponentEditMenuProps {
 enum ObjectTypes {
   EditComponent = 'edit-component',
   AddDataBinding = 'add-data-binding',
+  AddEntityBinding = 'add-entity-binding',
```
sheilaXu, line 24: is it used anywhere?

packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -145,7 +152,10 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
       const componentTypes = node?.components.map((component) => component.type) ?? [];
 
       const tagComponent = findComponentByType(node, KnownComponentType.Tag) as IAnchorComponentInternal;
-
+      const entityBindingComponent = findComponentByType(
```
sheilaXu, line 155: `componentTypes` and `additionalComponentData` is supposed to match, meaning if the first element in `componentTypes` is `EntityBinding`, then the first element in `additionalComponentData` should also be the additional data for entity binding. so you should update the logic below to enforce that instead of always adding entity binding data later

packages/scene-composer/src/components/panels/AddComponentMenu.tsx
```diff
@@ -31,12 +31,12 @@ type AddComponentMenuItem = ToolbarItemOptions & {
 const labelStrings: { [key in ObjectTypes]: MessageDescriptor } = defineMessages({
   [ObjectTypes.Component]: { defaultMessage: 'Add component', description: 'Menu Item label' },
   [ObjectTypes.Overlay]: { defaultMessage: 'Overlay', description: 'Menu Item label' },
-  [ObjectTypes.DataBinding]: { defaultMessage: 'Data binding', description: 'Menu Item label' },
+  [ObjectTypes.DataBinding]: { defaultMessage: 'Add entity binding', description: 'Menu Item label' },
```
sheilaXu, line 34: `ObjectTypes.DataBinding` should be renamed to `ObjectTypes.EntityBinding` to make it less confusing

packages/scene-composer/src/components/panels/ComponentEditMenu.tsx
```diff
@@ -32,13 +34,16 @@ type ComponentEditMenuItem = ToolbarItemOptions & {
...
   [ObjectTypes.EditComponent]: { defaultMessage: 'Edit component', description: 'Menu Item label' },
   [ObjectTypes.AddDataBinding]: { defaultMessage: 'Add data binding', description: 'Menu Item label' },
+  [ObjectTypes.AddEntityBinding]: { defaultMessage: 'Add entity binding', description: 'Menu Item label' },
+  [ObjectTypes.RemoveEntityBinding]: { defaultMessage: 'Remove entity binding', description: 'Menu Item label' },
   [ObjectTypes.RemoveAllDataBinding]: { defaultMessage: 'Remove all data binding', description: 'Menu Item label' },
   [ObjectTypes.RemoveOverlay]: { defaultMessage: 'Remove overlay', description: 'Menu Item label' },
 });
 
 const textStrings = defineMessages({
-  [ObjectTypes.AddDataBinding]: { defaultMessage: 'Add data binding', description: 'Menu Item' },
```
sheilaXu, line 40: this shouldn't be changed since it's for adding data binding to overlay which is not entity binding

packages/scene-composer/src/components/panels/ComponentEditMenu.tsx
```diff
@@ -21,6 +21,8 @@ interface ComponentEditMenuProps {
 enum ObjectTypes {
   EditComponent = 'edit-component',
   AddDataBinding = 'add-data-binding',
+  AddEntityBinding = 'add-entity-binding',
+  RemoveEntityBinding = 'remove-entity-binding',
   RemoveAllDataBinding = 'remove-all-data-binding',
```
sheilaXu, line 26: so `RemoveAllDataBinding` is not used anywhere and should be removed

packages/scene-composer/src/components/panels/ComponentEditMenu.tsx
divya-sea: Yes. This is used in the addComponentMenu.

packages/scene-composer/src/components/panels/SceneNodeInspectorPanel.tsx
```diff
@@ -57,7 +57,7 @@ export const SceneNodeInspectorPanel: React.FC = () => {
       description: 'Expandable Section title',
     },
     [KnownComponentType.DataBinding]: {
```
sheilaXu, line 59: we should rename `KnownComponentType.DataBinding` to `KnownComponentType.EntityBinding` to not get confused internally

packages/scene-composer/src/components/StateManager.tsx
divya-sea: We don't support multiple entity binding data. one scene node/ 3d objects can have only one entity data binding. I think it's not necessary to return array objects if this has only one binding.

packages/scene-composer/src/components/panels/ComponentEditMenu.tsx
divya-sea: What's the difference between entity binding vs data binding? are these two are different concepts?

packages/scene-composer/src/components/StateManager.tsx
divya-sea: I did not follow your comment. Could you give me an example?

haweston (approved):
We need to capture exactly what's being delayed til next PR for, but for the items we're doing in this PR it looks complete.

PR #1284, opened 2023-05-23
fix: graph css not distributed correctly

TheEvilDev:
## Overview
In previous builds, the storybook site could run because it was loading graph stylesheets from src. But when trying to consume this component in an external system for testing, the dist folder didn't contain the stylesheets.

I've updated the build scripts.

NOTE: I also identified an issue with our build where the `react-components` package was incorrectly referencing the styles from the `components` package, but instead of node modules it was copying it directly on the file system. This has also been corrected.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


examples/react-app/package.json
```diff
@@ -12,6 +12,7 @@
     "@iot-app-kit/source-iotsitewise": "*",
     "@iot-app-kit/source-iottwinmaker": "*",
     "@iot-app-kit/table": "*",
+    "cytoscape": "^3.25.0",
```
MO-Elmu, line 15: Why do we need to install cytoscape in the react-app ?

examples/react-app/src/index.tsx
```diff
@@ -13,20 +12,17 @@ import {
   applyMode,
 } from '@cloudscape-design/global-styles';
 
+import "@cloudscape-design/global-styles/index.css"
+import '@iot-app-kit/components/styles.css';
+
```
MO-Elmu, line 17: Same here, why we are importing cloudscape and styles in an application this should bundled and a customer who uses our component doesn't have to import these explicitly in their APP.

examples/react-app/package.json
MO-Elmu: I am sorry but this is not acceptable at all and we agreed before we don't want the customer to know or rely on our choice of graph viz library. This will really tie us to this library.

examples/react-app/src/index.tsx
MO-Elmu: But this means every customer who wants to use our Knowledge Graph component in their App they have to import cloudscape global styles. This is a very strange limitation and looks very bad. if this is the case then I will move knowledge Graph component from reac-components to its own component like scene composer.

PR #1322, opened 2023-05-24
fix(dashboard): change min widget sizing from 2 to 1

jmbuss:
## Overview
Refactor min widget height and width from 2 to 1 on resize action. Bugfix for trimRectPosition to ensure that the rectangle can never be rounded to 0 and create a widget with no height or width.

## Legal


tracy-french (approved):
Thank you for fixing this @jmbuss! :)

corteggiano (approved):

PR #1393, opened 2023-05-30
fix: graph css not distributed correctly

MO-Elmu:
## Overview

This PR moves all graph code folder under a new folder called knowledge-graph in react-components. And also add a parent component KnowledgeGraphPanel which represent the final shape of the knowledge graph component we are building. Also add react context for state management and text search backed by knowledge graph queries and relationship explore logic. UI wise it adds text search and Search, Clear and Explore buttons. After testing with real APP i found some issues with CSS loading and invalid error. CSS file without a loader doesn't get bundled and we need to copy them using commands in package.json since we are using tsconfig here no webpack ...etc. These CSS issues will be investigated and fixed in the next PR. We need this skeleton PR to unblock other team members who are waiting for this.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


.github/workflows/ui-test-for-commit.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           key: ${{ runner.os }}-playwright-${{ env.PLAYWRIGHT_VERSION }}
       - name: Install
         run:
-          npm ci --prefer-offline --no-audit --no-fund
```
MO-Elmu, line 25: The whole idea of using npm ci in the workflow  is not to use npm install. coz you need clean install to detect any mismatch or incompatible versions as it install directly from the lock.json file and compare with package.json. adding npm run bootstrap is basically using npm install (here using both ci and install) and this will update the lock file and this is not desirable in a workflow automated environment. I have a feeling this is a cache issue why not just remove the flag --prefer-offline and try. Since this flag is to speedup the ci command using cache.

mumanity (approved):

PR #1408, opened 2023-05-31
chore(dashboard): updating the right click context menu to cloudscape

mnischay:
## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/dashboard/src/components/contextMenu/menu.tsx
```diff
@@ -49,16 +55,14 @@ const Menu: React.FC<MenuProps> = ({ position, clickOutside, children }) => {
         style={{
           ...styles.popper,
           zIndex: position.z,
+          border: `${spaceScaledXxxs} solid ${colorBorderControlDefault}`,
+          minWidth: '192px', // 128+64
```
jmbuss, line 59: Is this from the previous implementation? What's the relevance of 128 + 64?

packages/dashboard/src/components/contextMenu/menu.css
```diff
@@ -4,9 +4,7 @@
...
 }
 
-.iot-context-menu {
-  min-width: var(--size-context-menu-min-width);
-  border-radius: var(--radius-context-menu);
-  background-color: var(--colors-dark-grey);
-  z-index: var(--stack-order-context-menu);
+.iot-context-list-container {
+  padding: 0;
+  margin: -2px;
```
jmbuss, line 9: what is this for? something around the border I'm guessing?

packages/dashboard/src/components/contextMenu/menu.tsx
mnischay: instead of 200px, just having a width which is a multiple of 8...

packages/dashboard/src/components/contextMenu/menu.css
mnischay: yes.. else the on hover border will move the outer border and the UX looks shifty

packages/dashboard/src/components/contextMenu/menu.tsx
jmbuss: can you add a note for that incase we refactor in the future? Then we know it wasn't specifically required to make something else work and can safely be increased / decreased?

packages/dashboard/src/components/contextMenu/menu.css
jmbuss: Does it match the same values as - `border: `${spaceScaledXxxs} solid ${colorBorderControlDefault}`? if so, can we make it match the same variable? Probably would have to move it into the style tag to do so. I don't think we'd catch that shifting behavior in any test so I could see us refactoring and causing that bug

packages/dashboard/src/components/contextMenu/menu.tsx
mnischay: updated

packages/dashboard/src/components/contextMenu/menu.css
mnischay: updated

jmbuss (approved):

PR #1464, opened 2023-06-06
feat(scene composer): fix autosuggest value issues in ValueDataBinding…

divya-sea:
…Builder

## Overview
* Fix auto suggested value.
* Fix scene node reference value.
### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/components/panels/scene-components/common/ValueDataBindingBuilder.tsx
```diff
@@ -128,6 +133,7 @@ export const ValueDataBindingBuilder: React.FC<IValueDataBindingBuilderProps> =
                       dataBindingConfig,
                     );
                     if (allowPartialBinding) {
+                      console.log('is this coming here', allowPartialBinding);
```
haweston, line 136: Do we need this log statement still?

packages/scene-composer/src/components/panels/scene-components/common/ValueDataBindingBuilder.tsx
```diff
@@ -91,8 +91,13 @@ export const ValueDataBindingBuilder: React.FC<IValueDataBindingBuilderProps> =
...
     setBuilderState(filterBuilderState(state));
-    setAutoSuggestValue(state.selectedOptions[ENTITY_ID_INDEX]?.value || autoSuggestValue);
-  }, [componentRef, binding, valueDataBindingProvider, dataBindingConfig]);
+  }, [binding, valueDataBindingProvider, dataBindingConfig]);
+
+  useEffect(() => {
+    const state = valueDataBindingStore.setBinding(componentRef, binding, dataBindingConfig);
+    setBuilderState(filterBuilderState(state));
+    setAutoSuggestValue(state.selectedOptions[ENTITY_ID_INDEX]?.value || '');
+  }, [componentRef]);
```
haweston, line 100: So the state is still dependent on binding and databindingConfig in this new useEffect but their not in the dependency array.  What's the advantage of splitting this into two use effects? Could you get the same effect just by updating the setAutoSuggestValue parameter to (value || '') instead?  What's the benefit of changing from autoSuggestValue last value vs ''?

PR #1465, opened 2023-06-06
feat(Knowledge Graph): adding clear and render graph results events

MO-Elmu:
## Overview
- Emit and event that other widgets (like the scene viewer) can listen to so that they know to clear their graph based visual items (like 3D highlights)
- Update graph result events to support emitting multiple entities.
This will allow a react developer to highlight multiple items in a linked scene such as 'all roof top units'.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


haweston (approved):
onClearGraph might not need nodes & edges as parameters as I think for now we'll just use it as a clear all, but it will be more flexible down the road with them and potentially reduce some business logic store.

PR #1571, opened 2023-06-12
feat: feature flag

corteggiano:
## Overview
create the appKit context component, which will hold feature flags and any other future configurable aspects such as translations, theme, etc.

2 ways to use this new config:

1. wrap all your widgets in a provider. example: `dashboard/src/components/dashboard/index.tsx`. For now the provider accepts a `customFeatureConfig` which toggles feature flags
2. do nothing and use default values chosen by us :) 

also created the following hooks: 
`useHasFeatureFlag` (featureName) => boolean
`useAppKitConfig` () => config

example of config we could have in the future:
```
{
  featureConfig: { 
    HAS_FEATURE_A: false, 
    HAS_FEATURE_B: true    
  },
  themeColors: {
    background: #...,
    mainText: #...
  }
};
```

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/eChartsConstants.ts
+++ b/packages/react-components/src/components/chart/eChartsConstants.ts
@@ -1,6 +1,5 @@
 import { format } from 'd3-format';
 import type {
-  EChartsOption,
   XAXisComponentOption,
   YAXisComponentOption,
   LegendComponentOption,
@@ -20,16 +19,19 @@ export const DEFAULT_Y_AXIS: YAXisComponentOption = {
   axisLabel: { formatter: format('.2s') },
   position: 'left',
 };
+
+// if you change this, please update the width calculation
+export const DEFAULT_MARGIN = 50;
+
 export const DEFAULT_LEGEND: LegendComponentOption = {
   show: true,
   type: 'scroll',
   orient: 'horizontal',
   left: 'left',
   bottom: '0',
+  padding: [5, DEFAULT_MARGIN / 2],
 };
 
-// if you change this, please update the width calculation
-export const DEFAULT_MARGIN = 50;
 export const DEFAULT_GRID = {
   left: DEFAULT_MARGIN,
   top: DEFAULT_MARGIN,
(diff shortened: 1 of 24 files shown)


packages/dashboard/src/components/dashboard/index.tsx
```diff
@@ -40,17 +41,19 @@ const Dashboard: React.FC<DashboardProperties> = ({
...
-              options={{
-                enableMouseEvents: true,
-                enableKeyboardEvents: true,
-              }}
-            >
-              <InternalDashboard onSave={onSave} editable={true} />
-            </DndProvider>
-          </Provider>
+          <AppKitConfig>
+            <Provider store={configureDashboardStore({ ...toDashboardState(dashboardConfiguration), readOnly })}>
```
mnischay, line 45: curious: why do use context instead of using some state management ?

packages/react-components/src/hooks/useAppKitConfig/useAppKitConfig.ts
```diff
@@ -0,0 +1,7 @@
+import { useContext } from 'react';
+import { AppKitContext, IAppKitConfig } from '../../components/iot-app-kit-config';
+import { DEFAULT_APP_KIT_CONFIG } from '../../components/iot-app-kit-config/defaultValues';
+
+export const useAppKitConfig = (): IAppKitConfig => {
```
mnischay, line 5: super nit: 

`export const useAppKitConfig = (): IAppKitConfig => useContext(AppKitContext) ?? DEFAULT_APP_KIT_CONFIG `;

packages/dashboard/src/components/dashboard/index.tsx
corteggiano: also considered state management but most of the app is context based so i mostly didnt want to throw off the current way of doing things.

i dont know the History of context vs state management but maybe someone that has been here longer has more info on that !

packages/react-components/src/components/iot-app-kit-config/types.ts
```diff
@@ -0,0 +1,5 @@
+export enum FEATURE_NAME {
```
mnischay, line 1: just for future ref: const enum are more performant than enums...

mnischay (approved):

packages/react-components/src/components/iot-app-kit-config/types.ts
```diff
@@ -0,0 +1,5 @@
+export enum FEATURE_NAMES {
```
mnischay, line 1: super nit: const enum is a bit better performance wise.

packages/react-components/src/components/iot-app-kit-config/types.ts
```diff
@@ -0,0 +1,5 @@
...
+  HAS_E_CHARTS = 'HAS_E_CHARTS',
```
mnischay, line 2: super nit: may be ENBALE_E_CHARTS ?

packages/react-components/src/hooks/useHasFeatureFlag/useHasFeatureFlag.ts
```diff
@@ -0,0 +1,13 @@
+import { FEATURE_NAMES } from '../../components/iot-app-kit-config/types';
+import { useAppKitConfig } from '../useAppKitConfig';
+
+export const useHasFeatureFlag = (flagName: string): boolean => {
+  const config = useAppKitConfig();
+
+  if (flagName in FEATURE_NAMES) {
```
mnischay, line 7: super nit  : FEATURE_FLAGS ?

packages/react-components/src/components/iot-app-kit-config/defaultValues.ts
```diff
@@ -0,0 +1,9 @@
+import { FEATURE_FLAGS } from './types';
```
jmbuss, line 1: should the feature flags and the default value be defined in dashboard? I think the config can just be defined here as an object with boolean values?

tracy-french:
What are the next steps for getting this work in? :)

mnischay:
> What are the next steps for getting this work in? :)

this needs to be rebased to the main...

packages/react-components/src/components/chart/eChartsConstants.ts
```diff
@@ -54,12 +53,3 @@ export const DEFAULT_DATA_ZOOM: DataZoomComponentOption = {
   moveOnMouseMove: true,
   moveOnMouseWheel: false,
 };
-
-export const DEFAULT_ECHART_OPTIONS: EChartsOption = {
```
corteggiano, line 58: not used anymore

jmbuss (approved):

PR #1606, opened 2023-06-15
fix(knowledge-graph): fix node to node selection and layout adjustment

MO-Elmu:
## Overview
This PR addresses 
1- node source to node target select bug.
2- Fix grid layout Padding and use grid layout for the nodes only case as it provides better visuals.
3- Modify CSS to provide min width and height and expand to 100% of the parent container.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
```diff
@@ -39,6 +40,38 @@ export const KnowledgeGraphContainer: React.FC<KnowledgeGraphInterface> = ({
...
+    ...defaultLayoutOptions,
+    name: 'breadthfirst',
+  };
+  const coselayout = {
+    ...defaultLayoutOptions,
+    name: 'grid',
+    fit: true,
+    padding: 50,
+  };
+
```
MO-Elmu, line 74: working on moving this to a separate layout file for better and clean code, as of now if i do so and import the layout it doesn't work and layout doesn't get applied unless its in the same file(very strange).I am working on a fix but I need this to ship before private beta.

packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
```diff
@@ -72,20 +105,26 @@ export const KnowledgeGraphContainer: React.FC<KnowledgeGraphInterface> = ({
...
+      console.log('selectedGraphNodeEntityId Inside = ', selectedGraphNodeEntityId);
     },
     [onEntitySelected, setSelectedGraphNodeEntityId]
   );
 
   const clickRelationshipHandler = useCallback(
-    ({ target }: EventObjectEdge) => {
-      const data = target.data() as EdgeData;
+    async ({ target }: EventObjectEdge) => {
+      const data = (await target.data()) as EdgeData;
```
MO-Elmu, line 127: Using async await here is a temporary hack to address a corner case, see the comment and the stackoverflow link in the code for more explanation.

PR #1612, opened 2023-06-15
feat(3D knowledge Graph): add support for selecting entity binding to…

haweston:
… trigger camera movement

## Overview
Scene Viewer parameter selectedDataBinding for initializing camera position (or maintaining it without using new set Hooks) only reacts to Tags. This change adds support for Entity Binding components as well which requires support for bindings that might not have a component name and instead just an Entity ID.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/SceneViewer.tsx
```diff
@@ -43,7 +43,10 @@ export const SceneViewer: React.FC<SceneViewerProps> = ({ sceneComposerId, confi
       return;
     }
 
-    const nodeRefs = composerApis.findSceneNodeRefBy(props.selectedDataBinding || '', [KnownComponentType.Tag]);
+    const nodeRefs = composerApis.findSceneNodeRefBy(props.selectedDataBinding || '', [
+      KnownComponentType.Tag,
+      KnownComponentType.DataBinding,
+    ]);
```
sheilaXu, line 49: this will be more likely to return multiple refs now. maybe you could check if the `selectedDataBinding` is an entity binding (e.g. only have `entityId`), if so, filter by `KnownComponentType.DataBinding`. otherwise filter by KnownComponentType.Tag.

sheilaXu: also, we should rename `KnownComponentType.DataBinding` to `KnownComponentType.EntityBinding` before release

sheilaXu:
your PR title and commit message are too long to be displayed fully, so if it's merged in this way, the auto generated CHANGELOG will also contain this incomplete commit message. you should update them so they can fit

packages/scene-composer/src/SceneViewer.tsx
haweston: Add  test for if componentName is present and then added unit test to cover new option.  Will handle the data type renaming in future PR.

sheilaXu (approved):

PR #1642, opened 2023-06-20
fix(KG Component): ux review changes

MO-Elmu:
## Overview
This PR addresses changes requested by UX team during the UX review,
1- when there is only one node result it looks very big and fits the whole screen, this is changed with fixed zooming and different layout.
2- The search input field should stretch up to 75% of the canvas width.
3- Canvas only top a bottom border lines is better.
4- should show magnifying glass in the search input field.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
+++ b/packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
@@ -2,7 +2,7 @@
 import React, { HTMLAttributes, useEffect, useCallback, useMemo, useState, useRef } from 'react';
 import { IntlProvider, FormattedMessage } from 'react-intl';
 import type { Core, EventObjectNode, EventObjectEdge } from 'cytoscape';
-import { Button, Container, Header, Input, SpaceBetween } from '@cloudscape-design/components';
+import { Button, Container, Header, Input, SpaceBetween, Grid } from '@cloudscape-design/components';
 import { TwinMakerKGQueryDataModule } from '@iot-app-kit/source-iottwinmaker';
 import GraphView from './graph/graph-view';
 import Toolbar from './graph/graph-toolbar';
@@ -24,7 +24,7 @@ export interface KnowledgeGraphInterface extends HTMLAttributes<HTMLDivElement> 
   onRelationshipUnSelected?: (e: EdgeData) => void;
   onGraphResultChange?: (nodes: NodeData[], edges?: EdgeData[]) => void;
   onClearGraph?: (nodes: NodeData[], edges?: EdgeData[]) => void;
-  queryData?: IQueryData;
+  queryData?: IQueryData | null;
 }
 export const ZOOM_INTERVAL = 0.1;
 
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
```diff
@@ -172,7 +177,7 @@ export const KnowledgeGraphContainer: React.FC<KnowledgeGraphInterface> = ({
         ? onGraphResultChange([...nodeData.values()], [...edgeData.values()])
         : onGraphResultChange([...nodeData.values()]);
     }
-    cy.current?.fit();
+    if (nodeData.size > 1) cy.current?.fit();
```
mumanity, line 180: Wouldn't this be `>==` or is being equal to 1 something you would want to ignore?

packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
```diff
@@ -208,31 +213,32 @@ export const KnowledgeGraphContainer: React.FC<KnowledgeGraphInterface> = ({
...
-              <FormattedMessage
-                id='KnowledgeGraphPanel.button.search'
-                defaultMessage='Search'
-                description='Search button text'
-              />
-            </Button>
-          </SpaceBetween>
-          {/* inline styling here for testing only this will be fixed in the next PR */}
+        <SpaceBetween direction='vertical' size='xl'>
+          <div style={{ minWidth: '500px' }}>
```
mumanity, line 217: Why `500px`? This would create issues in the future with responsive support. Suggest using a min of `400px` if possible.

packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
```diff
@@ -208,31 +213,32 @@ export const KnowledgeGraphContainer: React.FC<KnowledgeGraphInterface> = ({
...
+            </Grid>
+          </div>
           <div ref={containerRef} className={`${STYLE_PREFIX} ${className || ''}`.trim()} {...props}>
             <GraphView
               className={`${STYLE_PREFIX}-canvas`}
               ref={cy}
               stylesheet={stylesheet}
               elements={getElementsDefinition([...nodeData.values()], [...edgeData.values()])}
-              layout={edgeData.size > 0 ? breadthFirstlayout : coselayout}
+              layout={edgeData.size > 0 ? breadthFirstlayout : nodeData.size > 1 ? gridlayout : presetlayout}
```
mumanity, line 241: Same question here about `>` vs `>==`.

haweston (approved):

packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
MO-Elmu: yup equal to 1 we want to ignore, the UX felt when we have a one node result it fits the whole screen and it looks big and ugly, so instead we don't want it to fit the whole canvas.

packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
mumanity: Smaller phones are usually max-width around `400px`

MO-Elmu: This is to align with the canvas minWidth which is 500px (after experimentation this is the best minWidth for the graph canvas).

packages/react-components/src/components/knowledge-graph/KnowledgeGraphPanel.tsx
MO-Elmu: yup we want to address one corner case which is when the number of nodes =1

PR #1665, opened 2023-06-26
feat(dashboard): initial implementation of static tooltips(scooters) …

mnischay:
## Overview
this is an initial implementation of scooters
1. user can double click to enter the add scooter mode
2. user can click on any node to add a scooter
3. scooter moves along with that node when there is new data, 

things to do
1. the draggability is not available because React dnd seems to take over the chart
2. the default behaviour of the tooltip (show on hover) seems to be broken, but still shows on click

## Verifying Changes
refer to the screen shot and screen capture

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/a636d1f0-674a-497c-9300-3f6cd355486b

Screenshot 2023-06-27 at 11 44 50 AM

### DIFF
--- a/packages/react-components/src/components/BaseECharts/BaseECharts.tsx
+++ b/packages/react-components/src/components/BaseECharts/BaseECharts.tsx
@@ -1,36 +1,12 @@
-import React, { useMemo } from 'react';
-import {
-  StyleSettingsMap,
-  Threshold,
-  TimeSeriesDataQuery,
-  Viewport,
-  ThresholdSettings,
-  DataPoint,
-} from '@iot-app-kit/core';
+import React, { useEffect, useMemo, useRef } from 'react';
+import { StyleSettingsMap, Threshold, TimeSeriesDataQuery, Viewport, ThresholdSettings } from '@iot-app-kit/core';
 import { DataStream } from '@iot-app-kit/charts-core';
-import type { SeriesOption } from 'echarts';
-
+import type { EChartsType } from 'echarts';
 import { useTimeSeriesData } from '../../hooks/useTimeSeriesData';
 import { AxisSettings } from '../../common/chartTypes';
-import { DEFAULT_ECHART_OPTIONS, DEFAULT_X_AXIS, DEFAULT_Y_AXIS } from './eChartsConstants';
-import { StreamType } from '../../common/constants';
-import { useECharts } from '../../hooks/useECharts';
-
-// specifically converts dataStreams to series for bar, scatter, line charts
-const dataStreamsToEChartSeries = (dataStreams: DataStream[], chartType: string): SeriesOption[] => {
-  return dataStreams
-    .filter(({ streamType }) => streamType !== StreamType.ALARM) // need to filter out alarm streams. not supported in bar, scatter, line
-    .map(
-      (stream) =>
-        ({
-          name: stream.name ?? '',
-          data: stream.data.map((point: DataPoint) => [point.x, point.y]),
-          type: chartType,
-          itemStyle: { color: stream.color } ?? {},
-          lineStyle: { color: stream.color } ?? {},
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/BaseECharts/BaseECharts.tsx
```diff
@@ -66,31 +43,105 @@ export const BaseECharts = ({
...
+  useEffect(() => {
+    let chart: EChartsType | undefined;
+    if (chartRef.current !== null) {
+      chart = getInstanceByDom(chartRef.current);
+      chart?.on('click', (e) => {
+        if (tooltipAddMode.current) {
+          const length = (optionRef.current.graphic as GraphicInternal[])?.length ?? 0;
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          optionRef.current?.graphic?.push(getToolTipGraphic(e, length, chart, size));
```
jmbuss, line 117: what if we maintain state in the component for scooters? like a scooters useState list. we can convert this into a graphics list with all the other props we convert and call setOption in 1 place

jmbuss: we can also make sure the positions are correct in the same conversion step.

jmbuss: Also, I think it would be helpful to have a data-structure which represents the scooter, and then we convert that into the data-structure the chart requires to visualize it. I think that will make the code easier to reason about in the future in the case where we add another feature which also utilizes graphics

packages/react-components/src/components/BaseECharts/BaseECharts.tsx
```diff
@@ -66,31 +43,105 @@ export const BaseECharts = ({
     },
     styles,
   });
-
+  const optionRef = useRef({
+    ...DEFAULT_ECHART_OPTIONS,
+  });
+  const tooltipAddMode = useRef(false);
```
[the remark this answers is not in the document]
mnischay: We do not want to re-render, with echarts its all about set the right options

packages/react-components/src/components/BaseECharts/eChartsUtlis.ts
```diff
@@ -0,0 +1,107 @@
...
+    const newX = widthPerTime * index + Number(DEFAULT_MARGIN);
+
+    g.children[0].shape.x1 = newX;
+    g.children[0].shape.x2 = newX;
+    g.children[1].shape.x = newX - 30;
+    return g;
+  });
+};
+
+export const getToolTipGraphic = (
```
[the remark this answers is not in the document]
mnischay: hmmmm, this would be part of the on-click event handler, and move would be when the user moves or when the data changes... we could have the same function for both for both initializing and moving .. for now i have kept them separate given move can be simple.... i am not entirely sure about keeping them in the same context..

packages/react-components/src/components/BaseECharts/BaseECharts.tsx
```diff
@@ -66,31 +43,105 @@ export const BaseECharts = ({
     },
     styles,
   });
-
+  const optionRef = useRef({
```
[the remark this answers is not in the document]
mnischay: nope, i kept getting the old option value

packages/react-components/src/components/BaseECharts/BaseECharts.tsx
mnischay: yes, this is just a basic implementation, a lot of specific implementation details will change... wanted to have something in place so that we are working off of one file / base pattern, this is not the final solution

mnischay:
https://github.com/awslabs/iot-app-kit/pull/1676/files

PR #1671, opened 2023-06-27
feat(react-components): add chart component with updated api

jmbuss:
## Overview
Update Base chart API

Co-mingled charts example
image

Step charts example
image

significant digits example (2)
image

## Legal


mnischay (commented):
may be upload a screen capture or screen shot for reference ?

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useVisualizedDataStreams } from './useVisualizedDataStreams';
+import { convertOptions } from './converters/convertOptions';
+
+/**
+ * Base chart to display Line, Scatter, and Bar charts.
+ */
+const Chart = ({ viewport, queries, size, ...options }: ChartOptions) => {
+  const { hasError, isLoading, dataStreams } = useVisualizedDataStreams(queries, viewport);
+
+  const option = convertOptions(
```
mnischay, line 13: we need to keep an "global object" per chart for option.. https://echarts.apache.org/en/api.html#echartsInstance.getOption.. we cannot lose the object given we will it to add or delete elements

mnischay (approved):

PR #1676, opened 2023-06-30
feat(react-component): adding the initial implementation of the trend …

mnischay:
## Overview
1. Adding the initial implementation of the trend cursors
2. adding hot key support , using cmd as the button for now

## Verifying Changes
refer the screen capture 

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/ce54454f-e86e-4dea-b5a9-bbaaca8a7753

### DIFF
--- a/packages/react-components/src/components/chart/baseChart.tsx
+++ b/packages/react-components/src/components/chart/baseChart.tsx
@@ -1,28 +1,49 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import { useECharts } from '../../hooks/useECharts';
 import { ChartOptions } from './types';
 import { useVisualizedDataStreams } from './useVisualizedDataStreams';
 import { convertOptions } from './converters/convertOptions';
+import { SeriesOption, YAXisComponentOption } from 'echarts';
+import { convertYAxis } from './converters/convertAxis';
+import { convertSeriesAndYAxis, reduceSeriesAndYAxis } from './converters/convertSeriesAndYAxis';
+import { HotKeys, KeyMap } from 'react-hotkeys';
+import useTrendCursors from './useTrendCursors';
+
+const keyMap: KeyMap = {
+  commandDown: { sequence: 'command', action: 'keydown' },
+  commandUp: { sequence: 'command', action: 'keyup' },
+};
 
 /**
  * Base chart to display Line, Scatter, and Bar charts.
  */
 const Chart = ({ viewport, queries, size, ...options }: ChartOptions) => {
-  const { hasError, isLoading, dataStreams } = useVisualizedDataStreams(queries, viewport);
-
-  const option = convertOptions(
-    {
-      dataStreams,
-      hasError,
-      isLoading,
-    },
-    {
+  const { isLoading, dataStreams } = useVisualizedDataStreams(queries, viewport);
+  const { axis } = options;
+  const defaultSeries: SeriesOption[] = [];
+  const defaultYAxis: YAXisComponentOption[] = [convertYAxis(axis)];
+  // Need series data for the calculation of Y for rendering a circle at TS and series lines intersections
+  const { series, yAxis } = useMemo(
+    () =>
+      dataStreams
+        .map(convertSeriesAndYAxis(options as ChartOptions))
+        .reduce(reduceSeriesAndYAxis, { series: defaultSeries, yAxis: defaultYAxis }),
+    [dataStreams]
+  );
+  const [trendCursors, setTrendCursors] = useState(options.graphic ?? []);
+  const [isInCursorAddMode, setIsInCursorAddMode] = useState(false);
+  const option = {
+    ...convertOptions({
+      ...options,
       viewport,
       queries,
       size,
-      ...options,
-    }
-  );
+      seriesLength: series.length,
+    }),
+    series,
+    yAxis,
+    graphic: trendCursors,
+  };
 
   const { ref } = useECharts({
     option,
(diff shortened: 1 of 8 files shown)


packages/react-components/src/components/chart/utils/getInfo.ts
```diff
@@ -0,0 +1,68 @@
...
+import { SizeConfig } from '../types';
+export const isDurationViewport = (viewport: Viewport): viewport is DurationViewport =>
+  (viewport as DurationViewport).duration !== undefined;
+
+// TODO: test this once echarts live mode is supported
+export const viewportToSeconds = (viewport?: Viewport) => {
+  if (viewport && isDurationViewport(viewport)) {
+    switch (viewport?.duration) {
+      case '1m':
+        return { widthInMs: 60 * 1000, initial: Date.now() - 60 * 1000 };
```
jmbuss, line 14: I think we use this parseDuration library already in this package. Might help clean this up

packages/react-components/src/components/chart/utils/getInfo.ts
```diff
@@ -0,0 +1,68 @@
+import { DurationViewport, Viewport } from '@iot-app-kit/core/src';
+import { v4 as uuid } from 'uuid';
+import { DEFAULT_MARGIN, tooltipLineColor, tooltipLineWidth } from '../eChartsConstants';
+import { ElementEvent } from 'echarts';
+import { SizeConfig } from '../types';
+export const isDurationViewport = (viewport: Viewport): viewport is DurationViewport =>
+  (viewport as DurationViewport).duration !== undefined;
+
+// TODO: test this once echarts live mode is supported
+export const viewportToSeconds = (viewport?: Viewport) => {
```
jmbuss, line 10: should this be viewport to ms?

packages/react-components/src/components/chart/utils/getInfo.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const addNewGraphic = (e: ElementEvent, size: SizeConfig, viewport: Viewport | undefined) => ({
+  id: `tooltip-${uuid()}`,
+  action: '$merge',
+  type: 'group',
+  timestampInMs: calculateTimeStamp(e.offsetX, size?.width, viewport),
+  children: [
+    {
+      type: 'line',
+      z: 101,
```
jmbuss, line 49: we should leave some comments about why this number specifically or make it a variable in the other echarts constants. would be nice to have 1 place for all the layering info so that it's easy to debug

packages/react-components/src/components/chart/utils/getInfo.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const calculateTimeStamp = (xInPixel: number, widthInPixel: number, viewport?: Viewport) => {
+  const { widthInMs, initial } = viewportToSeconds(viewport);
+  // TODO: this results in a decimal , needs to decide precision
+  const delta = (widthInMs * (xInPixel - DEFAULT_MARGIN)) / (widthInPixel - 2 * DEFAULT_MARGIN);
+  return delta + initial;
+};
+
+export const addNewGraphic = (e: ElementEvent, size: SizeConfig, viewport: Viewport | undefined) => ({
```
jmbuss, line 41: is this generic to graphics or is this trend cursor specific? Might be good to name as such and add a comment saying what this structure is

packages/react-components/src/components/chart/eChartsConstants.ts
```diff
@@ -63,3 +60,10 @@ export const DEFAULT_ECHART_OPTIONS: EChartsOption = {
   tooltip: DEFAULT_TOOLTIP,
   dataZoom: DEFAULT_DATA_ZOOM,
 };
+
+// tool tip constants
+export const tooltipColors = ['#DA7596', '#2EA597', '#688AE8'];
```
jmbuss, line 65: Can we just use the colors we use for trendlines? I think theyll need to match right?

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -1,36 +1,92 @@
...
-      queries,
-      size,
-      ...options,
-    }
+  const { isLoading, dataStreams } = useVisualizedDataStreams(queries, viewport);
+  const { axis } = options;
+  const defaultSeries: SeriesOption[] = [];
+  const defaultYAxis: YAXisComponentOption[] = [convertYAxis(axis)];
+  // Need series data for the calculation of Y for rendering a circle at TS and series lines intersections
+  const { series, yAxis } = useMemo(
```
jmbuss, line 26: does this still need to be pulled out of convertOptions? I don't see series being used in this component except to be passed into convertOptions

jmbuss (changes requested):

packages/react-components/src/components/chart/eChartsConstants.ts
mnischay: ah! this is a leftover, don't think we need this.

packages/react-components/src/components/chart/baseChart.tsx
mnischay: in this CR, we don't need it, but we will need the series data in the next set of changes.

jmbuss: Is it possible to derive the info you need from the datastreams? otherwise let's just merge the converted options with the series and yAxis. I think it's weird to pass converted options into the convertOptions function.

mnischay: dataStreams could possibly used, but then we will have to process it similar to what we do for series... would rather prefer merging the output of convertOptions with series and YAxis

jmbuss (approved):
One last comment, non-blocking since I know this is still being worked on, but let's add some block comments to the util functions. This should make it easier to know exactly how adding a new trend cursor works. I can infer what the children object is based on the style variables, but would be easier to understand in a comment.

PR #1677, opened 2023-06-30
fix(react-components): ensure provider is unsubscribed correctly

jmbuss:
## Overview
Because the provider unsubscribe call must be done asynchronously, there is a bug where when the queries are updated, the provider ref is subscribed to and immediately unsubscribed to. This is because the ref utilized in the setTimeout gets updated to the new one. This leaves the previous provider orphaned. This change moves the provider from the react ref to a singleton store. This way we don't lose reference to it when the queries change. 

## Legal

### DIFF
--- a/packages/react-components/src/hooks/useTimeSeriesData/providerStore.ts
+++ b/packages/react-components/src/hooks/useTimeSeriesData/providerStore.ts
@@ -1,0 +1,25 @@
+import { ProviderWithViewport, TimeSeriesData } from '@iot-app-kit/core';
+
+type ProviderMap = { [key in string]: ProviderWithViewport<TimeSeriesData[]> };
+const providerStore = () => {
+  const providerMap: ProviderMap = {};
+
+  const set = (id: keyof ProviderMap, provider: ProviderMap[keyof ProviderMap]) => {
+    providerMap[id] = provider;
+    return provider;
+  };
+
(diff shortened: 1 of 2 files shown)


packages/react-components/src/hooks/useTimeSeriesData/providerStore.ts
```diff
@@ -0,0 +1,25 @@
+import { ProviderWithViewport, TimeSeriesData } from '@iot-app-kit/core';
+
+type ProviderMap = { [key in string]: ProviderWithViewport<TimeSeriesData[]> };
+const providerStore = () => {
```
tracy-french, line 4: This structure would benefit from having a doc comment.

packages/react-components/src/hooks/useTimeSeriesData/providerStore.ts
```diff
@@ -0,0 +1,25 @@
...
+  };
+
+  return {
+    get,
+    set,
+    remove,
+  };
+};
+
+export const ProviderStore = providerStore();
```
tracy-french, line 25: Why not make an object literal, instead of a function, which is a singleton? Are you taking advantage of the closure around providerMap in a way that's not achievable with a plain object?

tracy-french (approved):
Two small comments which are non-blocking for merge of this high-priority bug fix. Consider returning to them as a followup.

Great work and thank you for your contribution @jmbuss!

PR #1692, opened 2023-07-10
Initial Implementation of the animation feature

parsonjs:
## Overview
Added the ability to select and play animations within scene composer (this pr had to be re done due to the accidental inclusion of a file)

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
@@ -33,6 +33,7 @@ import { colors } from '../../../../utils/styleUtils';
 import { Anchor } from '../../../three';
 import { WidgetSprite, WidgetVisual } from '../../../three/visuals';
 import svgIconToWidgetSprite from '../common/SvgIconToWidgetSprite';
+import useBindingData from '../../../../hooks/useBindingData';
 
 export interface AnchorWidgetProps {
   node: ISceneNodeInternal;
@@ -68,6 +69,8 @@ export function AsyncLoadedAnchorWidget({
   const autoRescale = useMemo(() => {
     return tagSettings.autoRescale;
   }, [tagSettings.autoRescale]);
+  const bindings = useMemo(() => (valueDataBinding ? [valueDataBinding] : undefined), [valueDataBinding]);
+  const bindingData = useBindingData(bindings).data?.at(0);
 
   const tagVisible = useViewOptionState(sceneComposerId).componentVisibilities[KnownComponentType.Tag];
 
@@ -100,12 +103,13 @@ export function AsyncLoadedAnchorWidget({
 
   // Evaluate visual state based on data binding
   const visualState = useMemo(() => {
-    const values: Record<string, unknown> = dataBindingValuesProvider(dataInput, valueDataBinding, dataBindingTemplate);
+    const values: Record<string, unknown> =
+      bindingData ?? dataBindingValuesProvider(dataInput, valueDataBinding, dataBindingTemplate);
     const ruleTarget = ruleEvaluator(defaultIcon, values, rule);
     const ruleTargetInfo = getSceneResourceInfo(ruleTarget as string);
     // Anchor widget only accepts icon, otherwise, default to Info icon
     return ruleTargetInfo.type === SceneResourceType.Icon ? ruleTargetInfo.value : defaultIcon;
-  }, [rule, dataInput, valueDataBinding, defaultIcon]);
+  }, [rule, dataInput, valueDataBinding, defaultIcon, bindingData]);
 
   const defaultVisualMap = useMemo(() => {
     // NOTE: Due to the refactor from a Widget Visual (SVG to Mesh) to a Widget Sprite (SVG to Sprite)
@@ -178,9 +182,7 @@ export function AsyncLoadedAnchorWidget({
       if (isViewing) {
         if (event.eventObject instanceof Anchor) {
           if (onWidgetClick) {
-            const dataBindingContext = !valueDataBinding?.dataBindingContext
-              ? undefined
-              : applyDataBindingTemplate(valueDataBinding, dataBindingTemplate);
+            const dataBindingContext = applyDataBindingTemplate(valueDataBinding, dataBindingTemplate);
             const componentTypes = node.components.map((component) => component.type) ?? [];
             onWidgetClick({
               componentTypes,
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/AnchorWidget.spec.tsx
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/AnchorWidget.spec.tsx
@@ -19,6 +19,9 @@ jest.mock('../../common/SvgIconToWidgetSprite', () =>
 );
 
 jest.mock('../../../../../hooks/useTagSettings', () => jest.fn());
+jest.mock('../../../../../hooks/useBindingData', () =>
+  jest.fn().mockReturnValue({ data: [{ alarm_status: 'ACTIVE' }] }),
+);
 
 jest.mock('@react-three/fiber', () => {
   const originalModule = jest.requireActual('@react-three/fiber');
@@ -72,10 +75,8 @@ describe('AnchorWidget', () => {
       dataInput: { dataFrames: [], timeRange: { from: 0, to: 1 } },
       getObject3DBySceneNodeRef,
       noHistoryStates: {
+        ...useStore('default').getState().noHistoryStates,
         componentVisibilities: { [KnownComponentType.Tag]: tagVisible },
-        toggleComponentVisibility: jest.fn(),
-        setTagSettings: jest.fn(),
-        setConnectionNameForMatterportViewer: jest.fn(),
       },
     });
   };
@@ -107,6 +108,27 @@ describe('AnchorWidget', () => {
     setStore('test-ref', 'test-ref');
     const container = renderer.create(<AnchorWidget node={node} defaultIcon={DefaultAnchorStatus.Info} />);
 
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should render correctly with data binding rule', () => {
+    setStore('test-ref', 'test-ref');
+    const container = renderer.create(
+      <AnchorWidget
+        node={node}
+        defaultIcon={DefaultAnchorStatus.Info}
+        rule={{
+          statements: [
+            {
+              expression: "alarm_status == 'ACTIVE'",
+              target: 'iottwinmaker.common.icon:Error',
+            },
+          ],
+        }}
+      />,
+    );
+
+    expect(container.root.findByType('anchor').props.visualState).toEqual('Error');
     expect(container).toMatchSnapshot();
   });
 
@@ -212,10 +234,8 @@ describe('AnchorWidget onWidgetClick', () => {
       dataInput: { dataFrames: [], timeRange: { from: 0, to: 1 } },
       getObject3DBySceneNodeRef,
       noHistoryStates: {
+        ...useStore('default').getState().noHistoryStates,
         componentVisibilities: { [KnownComponentType.Tag]: tagVisible },
-        toggleComponentVisibility: jest.fn(),
-        setTagSettings: jest.fn(),
-        setConnectionNameForMatterportViewer: jest.fn(),
       },
     });
   };
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/__snapshots__/AnchorWidget.spec.tsx.snap
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/__snapshots__/AnchorWidget.spec.tsx.snap
@@ -189,6 +189,69 @@ exports[`AnchorWidget should render correctly with an offset 1`] = `
 </group>
 `;
 
+exports[`AnchorWidget should render correctly with data binding rule 1`] = `
+<group
+  visible={true}
+>
+  <group
+    scale={
+      Vector3 {
+        "x": 1,
+        "y": 1,
+        "z": 1,
+      }
+    }
+  >
+    <lineSegments>
+      <lineBasicMaterial
+        color="#ffffff"
+      />
+      <bufferGeometry
+        attach="geometry"
+      />
+    </lineSegments>
+    <anchor
+      isSelected={true}
+      onClick={[Function]}
+      position={
+        Array [
+          0,
+          0,
+          0,
+        ]
+      }
+      scale={
+        Array [
+          0.5,
+          0.5,
+          1,
+        ]
+      }
+      visualState="Error"
+    >
+      <div
+        data-test-id="Selected"
+      />
+      <div
+        data-test-id="Info"
+      />
+      <div
+        data-test-id="Warning"
+      />
+      <div
+        data-test-id="Error"
+      />
+      <div
+        data-test-id="Video"
+      />
+      <div
+        data-test-id="Custom"
+      />
+    </anchor>
+  </group>
+</group>
+`;
+
 exports[`AnchorWidget should render correctly with non default tag settings 1`] = `
 <group
   visible={true}
--- a/packages/scene-composer/src/components/StateManager.spec.tsx
+++ b/packages/scene-composer/src/components/StateManager.spec.tsx
@@ -27,7 +27,7 @@ import DefaultErrorFallback from './DefaultErrorFallback';
 import { numberStream, stringStream, viewport } from '../../tests/data/mockDataStreams';
 import { DataStream } from '@iot-app-kit/core';
 import useActiveCamera from '../hooks/useActiveCamera';
-import { KnownComponentType, SceneComposerInternalConfig } from '..';
+import { COMPOSER_FEATURES, KnownComponentType, SceneComposerInternalConfig } from '..';
 import * as THREE from 'three';
 import { SCENE_CAPABILITY_MATTERPORT } from '../common/constants';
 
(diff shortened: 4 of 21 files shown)


packages/scene-composer/src/components/toolbars/floatingToolbar/items/AddObjectMenu.tsx
```diff
@@ -276,10 +277,14 @@ export const AddObjectMenu = (): JSX.Element => {
           uri: modelUri,
           modelType: modelType ?? ext.toUpperCase(),
         };
-
+        //does this model have an animation component if it does also add the animation component
```
haweston, line 280: Where's the check for if it has animations? Looks like this add animations to all new gltf regardless.

packages/scene-composer/src/interfaces/components.ts
```diff
@@ -36,10 +37,16 @@ export interface IModelRefComponent extends ISceneComponent {
 }
 
 export interface ISubModelRefComponent extends ISceneComponent {
+  [x: string]: any;
```
haweston, line 40: Why is there a change to subModelRefComponent?

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+import * as THREE from 'three';
+import {  useFrame } from '@react-three/fiber';
+import { GLTF } from 'three/examples/jsm/loaders/GLTFLoader';
+import { useGLTF } from '../ModelRefComponent/GLTFLoader';
+import { useSceneComposerId } from '../../../common/sceneComposerIdContext';
+import { ISceneNodeInternal, useEditorState, useViewOptionState } from '../../../store';
+
+import { IAnimationComponent } from '../../../interfaces';
+
+function pauseAnimation(gltfModel, AnimationMixer, isPaused, AnimationList){
```
haweston, line 11: the linter probably doesn't like that these don't have types on the parameters.

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+  try{
+    AnimationList.forEach( function ( clip ) {
+      const animation = THREE.AnimationClip.findByName( gltfModel, clip );
+      AnimationMixer.clipAction(animation).paused = isPaused
```
haweston, line 15: do this function both pause and unpause? from the function name I'd expect only unpause. maybe add a function comment if it does more then that.

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+    } );    
+  }
+  catch (error) {
+   // console.error(error);
+  }
+}
+
+function Animate(gltfModel, AnimationMixer, AnimationList){
```
haweston, line 23: nitpick: I might rename this to a verbName like playAnimation to match the other function. Also the class has a lot phrase starting with Animat... so it might stand out more when scanning.

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+    } );
+  }
+  catch (error) {
+    //console.log(error)
+  }
+    useFrame((state, delta) => { 
+    AnimationMixer?.update(delta)
+  });
+}
+export const isPaused = false;
```
haweston, line 37: Is this used?

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export function getClip(uri){
```
haweston, line 39: Is this repulling this from s3 via a URI? If the gltf has already be downloaded by the model system we don't want to re-download it. We'd rather have use a reference to the existing model.

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const gltf = useGLTF(
+    uri,
+  ) as GLTF;
+  return gltf.animations
+}
+const AnimationComponent = ({ component, node }: { component: IAnimationComponent; node: ISceneNodeInternal }) => {
+  try{
+  const sceneComposerId = useSceneComposerId();
```
haweston, line 47: indention is missing and will fail lint probably

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const isGlobalAnimationPaused = (!useViewOptionState(sceneComposerId).componentVisibilities["Animation"]);
+  const { getObject3DBySceneNodeRef } = useEditorState(sceneComposerId);
+  const object = getObject3DBySceneNodeRef(node.ref);
+  let gltf
```
haweston, line 51: The three below if statements should be moved into useEffect hooks and the values stored with useState to reduce churn on rerender.

packages/scene-composer/src/components/panels/scene-components/AnimationComponentEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+    
+  const onUpdateCallback = useCallback(
+    (componentPartial: any, replace?: boolean) => {
+      const componentPartialWithRef: ISceneComponentInternal = { ref: component.ref, ...componentPartial };
+      updateComponentInternal(node.ref, componentPartialWithRef);
+    },
+    [node.ref, component.ref],
+  );
+
+function updateAnimations(detail){
```
haweston, line 35: fix indentation

haweston: Turn into a useCallback to reduce rerender churn

packages/scene-composer/src/components/panels/scene-components/AnimationComponentEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+  component,
+}: AnimationComponentEditorProps) => {
+  const sceneComposerId = useContext(sceneComposerIdContext);
+  const updateComponentInternal = useStore(sceneComposerId)((state) => state.updateComponentInternal);
+  const AnimationComponent = component as IAnimationComponentInternal;
+  const animationList = getClip(AnimationComponent.uri);
+  const intl = useIntl();
+
+  
+  var selectedOptions =  AnimationComponent.currentAnimations?.map(animation => ({ label: animation, value: animation})) as OptionDefinition[]
```
tracy-french, line 24: Why use var?

packages/scene-composer/src/components/panels/scene-components/AnimationComponentEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+  node,
+  component,
+}: AnimationComponentEditorProps) => {
+  const sceneComposerId = useContext(sceneComposerIdContext);
+  const updateComponentInternal = useStore(sceneComposerId)((state) => state.updateComponentInternal);
+  const AnimationComponent = component as IAnimationComponentInternal;
+  const animationList = getClip(AnimationComponent.uri);
+  const intl = useIntl();
+
+
```
tracy-french, line 23: Why double newline?

packages/scene-composer/src/components/panels/scene-components/AnimationComponentEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+export type AnimationComponentEditorProps = IComponentEditorProps;
+
+export const AnimationComponentEditor: React.FC<AnimationComponentEditorProps> = ({
+  node,
+  component,
+}: AnimationComponentEditorProps) => {
+  const sceneComposerId = useContext(sceneComposerIdContext);
+  const updateComponentInternal = useStore(sceneComposerId)((state) => state.updateComponentInternal);
+  const AnimationComponent = component as IAnimationComponentInternal;
```
tracy-french, line 19: Why does `component` need to be casted? Can you not make `AnimationComponentEditorProps` use `IAnimationComponentInternal` as the type of `component`?

packages/scene-composer/src/components/panels/scene-components/AnimationComponentEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+    detail.selectedOptions?.forEach((element) => {
+        selectedOptions?.push(element)
+    });
+    
+    const currentAnimations = [] as string[]
+    
+    selectedOptions?.forEach( (element) => {
+        currentAnimations.push(element?.value as string)
+    });
+
```
tracy-french, line 51: This section does not appear to be idiomatic JavaScript. I say that because the imperative, mutative array operations. I would expect to see `Array.map` being used.

packages/scene-composer/src/components/panels/scene-components/AnimationComponentEditor.tsx
tracy-french: Linter off?

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { GLTF } from 'three/examples/jsm/loaders/GLTFLoader';
+import { useGLTF } from '../ModelRefComponent/GLTFLoader';
+import { useSceneComposerId } from '../../../common/sceneComposerIdContext';
+import { ISceneNodeInternal, useEditorState, useViewOptionState } from '../../../store';
+
+import { IAnimationComponent } from '../../../interfaces';
+
+function pauseAnimation(gltfModel, AnimationMixer, isPaused, AnimationList){
+  try{
+    AnimationList.forEach( function ( clip ) {
```
tracy-french, line 13: Prefer arrow functions over this syntax.

packages/scene-composer/src/interfaces/components.ts
tracy-french, line 40: Please do not use `any`. It removes type checking in TypeScript applications and it's effects can have widespread repercussions.

tracy-french:
Hi @parsonjs! I added a few comments, which I hope will be helpful. This looks super interesting! Looking forward to seeing more. Thank you for contributing!

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    AnimationList.forEach(function (clip) {
+      const animation = THREE.AnimationClip.findByName(object.animations, clip);
+      AnimationMixer.clipAction(animation).paused = isPaused;
+    });
+  } catch (error) {
+    // console.error(error);
+  }
+};
+
+const list: any[] = [];
```
haweston, line 26: I'm worried that because this variable is in the file but not scoped to a component instance that it may get strange scope changes once we you have multiple components. It's set inside the individual component instance but then the exported getClip function isn't a member variable for the component and it gets called in other files.

haweston, line 26: This list variable doesn't read easily and gets confusing quickly. It looks like a component specific list, but each component pushes just one instance of a component, obj pair to it... for each time the useEffect gets triggered. The multiple instances get ignored by the getClip as it returns early.  But it seems like this should really be a map hold one entry per animation component and that it should be moved into react statement management to control updates to it and so that it has a clear owner.

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  useFrame((state, delta) => {
+    if (AnimationMixer) {
+      AnimationMixer.update(delta);
+    }
+  });
+
+  return <Fragment />;
+};
+export function getClip(component: IAnimationComponentInternal): THREE.Object3D | undefined {
```
haweston, line 69: This isn't a getClip function. This is a get object3D function that gets consumed elsewhere for the animation list... which leads to the variable name being misleading (it's animationList but it's actually an object3D). We can just return the obj.animations instead and update the return type if we want to get all the clips.

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+import React, { Fragment, useEffect } from 'react';
+import * as THREE from 'three';
+import { useFrame } from '@react-three/fiber';
+
+import { useSceneComposerId } from '../../../common/sceneComposerIdContext';
+import { ISceneNodeInternal, useEditorState, useViewOptionState, IAnimationComponentInternal } from '../../../store';
+import { IAnimationComponent } from '../../../interfaces';
+
+const toggleIsAnimationPaused = (
+  object: THREE.Object3D,
```
haweston, line 11: Could you use https://threejs.org/docs/index.html#api/en/animation/AnimationMixer.getRoot to skip passing in the object and just pass in the AnimationMixer?

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const AnimationComponent: React.FC<AnimationComponentProps> = ({ component, node }) => {
+  const sceneComposerId = useSceneComposerId();
+  const isGlobalAnimationPaused = !useViewOptionState(sceneComposerId).componentVisibilities['Animation'];
+  let obj;
+  const { getObject3DBySceneNodeRef } = useEditorState(sceneComposerId);
+  const object = getObject3DBySceneNodeRef(node.ref);
+  const currentAnimations: any[] = [];
+  if (object) {
```
haweston, line 39: This is better as a useEffect and setState. As is it's getting set ever rerender which is potentially triggering the other useEffect that triggers on currentAnimations.

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    obj = object.getObjectByName('Scene');
+    currentAnimations.push(component.currentAnimations);
+  }
+
+  let AnimationMixer: THREE.AnimationMixer;
+  useEffect(() => {
```
haweston, line 45: We probably want to split this useEffect into two things.  We only want to update the AnimationMixer if the obj changes. We only need to set play when the obj.animations or current animation changes.  we only need to toggle pause when the global changes etc.

packages/scene-composer/src/components/three-fiber/ModelRefComponent/GLTFModelComponent.tsx
```diff
@@ -109,7 +110,11 @@ export const GLTFModelComponent: React.FC<GLTFModelProps> = ({
   const clonedModelScene = useMemo(() => {
     const result = SkeletonUtils.clone(gltf.scene);
     result.traverse((obj) => processObject(component, obj, { maxAnisotropy }));
-
+    if (result.animations.length == 0) {
```
haweston, line 113: Do we only need to copy animations for the top level object or should we add that to the result traverse?

packages/scene-composer/src/components/toolbars/floatingToolbar/items/AddObjectMenu.tsx
```diff
@@ -276,10 +277,13 @@ export const AddObjectMenu = (): JSX.Element => {
           uri: modelUri,
           modelType: modelType ?? ext.toUpperCase(),
         };
-
+        const animationComponent: IAnimationComponent = {
```
haweston, line 280: This still adds an animation component to all models even if they can't have an animation.  The gltfComponent is also hold over, it could be any model type not just gltf.

packages/scene-composer/src/components/three-fiber/ModelRefComponent/GLTFModelComponent.tsx
parsonjs: I think so gltfs can only have animations in the top level

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
parsonjs: yes

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
parsonjs: I did some tests and if you move the pausing into a different use effect you can have a situation where if you mess around with node selection an animation plays that shouldnt

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }, [object, component]);
+
+  const AnimationMixer = useMemo(() => {
+    let AnimationMixer;
+    if (scene) {
+      if (component.uri) ObjectNameMap.set(component.uri, scene);
+      AnimationMixer = new THREE.AnimationMixer(scene);
+    }
+    return AnimationMixer;
+  }, [object, scene, component, node]);
```
haweston, line 52: We can remove object & node from the dependency list

packages/scene-composer/src/components/three-fiber/AnimationComponent/AnimationComponent.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  useEffect(() => {
+    if (scene?.animations) {
+      currentAnimations?.forEach((clipGroup) => {
```
haweston, line 56: clipGroup isn't an accurate variable name for a single clip/animation

PR #1700, opened 2023-07-18
feat(dashboard): rectangle drag and drop

shlokj:
## Overview

This is the first PR for the symbol library project. Changes include allowing users to drag and drop a rectangle from the component bar.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/dashboard/src/customization/widgets/rectangleSymbol/component.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { RectangleWidget } from '../types';
+
+const RectangleWidgetComponent: React.FC<RectangleWidget> = (widget) => {
+  const fitFull = {
```
jmbuss, line 5: could this be a class maybe? Looks like we are doing a lot of 100% width height

packages/dashboard/src/customization/widgets/types.ts
```diff
@@ -90,6 +90,13 @@ export type TextProperties = {
   href?: string;
 };
 
+export type RectangleProperties = {
+  style: 'solid' | 'dashed' | 'dotted';
+  fill: string;
+  strokeColor: string;
```
jmbuss, line 96: these names are kind of specific to svg properties. What about something like color / thickness / backgroundColor? they would be more similar to the property definition we established for the new chart component

jmbuss: are these all required props or can we make some optional and set defaults?

packages/dashboard/src/customization/widgets/rectangleSymbol/component.tsx
```diff
@@ -0,0 +1,38 @@
...
+    width: '100%',
+    height: '100%',
+  };
+
+  const { style, fill, strokeColor, strokeWidth } = widget.properties;
+
+  let strokeString = '1,0';
```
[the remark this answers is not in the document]
shlokj: makes sense - i pulled the constants out into a new constants.ts in the widgets directory.

packages/dashboard/src/customization/widgets/types.ts
shlokj: > these names are kind of specific to svg properties. What about something like color / thickness / backgroundColor? they would be more similar to the property definition we established for the new chart component

right. i think we can keep `fill` as is. we can't make it `fillColor` since 'none' is an acceptable value for it. however, i do agree we can change the others:

`style` --> `borderStyle`
`strokeColor` --> `borderColor`
`strokeWidth` --> `borderThickness`

shlokj: > are these all required props or can we make some optional and set defaults?

good point. we can make all of them optional.

packages/dashboard/src/customization/widgets/rectangleSymbol/component.tsx
```diff
@@ -0,0 +1,39 @@
...
+    height: '100%',
+  };
+
+  const { borderStyle, fill, borderColor, borderThickness } = widget.properties;
+
+  let strokeString = SVG_STROKE_SOLID;
+
+  if (borderStyle == 'dashed') {
+    strokeString = SVG_STROKE_DASHED;
+  } else if (borderStyle == 'dotted') {
```
mnischay, line 17: same as above

packages/dashboard/src/customization/widgets/rectangleSymbol/icon.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
mnischay, line 1: curious: is it better to just save this as an svg file and import it ?

packages/dashboard/src/customization/api.ts
```diff
@@ -66,8 +66,14 @@ const resetMaps = () => {
   clearArr(ComponentLibraryComponentOrdering);
 };
 
-export const setupDashboardPlugins = (plugins: DashboardPlugin[]) => {
-  resetMaps();
+export const useDashboardPlugins = (plugins: DashboardPlugin[]) => {
+  const [isLoaded, setLoaded] = useState(false);
 
-  plugins.forEach((plugin) => plugin.install({ registerWidget }));
+  useEffect(() => {
```
jmbuss, line 72: I think the useeffect is actually causing more problems since it runs after render. Since we use this hook in the root component, this means that we won't have any plugins installed when the components that need them render. For instance Grid and Component Palette. We can just get rid of the useeffect. Otherwise, we'll have to delay rendering until the second pass every time. Also we need to make use of the feature flags in this hook.

jmbuss (changes requested):

packages/dashboard/src/customization/widgets/rectangleSymbol/icon.tsx
shlokj: i don't see how we would do that because the svg depends upon the props to be rendered

packages/dashboard/src/customization/widgets/rectangleSymbol/component.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { RectangleWidget } from '../types';
+import { SVG_STROKE_DASHED, SVG_STROKE_DOTTED, SVG_STROKE_SOLID } from '../constants';
+
+const RectangleWidgetComponent: React.FC<RectangleWidget> = (widget) => {
+  const fitFull = {
+    width: '100%',
+    height: '100%',
+  };
+
+  const { borderStyle, fill, borderColor, borderThickness } = widget.properties;
```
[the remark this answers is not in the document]
shlokj: added defaults

packages/dashboard/src/customization/api.ts
shlokj: done

jmbuss (approved):
Looks great!

PR #1715, opened 2023-07-19
feat(react-component): adding resizability for Charts components

mnischay:
## Overview
adding react-resizable to make the width customizable of chart and legend sections.
the legend component is using cloudscape component, this is a placeholder and will be updated based on https://quip-amazon.com/3arVALfv3isY/Legend-Table-Options

## Verifying Changes
refer to screen recording

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/5fff6409-626f-42e7-8dd3-551e77eba280


packages/react-components/src/components/chart/utils/getInfo.ts
```diff
@@ -104,11 +106,13 @@ export const calculateTrendCursorsSeriesMakers = (
       value = delta + valueMin;
     }
 
+    // TODO: precision to be decided
```
corteggiano, line 109: how will this be decided? do we have an item in Asana to remind ourselves to get back to this?

packages/react-components/src/components/chart/useTrendCursors.ts
```diff
@@ -47,10 +58,6 @@ const useTrendCursors = (
       }
     };
   }, [ref, graphic, size, isInCursorAddMode, setGraphic, viewport, theme, series, yMin, yMax]);
-
-  useEffect(() => {
```
corteggiano, line 51: nice catch! i was looking everywhere to find where the console log was coming from 😭

packages/react-components/src/components/chart/types.ts
```diff
@@ -68,7 +68,11 @@ export type ChartStyleSettings = {
   [refId: string]: ChartStyleSettingsOptions;
 };
 
-export type InternalGraphicComponentGroupOption = { timestampInMs: number } & {
+export type InternalGraphicComponentGroupOption = {
+  timestampInMs: number;
+  yAxisMarkerValue: number[];
+  headerColor: string;
+} & {
```
corteggiano, line 75: is there a reason why this is an intersection and not one type?

tracy-french: +1

packages/react-components/src/components/chart/useChartsLegend.tsx
```diff
@@ -0,0 +1,72 @@
...
+            backgroundColor: e.lineColor,
+          }}
+        ></div>
+        <div>{e.name}</div>
+      </div>
+    ),
+    isRowHeader: true,
+  };
+
+  const getTcHeader = (text: string) => text.split('\n')[0].split('|')[1].replaceAll('}', '').trim();
```
tracy-french, line 32: How did you know what to split/replace here?

packages/react-components/src/components/chart/useTrendCursors.ts
```diff
@@ -32,7 +32,18 @@ const useTrendCursors = (
...
+              size,
+              trendCursorStaticIndex++,
+              graphic,
+              setGraphic,
+              series,
+              yMin,
+              yMax,
+              viewport,
+              chart
+            )
```
tracy-french, line 46: Have you considered using an object-based API, i.e., `addTrendCursor({ e, size, [..] })`? It would remove the need to define arguments in order.

packages/react-components/src/components/chart/useChartsLegend.tsx
```diff
@@ -0,0 +1,72 @@
...
+const useChartsLegend = ({
+  graphic,
+  series,
+}: {
+  graphic: InternalGraphicComponentGroupOption[];
+  series: SeriesOption[];
+}) => {
+  const legendColumnDefinition = {
+    id: 'Legends',
+    header: <div className='base-chart-legend-col-header'>Legends</div>,
```
corteggiano, line 17: can we rename this column `properties` since its representing an asset prop in a chart

tracy-french (commented):
Overall, looks great. @corteggiano already left a review, so I'll let her own approving the PR.

packages/react-components/src/components/chart/utils/getInfo.ts
mnischay: how, I am not entirely sure and hence have created this task already https://app.asana.com/0/1204920453792313/1205045927987035/f

packages/react-components/src/components/chart/useChartsLegend.tsx
mnischay: this is the rich text format of echarts, let me add a comment here

packages/react-components/src/components/chart/useTrendCursors.ts
mnischay: yeah, if I add any more properties, il update this to an object, leaving it as is for now. No specific reason

packages/react-components/src/components/chart/types.ts
mnischay: Ideally this should be just
 ```
GraphicComponentGroupOption & {timestamp : number , yAxisMarkerValue: number[];
  headerColor: string;}
```
the typescript is not happy about GraphicComponentGroupOption, and it throws error whenever we try to access the children.

the timestamp yAxisMarkerValue and headerColor is not part of the GraphicComponentGroupOption. these fields are introduced to access then easier in legend else we will have to parse the children / series objects

packages/react-components/src/components/chart/useChartsLegend.tsx
mnischay: I am not entirely sure what you are referring here.. is it he classname ?

corteggiano (approved):

PR #1720, opened 2023-07-21
feat(react-components): add menu component

jmbuss:
## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/react-components/src/components/menu/menu.css
```diff
@@ -0,0 +1,25 @@
+.iot-context-menu-shadow {
+  box-shadow: rgba(0, 7, 22, 0.1) 0px 4px 20px 1px;
```
mnischay, line 2: is there a design token color we could use ? probably have to move to js to use the design token

packages/react-components/src/components/menu/option.tsx
```diff
@@ -0,0 +1,95 @@
...
+    else if (onKeyboardEnter && isEnterEvent(e)) onKeyboardEnter(e);
+    else if (onClick && isPointerEvent(e)) onClick(e);
+  };
+
+  const hoverStyle = hover && {
+    backgroundColor: colorBackgroundDropdownItemHover,
+    boxShadow: `0 0 0 ${spaceScaledXxxs} ${colorBorderControlDefault}`,
+  };
+
+  const borderRadius = hover ? borderRadiusDropdown : '0px';
```
mnischay, line 68: super nit, zero doesnt need px

packages/react-components/src/components/menu/positionable.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+type Position = { x: number; y: number; z?: number };
+type ReferenceElement = (setRef: RefCallback<null>) => React.ReactNode;
+
+type AbsoluteMenu = {
+  position: Position;
+  referenceElement?: never;
+};
+type RelativeMenu = {
+  position?: never;
```
mnischay, line 20: curious .. why never ?

packages/react-components/src/components/menu/index.tsx
```diff
@@ -0,0 +1,7 @@
+import { PositionableMenu, PositionableMenuProps } from './positionable';
```
mnischay, line 1: I didnt realize this is in draft, ignore my comments if its still WIP

packages/react-components/src/components/menu/positionable.tsx
jmbuss: using this to indicate an XOR type for positioning. It can either be relative or absolute. relative uses a reference element, absolute uses a position

packages/react-components/src/components/menu/menu.css
jmbuss: unfortunately no, cloudscape doesn't expose tokens for drop shadows

packages/react-components/src/components/menu/option.css
```diff
@@ -0,0 +1,10 @@
+.menu-option {
+  cursor: pointer;
+  display: flex;
+  grid-gap: 1rem;
```
mnischay, line 4: we usually use pixels ?

packages/react-components/stories/menu/menu.stories.tsx
```diff
@@ -0,0 +1,212 @@
...
+    <MenuOption label='Option 3' />
+    <MenuOption label='Option 4' />
+    <MenuOption label='Option 5' />
+  </Menu>
+);
+
+export const Absolute: ComponentStory<FC<StoryInputs>> = ({ position }) => (
+  <PositionableMenu
+    onClickOutside={() => {
+      console.log('clicked outside');
```
mnischay, line 41: is this console.log intentionally left ?

mnischay (approved):
there might be one two nits, but overall LGTM

PR #1721, opened 2023-07-24
feat(react-components): adding resizing to trend cursors

mnischay:
## Overview
on resize the TC will stay at the user place timestamp. i.e they will move along with the chart both width and height wise

## Verifying Changes
refer to the screen capture 

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/fdf2306f-2e80-4388-a12e-aad1474e1aec


packages/react-components/src/components/chart/addTrendCursor.ts
```diff
@@ -60,13 +56,14 @@ const addTCLine = (
     const graphicIndex = graphic.findIndex((g) => g.children[0].id === event.target.id);
     const timeInMs = calculateTimeStamp(event.offsetX ?? 0, size.width, viewport);
 
-    // update the x of header and close button
-    graphic[graphicIndex].children[1].x = setXWithBounds(size, event.offsetX ?? 0);
-    graphic[graphicIndex].children[2].x = setXWithBounds(size, event.offsetX ?? 0) + TREND_CURSOR_CLOSE_BUTTON_X_OFFSET;
+    // update the X value of the TC
+    graphic[graphicIndex].x = setXWithBounds(size, event.offsetX ?? 0);
+    // specifically setting the line graphic x value to 0 so that it follows the parent's X
+    graphic[graphicIndex].children[0].x = 0;
```
jmbuss, line 62: I think we should make a wrapper class or object for graphic and have some methods to modify it's properties. will be easier to understand

mnischay: broke down the ondrag handler into smaller functions. I dont think we need a refactor, but that will be out of scope for this cr.

jmbuss (approved):
great progress! I think we can address some ideas from my comments later.

PR #1746, opened 2023-07-25
feat(composer): add hook to convert data bindings to queries

sheilaXu:
## Overview
add hook to convert data bindings to queries

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/hooks/useBindingQueries.ts
+++ b/packages/scene-composer/src/hooks/useBindingQueries.ts
@@ -1,0 +1,37 @@
+import { useMemo } from 'react';
+import { Query, TimeSeriesData } from '@iot-app-kit/core';
+import { isEmpty } from 'lodash';
+
+import { useSceneComposerId } from '../common/sceneComposerIdContext';
+import { IValueDataBinding } from '../interfaces';
+import { useStore } from '../store';
+
+/**
+ * Create query objects for data bindings.
+ *
+ * @param bindings the data bindings to create query object
+ * @returns `queries` as an array of query objects for each data binding in the same order.
+ *          If the query object cannot be created for any of the bindings,
+ *          undefined will be returned in the result array at the same index.
+ */
+const useBindingQueries = (
+  bindings: IValueDataBinding[],
+): // TODO: add data type for static data when available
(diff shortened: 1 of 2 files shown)


packages/scene-composer/src/hooks/useBindingQueries.ts
```diff
@@ -0,0 +1,37 @@
...
+import { useStore } from '../store';
+
+/**
+ * Create query objects for data bindings.
+ *
+ * @param bindings the data bindings to create query object
+ * @returns `queries` as an array of query objects for each data binding in the same order.
+ *          If the query object cannot be created for any of the bindings,
+ *          undefined will be returned in the result array at the same index.
+ */
```
tracy-french, line 16: Nice doc comment!

packages/scene-composer/src/hooks/useBindingQueries.ts
```diff
@@ -0,0 +1,37 @@
+import { useMemo } from 'react';
+import { Query, TimeSeriesData } from '@iot-app-kit/core';
+import { isEmpty } from 'lodash';
+
+import { useSceneComposerId } from '../common/sceneComposerIdContext';
+import { IValueDataBinding } from '../interfaces';
+import { useStore } from '../store';
+
+/**
+ * Create query objects for data bindings.
```
tracy-french, line 10: What does data binding refer to in this context? It's a general term yet seems to be referring to a specific data type utilized by the system. Would it be helpful to expand the comment here with additional info into what the data bindings are?

tracy-french (approved):
Thank you, @sheilaXu. :)

packages/scene-composer/src/hooks/useBindingQueries.ts
sheilaXu: Data binding is a specific feature in the context of the scene composer, or even in TwinMaker. It is a set of ids (currently support entityId, componentName, propertyName) to uniquely identify a data to be used by scene widgets to change their visual state. This hook is supposed to be used internally only, so whoever use it should know the definition.

PR #1761, opened 2023-08-01
feat(dashboard): add advanced search using knowledge graph to query editor

dpportet:
## Overview
Adding advanced search mode to the query editor. This change adds two tabs to the resource exploration experience. The first is the existing query editor experience that allows drill-down exploration of assets and multi selection to choose properties. The second tab is the new advanced search which uses TwinMaker's KnowledgeGraph API's for efficiently searching through a large number of assets and properties to aid in selecting a property to add to a dashboard.

Some followups I will do:
* Using i18n strings
* Including assetName in table

## Verifying Changes
`<QueryEditor />` can replace `<ResourceExplorer /> ` in the internal dashboard implementation to view an example in our storybook or use in other components. QueryEditor will require a sitewise client and optionally the TwinMaker data source to work.

## Legal

### DIFF
--- a/packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
+++ b/packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
@@ -1,0 +1,114 @@
+import * as React from 'react';
+import SpaceBetween from '@cloudscape-design/components/space-between';
+import Button from '@cloudscape-design/components/button';
+import FormField from '@cloudscape-design/components/form-field';
+import Input from '@cloudscape-design/components/input';
+import { useState, useMemo, useCallback } from 'react';
+import { ExecuteQueryCommandOutput } from '@aws-sdk/client-iottwinmaker';
+import { createKnowledgeGraphQueryClient } from '@iot-app-kit/react-components/src/components/knowledge-graph/KnowledgeGraphQueries';
+import { TwinMakerKGQueryDataModule } from '@iot-app-kit/source-iottwinmaker';
+import { AdvancedResultsTable } from './advancedResultsTable';
+import { Property } from './model/property';
+
+export interface AdvancedSearchProps {
+  kGDatamodule: TwinMakerKGQueryDataModule;
+  onSelect: (properties: Property[]) => void;
+}
+
+/**
+ * When SiteWise sync is turned on with Knowledge Graph
+ * it adds a definition field for the propertyId and displayName
+ * and adds propertyName as a field.
+ */
+function executeQueryToProperty(res: ExecuteQueryCommandOutput) {
+  const properties: Property[] = [];
+
+  const rows = res.rows;
+  const columnDescriptions = res.columnDescriptions;
+
+  let propertyIdIndex, propertyNameIndex, displayNameIndex, assetNameIndex;
+
+  if (rows && columnDescriptions) {
+    // First get the index of each property in the rowData
+    for (let columnNumber = 0; columnNumber < columnDescriptions.length; columnNumber++) {
+      const itemName = columnDescriptions[columnNumber].name;
+      switch (itemName) {
+        case 'propertyId':
+          propertyIdIndex = columnNumber;
+          break;
+        case 'propertyName':
+          propertyNameIndex = columnNumber;
+          break;
+        case 'displayName':
+          displayNameIndex = columnNumber;
+          break;
+        case 'assetName':
+          assetNameIndex = columnNumber;
+          break;
+      }
+    }
+
+    // Construct the properties using the previously determined index
+    if (
+      propertyIdIndex != undefined &&
+      propertyNameIndex != undefined &&
+      displayNameIndex != undefined &&
+      assetNameIndex != undefined
+    ) {
+      for (const row of rows) {
+        const rowData = row.rowData;
+        if (rowData) {
+          // Knowledge Graph returns rows that aren't actual properties
+          // that contain assetId and don't have a propertyName so we don't want to include those
+          if (rowData[propertyIdIndex]) {
+            properties.push({
+              propertyId: rowData[propertyIdIndex] as string,
+              propertyName: rowData[propertyNameIndex] as string,
+              displayName: rowData[displayNameIndex] ? (rowData[displayNameIndex] as string) : '-',
+              assetName: rowData[assetNameIndex] as string,
+            });
+          }
+        }
+      }
+    } else {
+      console.log('Error parsing Knowledge Graph response. Unable to determine field indices.');
+    }
+  }
+
+  return properties;
+}
+
+/** Advanced search for exploring your AWS IoT SiteWise properties. */
+export function AdvancedSearch({ kGDatamodule, onSelect }: AdvancedSearchProps) {
+  const [searchTermInput, setSearchTermInput] = useState('');
+  const [properties, setProperties] = useState<Property[]>([]);
+
+  const kGResponse = (res: ExecuteQueryCommandOutput) => {
+    setProperties(executeQueryToProperty(res));
+  };
+
+  const knowledgeGraphQueryClient = useMemo(() => {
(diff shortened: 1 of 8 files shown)


packages/dashboard/src/components/queryEditor/queryEditor.tsx
```diff
@@ -1,44 +1,93 @@
...
-  );
+  const resourceExplorerTab = {
+    label: 'Browse assets',
+    id: 'first',
+    content: (
+      <Box padding='s'>
+        <SpaceBetween size='m' direction='vertical'>
+          <TextContent>
+            <h3>Browse Assets</h3>
+          </TextContent>
```
tracy-french, line 51: Why are you not using the Cloudscape `<Header />` component?

packages/dashboard/src/components/queryEditor/advancedSearch/model/property.ts
```diff
@@ -0,0 +1,5 @@
+export interface Property {
+  propertyId: string;
+  propertyName: string;
+  displayName: string;
+}
```
tracy-french, line 5: Is it possible to pull this type from the SDK?

packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const knowledgeGraphQueryClient = useMemo(() => {
+    return createKnowledgeGraphQueryClient(kGDatamodule, kGResponse);
+  }, [kGDatamodule, kGResponse]);
+
+  const onSearchClicked = useCallback(() => {
+    if (searchTermInput) {
+      knowledgeGraphQueryClient.findPropertiesByEntityOrPropertyNameOrDisplayName(searchTermInput);
+    }
+  }, [knowledgeGraphQueryClient, searchTermInput]);
```
tracy-french, line 66: What is `useCallback` being utilized here for?

packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+/*
+    The return types are incorrect for Knowledge Graph
+    when selecting an edge instead of a node
+    so we can't properly type here.
+
+    When SiteWise sync is turned on with Knowledge Graph
+    it adds a definition field for the propertyId and displayName
+    and adds propertyName as a field.
+  */
```
tracy-french, line 26: Why is the formatting off on this comment?

packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+              propertyName: data.propertyName,
+            });
+          }
+        }
+      }
+    }
+  }
+
+  return properties;
+}
```
tracy-french, line 47: Iterating over a 2D array can be expensive. Is it possible we'll run into scalability issues here?

packages/dashboard/src/components/queryEditor/__tests__/advancedSearch.spex.tsx
```diff
@@ -0,0 +1,471 @@
+import { QueryEditor } from '../queryEditor';
+import { type IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
+import React from 'react';
+import { render, screen, waitFor, fireEvent } from '@testing-library/react';
```
tracy-french, line 4: Why not use '@testing-library/user-event' for testing events? I believe I am using this in the other tests.

packages/dashboard/src/components/queryEditor/__tests__/advancedSearch.spex.tsx
```diff
@@ -0,0 +1,471 @@
...
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { TwinMakerKGQueryDataModule } from '@iot-app-kit/source-iottwinmaker';
+
+const queryClient = new QueryClient();
+
+const Wrapper = ({ children }: React.PropsWithChildren) => (
+  <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+);
+
+describe('Advanced Search', () => {
```
tracy-french, line 15: Great tests!

tracy-french (approved):
@dpportet thank you for your contribution! This is gonna be a big win! There's a few things to consider moving forward and we can focus on those before completing the integration. Keep it up!

chejimmy (approved):
LGTM!

packages/dashboard/src/components/queryEditor/queryEditor.tsx
```diff
@@ -1,44 +1,93 @@
...
+      <QueryEditorErrorBoundary>
+        <Tabs tabs={[resourceExplorerTab, advancedSearchTab]} />
+      </QueryEditorErrorBoundary>
+    );
+  } else {
+    return (
+      <QueryEditorErrorBoundary>
+        <Tabs tabs={[resourceExplorerTab]} />
+      </QueryEditorErrorBoundary>
+    );
```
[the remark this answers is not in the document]
dpportet: I'll take a look in a followup PR.

packages/dashboard/src/components/queryEditor/queryEditor.tsx
dpportet: Because I did not know about the Cloudscape `<Header /> ` component but I will add that in a followup PR.

packages/dashboard/src/components/queryEditor/advancedSearch/model/property.ts
dpportet: I don't think there's a single interface that contains all the fields we want to display in the table. My thinking was this interface represents the fields we're displaying in the table and so any response we would want to convert into this form. 

Specifically for Knowledge Graph responses, when we select different fields in the query (for example `Select e.entityName as assetName` then the response does not have Typescript types for the fields, so this interface represents the form we will convert the response to in order to display in the table.

packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
dpportet: I don't have a strong reason, in the `KnowledgeGraphPanel` the client was being used like this so I copied it. It stores the response so if the same request is made again it won't re-request the data. If there's any reason not to use it I can change it.

packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
dpportet: I thought the linting would fix it, updated it in a new commit I pushed.

packages/dashboard/src/components/queryEditor/advancedSearch/advancedSearch.tsx
dpportet: I changed this in my latest commit. As we discussed, the new parsing should be O(n) + O(f) = O(n) time complexity where n is the number of matching properties and f is the number of fields we are selecting. This shouldn't result in scalability issues, but we should specifically test the use case of thousands of matching properties to ensure the parsing doesn't cause scalability issues.

packages/dashboard/src/components/queryEditor/__tests__/advancedSearch.spex.tsx
dpportet: I didn't know about `@testing-library/user-event` and found `fireEvent` from the `@testing-library/react` docs. I can switch to `user-event`, what's the benefit?

packages/dashboard/src/components/queryEditor/__tests__/advancedSearch.spex.tsx
dpportet: Thanks

PR #1762, opened 2023-08-01
feat(echarts): threshold support

corteggiano:
## Overview
Support thresholds in echarts bar, scatter, line chart
image

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/converters/convertThresholds.ts
+++ b/packages/react-components/src/components/chart/converters/convertThresholds.ts
@@ -1,0 +1,68 @@
+import { AnnotationValue, COMPARISON_OPERATOR, ComparisonOperator, Threshold } from '@iot-app-kit/core';
+import { COMPARATOR_MAP } from '../../../common/constants';
+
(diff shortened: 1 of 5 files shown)


packages/react-components/src/components/chart/converters/convertThresholds.ts
```diff
@@ -0,0 +1,66 @@
...
+    case COMPARISON_OPERATOR.GT:
+    case COMPARISON_OPERATOR.GTE:
+    case COMPARISON_OPERATOR.EQ:
+      return value;
+    case COMPARISON_OPERATOR.LT:
+    case COMPARISON_OPERATOR.LTE:
+      return 0;
+    default:
+      return undefined;
+  }
```
tracy-french, line 15: I'm curious, is the default actually possible? If it's not possible, depending on your philosophy, this could be a great place to throw an error and declare an explicit return annotation as a union with `never`. What's going to happen is when a new operator added, this function will throw an error to tell the developer they need to extend this switch statement. Otherwise, its possible this function may be skipped if no immediate downstream impact is observed.

tracy-french (approved):
Looks great! Consider my question about the switch statements in future iterations. Thank you for your work, @corteggiano!

PR #1767, opened 2023-08-03
feat(react-component): adding context menu per chart

mnischay:
## Overview
Adding context menu for charts. this will be per chart i.e when in dashboard this will not be shown instead the dashboard will be updated with the TC options in the follow up CR. 

## Verifying Changes
Refer to the screen capture

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/76a8f5e2-6daa-45b6-9fb6-e46372f96cd0

### DIFF
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -99,6 +99,7 @@
     "@iot-app-kit/core-util": "7.1.0",
     "@iot-app-kit/source-iottwinmaker": "7.1.0",
     "color": "^4.2.3",
+    "copy-to-clipboard": "^3.3.3",
     "d3-array": "^3.2.3",
     "d3-format": "^3.1.0",
     "d3-shape": "^3.2.0",
--- a/packages/react-components/src/components/chart/contextMenu/ChartContextMenu.spec.tsx
+++ b/packages/react-components/src/components/chart/contextMenu/ChartContextMenu.spec.tsx
@@ -1,0 +1,20 @@
+import { describe, expect } from '@jest/globals';
+import { render } from '@testing-library/react';
+import React from 'react';
+import ChartContextMenu from './ChartContextMenu';
+
+describe('ChartContextMenu', () => {
+  const menuOptionClickHandler = jest.fn();
+  const onOutSideClickHandler = jest.fn();
+  const component = render(
+    <ChartContextMenu
+      menuOptionClickHandler={menuOptionClickHandler}
+      onOutSideClickHandler={onOutSideClickHandler}
+      position={{ x: 100, y: 100 }}
+      trendCursors={[]}
+    />
+  );
+  it('renders', () => {
+    expect(component).not.toBeNull();
+  });
+});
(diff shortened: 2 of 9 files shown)


packages/react-components/package.json
```diff
@@ -99,6 +99,7 @@
     "@iot-app-kit/core-util": "7.1.0",
     "@iot-app-kit/source-iottwinmaker": "7.1.0",
     "color": "^4.2.3",
+    "copy-to-clipboard": "^3.3.3",
```
tracy-french, line 102: https://github.com/sudodoki/copy-to-clipboard uses a deprecated API and hasn't been updated in a year. Might want to consider https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard#writing_to_the_clipboard.

packages/react-components/src/components/chart/contextMenu/ChartContextMenu.spec.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <ChartContextMenu
+      menuOptionClickHandler={menuOptionClickHandler}
+      onOutSideClickHandler={onOutSideClickHandler}
+      position={{ x: 100, y: 100 }}
+      trendCursors={[]}
+    />
+  );
+  it('renders', () => {
+    expect(component).not.toBeNull();
+  });
```
tracy-french, line 19: Nothing more we should test, such as the existence of each action? Where would we catch the issue if one of the context menu actions was disable on accident?

packages/react-components/src/components/chart/hooks/useTrendCursorsEvents.ts
```diff
@@ -151,17 +172,46 @@ const useTrendCursorsEvents = ({
...
+  const copyTrendCursorData = (posX: number) => {
+    const timestampOfClick = calculateTimeStamp(posX, sizeRef.current.width, viewportRef.current);
+    const toBeCopiedGraphicIndex = calculateNearestTcIndex(graphicRef.current, timestampOfClick);
+    // using copy-to-clipboard library to copy in a Excel sheet pastable format
+    copy(formatCopyData(graphicRef.current[toBeCopiedGraphicIndex], seriesRef.current), { format: 'text/plain' });
+  };
+
+  // this handles the user interaction via context menu
+  const onContextMenuClickHandler = ({ action, posX }: { action: Action; posX: number }) => {
+    switch (action) {
```
tracy-french, line 201: Without a default, what happens if a fourth action is added and this switch statement isn't updated?

tracy-french (approved):
Thank you @mnischay! Please consider my feedback in future iterations. :)

packages/react-components/src/components/chart/hooks/useTrendCursorsEvents.ts
mnischay: No action occurs, did not want to have any action if by chance someone changes the Action type

packages/react-components/src/components/chart/contextMenu/ChartContextMenu.spec.tsx
mnischay: I am looking into how can we trigger events on charts, havent been successful so far.

packages/react-components/package.json
mnischay: given the number of usages / downloads, i assumed it to be ok to use... https://www.npmjs.com/package/copy-to-clipboard

PR #1768, opened 2023-08-03
fix(data overlay): add onWidgetClick and onSelectionChange event support to data overlays

haweston:
## Overview
Click an overlay doesn't have data binding information in the onSelectionChange event fired from the SceneViewer and the onWidgetClick event is fired at all.

## Verifying Changes
Create scene with dataoverlay that has 2+ data bindings added to it.

Connect to scene in SceneViewer mode on storybook

Click the overlay widget

verify that the storybook actions log both selection and widget change events with databindings that look like this for array of 2 data bindings:

Screenshot 2023-08-03 at 4 16 08 PM

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/StateManager.tsx
+++ b/packages/scene-composer/src/components/StateManager.tsx
@@ -31,16 +31,9 @@ import {
 } from '../interfaces';
 import { SceneLayout } from '../layouts/SceneLayout';
 import useLifecycleLogging from '../logger/react-logger/hooks/useLifecycleLogging';
-import {
-  IAnchorComponentInternal,
-  ICameraComponentInternal,
-  IEntityBindingComponentInternal,
-  RootState,
-  useStore,
-  useViewOptionState,
-} from '../store';
+import { ICameraComponentInternal, RootState, useStore, useViewOptionState } from '../store';
 import { getCameraSettings } from '../utils/cameraUtils';
-import { applyDataBindingTemplate } from '../utils/dataBindingTemplateUtils';
+import { getAdditionalComponentData } from '../utils/eventDataUtils';
 import { combineTimeSeriesData, convertDataStreamsToDataInput } from '../utils/dataStreamUtils';
 import { findComponentByType } from '../utils/nodeUtils';
 import sceneDocumentSnapshotCreator from '../utils/sceneDocumentSnapshotCreator';
@@ -158,32 +151,8 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
     if (onSelectionChanged) {
       const node = getSceneNodeByRef(selectedSceneNodeRef);
       const componentTypes = node?.components.map((component) => component.type) ?? [];
+      const additionalComponentData: AdditionalComponentData[] = getAdditionalComponentData(node, dataBindingTemplate);
 
-      const tagComponent = findComponentByType(node, KnownComponentType.Tag) as IAnchorComponentInternal;
-      const entityBindingComponent = findComponentByType(
-        node,
-        KnownComponentType.EntityBinding,
-      ) as IEntityBindingComponentInternal;
-      const additionalComponentData: AdditionalComponentData[] = [];
-      if (tagComponent) {
-        additionalComponentData.push({
-          chosenColor: tagComponent.chosenColor,
-          navLink: tagComponent.navLink,
-          dataBindingContext: !tagComponent.valueDataBinding?.dataBindingContext
-            ? undefined
-            : applyDataBindingTemplate(tagComponent.valueDataBinding, dataBindingTemplate),
-        });
-      }
-      // Add entityID info part of additional component data
-      // We assumed IDataBindingMap will have only one mapping as data binding
-      // will always have only one entity data.
-      if (entityBindingComponent) {
-        additionalComponentData.push({
-          dataBindingContext: !entityBindingComponent?.valueDataBinding?.dataBindingContext
-            ? undefined
-            : entityBindingComponent?.valueDataBinding.dataBindingContext,
-        });
-      }
       onSelectionChanged({
         componentTypes,
         nodeRef: selectedSceneNodeRef,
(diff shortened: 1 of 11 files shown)


packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayContainer.tsx
```diff
@@ -57,12 +62,34 @@ export const DataOverlayContainer = ({ component, node }: DataOverlayContainerPr
...
+              dataBindingContext.push(applyDataBindingTemplate(item.valueDataBinding, dataBindingTemplate));
+            }
+          });
+          const componentTypes = node.components.map((component) => component.type) ?? [];
+          onWidgetClick({
+            componentTypes,
+            nodeRef: node.ref,
+            additionalComponentData: [
+              {
+                dataBindingContext,
```
sheilaXu, line 80: should it be `dataBindingContexts`?

packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -184,6 +188,15 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
...
         });
       }
+      if (dataOverlayComponent) {
+        const dataBindingContext: unknown[] = [];
+        dataOverlayComponent.valueDataBindings.forEach((item) => {
+          if (item.valueDataBinding) {
+            dataBindingContext.push(applyDataBindingTemplate(item.valueDataBinding, dataBindingTemplate));
+          }
+        });
+        additionalComponentData.push({ dataBindingContext });
```
sheilaXu, line 198: same here `dataBindingContexts`

packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -184,6 +188,15 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
             : entityBindingComponent?.valueDataBinding.dataBindingContext,
         });
       }
+      if (dataOverlayComponent) {
```
sheilaXu, line 191: I remember made comment previously related to this part that we shouldn't push additionalComponentData in this hard coded order. the component data should match with the type of components in `componentTypes`. 

so we should loop through the `componentTypes` array, and foreach type, check its type to see what additional data to be pushed. for type that doesn't have additional data, we should either push `undefined` or empty at that index so the `componentTypes` and `additionalComponentData` can match

haweston: Yeah now that we have more than 2. I think that should be a quick change.

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayContainer.tsx
haweston: Agreed, I'll fix.

packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -159,31 +188,26 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
...
-      if (entityBindingComponent) {
-        additionalComponentData.push({
-          dataBindingContext: !entityBindingComponent?.valueDataBinding?.dataBindingContext
-            ? undefined
-            : entityBindingComponent?.valueDataBinding.dataBindingContext,
-        });
-      }
+      node?.components.forEach((component) => {
+        const type = component.type;
+        let actualComponent: any;
```
sheilaXu, line 194: what is this?

haweston: That's dead code from a middle attempt at redoing the switch to pass lint. Removing it.

packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -148,6 +151,32 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
...
+  };
+  const getBindingsFromDataOverlay = (component: IDataOverlayComponentInternal): IDataOverlayInfo => {
+    const dataBindingContexts: unknown[] = [];
+    component.valueDataBindings.forEach((item) => {
+      if (item.valueDataBinding) {
+        dataBindingContexts.push(applyDataBindingTemplate(item.valueDataBinding, dataBindingTemplate));
+      }
+    });
+    return { dataBindingContexts };
+  };
```
sheilaXu, line 178: useCallback?

packages/scene-composer/src/utils/eventDataUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+      case KnownComponentType.EntityBinding:
+        additionalComponentData.push(getBindingsFromEntityBinding(component as IEntityBindingComponentInternal));
+        break;
+      case KnownComponentType.DataOverlay:
+        additionalComponentData.push(
+          getBindingsFromDataOverlay(component as IDataOverlayComponentInternal, dataBindingTemplate),
+        );
+        break;
+      default:
+        break;
```
sheilaXu, line 74: we should return `undefined` or `{}` for default case so when we have a node with 
```
components: [{ type: 'Random' }, { type: 'Tag', ...}]
```
the onSelectionChanged event will return 
```
{
  componentTypes: ['Random', 'Tag'],
  additionalComponentData: [undefined/{}, {...<some data binding>...}]
}
```
so consumer knows the `<some data binding>` is for Tag component

haweston: Right that will keep the component type count and the additional data counts the same.

packages/scene-composer/src/components/StateManager.tsx
haweston: moved to static util function library instead

sheilaXu (approved):

PR #1769, opened 2023-08-03
feat(react-components): add a sitewise connected chart story

jmbuss:
## Overview
Adding a story for base chart that uses a real sitewise query.

Adds utils for creating the query with the correct clients.

## Legal

### DIFF
--- a/packages/react-components/.env.example
+++ b/packages/react-components/.env.example
@@ -1,0 +1,9 @@
+AWS_ACCESS_KEY_ID=xxxxxxxxxx
+AWS_SECRET_ACCESS_KEY=xxxxxxxxxx
+AWS_SESSION_TOKEN=xxxxxxxxxx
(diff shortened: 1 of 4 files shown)


chejimmy (approved):
Nice! Fine addition to the collection!

packages/react-components/.env.example
```diff
@@ -0,0 +1,9 @@
+AWS_ACCESS_KEY_ID=xxxxxxxxxx
```
chejimmy, line 1: instruction/guide to setup the env?

mnischay: +1

mnischay (approved):

tracy-french (approved):
Great stuff!

packages/react-components/.env.example
jmbuss: yup, these are the variables you need to run that story!

PR #1772, opened 2023-08-04
fix(react-components): add memoization for chart reactivity

jmbuss:
## Overview
Many structural changes included in this PR.

useECharts has been split into several smaller hooks.
1. useEcharts -> initializes echarts and attaches to the dom
2. useResizeableEChart -> handles resizing of the chart and updating echart
3. useGroupableEChart -> updates the echarts groupid
4. useLoadableEChart -> updates the loading animation state
5. useEChartOptions -> handles updating echarts options

added a hook for deriving a chart id

added a hook for calculated style settings for each datastream

added a hook for converted series and yAxis

moved base chart stories into a separate story so that we can utilize styorbook argtypes which allow us to more easily test prop reactivity.

## Legal


packages/react-components/src/components/chart/hooks/useTrendCursors.ts
```diff
@@ -13,12 +13,12 @@ const useTrendCursors = ({
   series,
   yMax,
   yMin,
-  chartId,
+  // chartId,
```
mnischay, line 16: can we uncomment this ? we can remove this launch if its logging too much.

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -63,6 +67,15 @@ const Chart = ({ viewport, queries, size = { width: 500, height: 500 }, ...optio
...
   const { ref } = useECharts({
     option,
+    settings: {
+      /**
+       * if datastreams changes because we've modified the query
+       * ensure that series is updated - https://echarts.apache.org/en/api.html#echartsInstance.setOption
+       * TECHDEBT: hardcoding this prevents the normal merge animation from happening when new data points
+       * are added.
+       */
+      replaceMerge: ['series'],
```
mnischay, line 77: in the case of a series being added or deleted i.e query change, can we detect the delta and one update the options for one cycle / remove the series similar to the delete of the trend cursors ? replaceMerge may be a expensive and affect performance ?

packages/react-components/src/components/chart/hooks/useTrendCursors.ts
jmbuss: ahh yup, commented it out so I could see my own debug logs :D

packages/react-components/src/components/chart/baseChart.tsx
jmbuss: Ya, I had planned to circle back and optimize this some more. I left a techdebt tag so we don't forget. I want to include this in the refactor of the series / yAxis converter since changes to either should set this setting. I can do that now, but I was afraid it might cause too many conflicts?

mnischay (changes requested):
two major callouts..
1. setting the options in series handler as an exception not everytime... (not a deal breaker, but something to think about )
2. ymax and ymin has to be set for the yaxis, else all calculation will be wrong (sorry! this def needs change :| )

PR #1777, opened 2023-08-07
feat(react-component): adding sorting ability for the chart legends

mnischay:
## Overview
adding sorting ability for the legend table. using the out of the box useCollection hook for sorting. 

## Verifying Changes
refer to screen capture

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
+++ b/packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
@@ -45,11 +45,15 @@ const useChartsLegend = ({
   ]);
   const [items, setItems] = useState<Array<object>>([]);
   useEffect(() => {
-    const tcColumnDefinitions = graphic.map((g) => ({
-      id: g.id as string,
-      header: getHeaderNode(g),
-      cell: (e: { [x: string]: string }) => e[g.id as string],
-    }));
+    const tcColumnDefinitions = graphic.map((g) => {
+      const id = g.id as string;
+      return {
+        id,
+        header: getHeaderNode(g),
+        cell: (e: { [x: string]: number }) => e[id],
+        sortingField: id,
+      };
+    });
     setColumnDefinitions([legendColumnDefinition, ...tcColumnDefinitions]);
 
     // yAxisMarkerValue contains the y values for each TC
--- a/packages/react-components/src/components/chart/legend/legend.tsx
+++ b/packages/react-components/src/components/chart/legend/legend.tsx
@@ -5,12 +5,19 @@ import useChartsLegend from '../hooks/useChartsLegend';
 import { SeriesOption } from 'echarts';
 
 import './legend.css';
+import { useCollection } from '@cloudscape-design/collection-hooks';
 
 const Legend = (legendOptions: { series: SeriesOption[]; graphic: InternalGraphicComponentGroupOption[] }) => {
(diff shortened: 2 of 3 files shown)


packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
```diff
@@ -45,11 +45,15 @@ const useChartsLegend = ({
   ]);
   const [items, setItems] = useState<Array<object>>([]);
   useEffect(() => {
-    const tcColumnDefinitions = graphic.map((g) => ({
-      id: g.id as string,
-      header: getHeaderNode(g),
-      cell: (e: { [x: string]: string }) => e[g.id as string],
-    }));
+    const tcColumnDefinitions = graphic.map((g) => {
+      const id = g.id as string;
```
boweihan, line 49: Looks like this comes from `OptionId` in e-charts - https://github.com/apache/echarts/blob/master/src/util/types.ts#L626. Better to check/validate or cast this to a string rather than assume. I know this isn't directly your changes but just something I noticed. Accidentally pulling a number through could cause side effects on downstream operations.

packages/react-components/src/components/chart/legend/legend.tsx
```diff
@@ -5,12 +5,19 @@ import useChartsLegend from '../hooks/useChartsLegend';
 import { SeriesOption } from 'echarts';
 
 import './legend.css';
+import { useCollection } from '@cloudscape-design/collection-hooks';
 
 const Legend = (legendOptions: { series: SeriesOption[]; graphic: InternalGraphicComponentGroupOption[] }) => {
-  const { items, columnDefinitions } = useChartsLegend(legendOptions);
+  const { items: allItems, columnDefinitions } = useChartsLegend(legendOptions);
+
+  const { items, collectionProps } = useCollection(allItems, {
```
boweihan, line 13: Neat!

packages/react-components/src/components/chart/legend/legend.tsx
```diff
@@ -5,12 +5,19 @@ import useChartsLegend from '../hooks/useChartsLegend';
...
+    sorting: {},
```
boweihan, line 14: Does this apply any default sort based on the behaviour of `useCollection`? Any behaviour change that we can test?

mnischay: useCollection is from cloudscape so did not add any specific test for sorting, given its handled outside our component

packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
mnischay: This ID is set by us during creation of a TC, so it will be always a string, The type is provided by the echarts and so they have kept it as wide as possible...  I guess we could convert it instead of just casting... ```String(g.id)``` which may avoid run time error

packages/react-components/src/components/chart/legend/legend.tsx
mnischay: no default sort is applied, we could, but letting it to render the initial data as is

packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
boweihan: Got it, better to cast imo, doesn't have to be done in this PR.

boweihan (approved):

PR #1778, opened 2023-08-08
feat(composer): implement useBindingData and integrate to Tag

sheilaXu:
## Overview
- implement useBindingData that can fetch data for all data bindings
- integrate the hook to Tag

https://github.com/awslabs/iot-app-kit/assets/9315598/f639f630-3e85-401b-b287-850367f5f5ea

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/StateManager.tsx
+++ b/packages/scene-composer/src/components/StateManager.tsx
@@ -24,6 +24,7 @@ import useActiveCamera from '../hooks/useActiveCamera';
 import useMatterportViewer from '../hooks/useMatterportViewer';
 import {
   AdditionalComponentData,
+  COMPOSER_FEATURES,
   ExternalLibraryConfig,
   KnownComponentType,
   KnownSceneProperty,
@@ -38,6 +39,7 @@ import { combineTimeSeriesData, convertDataStreamsToDataInput } from '../utils/d
 import { findComponentByType } from '../utils/nodeUtils';
 import sceneDocumentSnapshotCreator from '../utils/sceneDocumentSnapshotCreator';
 import { createStandardUriModifier } from '../utils/uriModifiers';
+import useFeature from '../hooks/useFeature';
 
 import IntlProvider from './IntlProvider';
 import { LoadingProgress } from './three-fiber/LoadingProgress';
@@ -87,10 +89,10 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
   const matterportModelId = useStore(sceneComposerId)((state) =>
     state.getSceneProperty<string>(KnownSceneProperty.MatterportModelId),
   );
-  const { connectionNameForMatterportViewer, setConnectionNameForMatterportViewer } =
+  const { connectionNameForMatterportViewer, setConnectionNameForMatterportViewer, setViewport, setAutoQueryEnabled } =
     useViewOptionState(sceneComposerId);
   const { enableMatterportViewer } = useMatterportViewer();
-
+  const autoQueryEnabled = !!config?.featureConfig?.[COMPOSER_FEATURES.AutoQuery];
   const dataProviderRef = useRef<ProviderWithViewport<TimeSeriesData[]> | undefined>(undefined);
   const prevSelection = useRef<string | undefined>(undefined);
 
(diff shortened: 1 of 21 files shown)


packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -101,6 +103,8 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
     [sceneContentUri, baseUrl],
   );
 
+  const isViewing = config.mode === 'Viewing';
```
mumanity, line 106: nit-pick: When comparing strings I'm a big fan of using `.toLowerCase()` to ensure that the check is not case-sensitive. 

ex: `const isViewing = config.mode.toLowerCase() === 'viewing';`

mumanity (approved):

packages/scene-composer/src/components/StateManager.tsx
sheilaXu: the case shouldn't have any problem because the type of mode is `'Editing' | 'Viewing'` and when you compare it with other strings, there will be an error showing `This comparison appears to be unintentional because the types 'OperationMode | undefined' and '"viewing"' have no overlap.`

packages/scene-composer/src/hooks/useBindingData.ts
```diff
@@ -0,0 +1,88 @@
...
+          settings: {
+            // only support default settings for now until when customization is needed
+            fetchFromStartToEnd: true,
+          },
+        });
+
+        provider.subscribe({
+          // TODO: support static data
+          next: (results: TimeSeriesData[]) => {
+            if (isEmpty(results.at(0)?.dataStreams)) {
```
divya-sea, line 53: Make it a constant `const dataStreams = results?.[0]?.dataStreams`;

packages/scene-composer/src/hooks/useBindingData.ts
```diff
@@ -0,0 +1,88 @@
...
+              log?.info('No data returned');
+              return;
+            }
+
+            if (results.length > 1 || results[0].dataStreams.length > 1) {
```
divya-sea, line 58: here also can be simplified `dataStreams.length`. This is easy to read.

packages/scene-composer/src/hooks/useBindingData.ts
```diff
@@ -0,0 +1,88 @@
...
+              log?.warn('Multiple data returned, only use the first one');
+            }
+
+            const newData = [...(data.current ?? [])];
+            newData[index] = {
+              [(bindings[index]?.dataBindingContext as ITwinMakerEntityDataBindingContext)?.propertyName]:
+                results[0].dataStreams[0].data[results[0].dataStreams[0].data.length - 1]?.y,
```
divya-sea, line 65: Again this can be simplified `dataStreams[0]?.data.length`

packages/scene-composer/src/hooks/useBindingData.ts
```diff
@@ -0,0 +1,88 @@
...
+            };
+            data.current = newData;
```
divya-sea, line 67: Here you can simply setData(newdata)

packages/scene-composer/src/hooks/useBindingData.ts
```diff
@@ -0,0 +1,88 @@
...
+            setLastDataTime(Date.now());
+          },
+        });
+        return provider;
+      }
+      return undefined;
```
divya-sea, line 73: We don't need this if we return at the beginning of the code.

packages/scene-composer/src/hooks/useBindingData.ts
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    return () => {
+      providers?.forEach((provider) => provider?.unsubscribe());
+    };
+  }, [queries, autoQueryEnabled, viewport]);
+
+  const result = useMemo(() => {
```
divya-sea, line 81: you can simply return useMemo(() => ({data}),[data])

packages/scene-composer/src/hooks/useBindingData.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        };
+      }
+      return undefined;
+    });
+    let hook;
+    act(() => {
+      hook = renderHook(() => useBindingData(bindings));
+    });
+
+    expect(hook.result.current.data).toEqual(expectedData);
```
divya-sea, line 68: Should not this call unsubscribe too at the end?

PR #1781, opened 2023-08-08
fix(react-component): removing the dependency of the yMin and Ymax and reading the series line markers for TC from echarts

mnischay:
removing the dependency of the yMin and Ymax and reading the series line markers for TC from echarts
No net new changes except that instead of ymin and ymax, reading the pixel value from echarts

## Legal


jmbuss (approved):
👍 👍

PR #1783, opened 2023-08-08
fix(react-components): add memoization for chart reactivity and refactor echarts hook

jmbuss:
## Overview
Many structural changes included in this PR.

useECharts has been split into several smaller hooks.
1. useEcharts -> initializes echarts and attaches to the dom
2. useResizeableEChart -> handles resizing of the chart and updating echart
3. useGroupableEChart -> updates the echarts groupid
4. useLoadableEChart -> updates the loading animation state
5. useEChartOptions -> handles updating echarts options

* added a hook for deriving a chart id

* added a hook for calculated style settings for each datastream

* added a hook for converted series and yAxis

moved base chart stories into a separate story so that we can utilize styorbook argtypes which allow us to more easily test prop reactivity.

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/baseChart.tsx
+++ b/packages/react-components/src/components/chart/baseChart.tsx
@@ -1,20 +1,26 @@
-import React, { SyntheticEvent, useMemo, useState } from 'react';
-import { useECharts } from '../../hooks/useECharts';
+import React, { useState } from 'react';
+import {
+  useECharts,
+  useResizeableEChart,
+  useGroupableEChart,
+  useLoadableEChart,
+  useEChartOptions,
+} from '../../hooks/useECharts';
 import { ChartOptions } from './types';
 import { useVisualizedDataStreams } from './hooks/useVisualizedDataStreams';
-import { convertOptions } from './converters/convertOptions';
-import { ElementEvent, SeriesOption, YAXisComponentOption } from 'echarts';
-import { convertYAxis } from './converters/convertAxis';
-import { convertSeriesAndYAxis, reduceSeriesAndYAxis } from './converters/convertSeriesAndYAxis';
+import { useConvertedOptions } from './converters/convertOptions';
+import { ElementEvent } from 'echarts';
+import { useSeriesAndYAxis } from './converters/convertSeriesAndYAxis';
 import { HotKeys, KeyMap } from 'react-hotkeys';
 import useTrendCursors from './hooks/useTrendCursors';
-import { calculateYMaxMin } from './utils/getInfo';
-import { Resizable, ResizeCallbackData } from 'react-resizable';
+import { Resizable } from 'react-resizable';
 import Legend from './legend/legend';
-import { CHART_RESIZE_INITIAL_FACTOR, CHART_RESIZE_MAX_FACTOR, CHART_RESIZE_MIN_FACTOR } from './eChartsConstants';
-import { v4 as uuid } from 'uuid';
-import './chart.css';
+import { useChartStyleSettings } from './converters/convertStyles';
 import ChartContextMenu, { Action } from './contextMenu/ChartContextMenu';
+import { useChartId } from './hooks/useChartId';
+import { useChartSetOptionSettings } from './hooks/useChartSetOptionSettings';
+
+import './chart.css';
 
 const keyMap: KeyMap = {
   commandDown: { sequence: 'command', action: 'keydown' },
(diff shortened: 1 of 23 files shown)


packages/react-components/src/components/chart/hooks/useChartSetOptionSettings.ts
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * if the datastreams change update echarts using the replaceMerge stratgey
+     * so that it is ensured that orphaned data points are removed.
+     *
+     * see setOption api for more information on settings
+     * https://echarts.apache.org/en/api.html#echartsInstance.setOption
+     *
+     * NOTE: we can use this hook to refactor trend cursor set option also
+     */
+    const settings = datastreamsDepsRef.current !== datastreamsDeps ? { replaceMerge: ['series'] } : undefined;
```
mnischay, line 21: This should be only when datastreamsRef has streams which datastreamsDeps does not right ?

packages/react-components/src/components/chart/tests/handleSync.spec.tsx
```diff
@@ -21,13 +21,14 @@ describe('handleSync', () => {
...
 
   const ref = renderHook(() =>
-    useECharts({
-      option: {},
-      loading: false,
-      size: mockSize,
-      theme: 'dark',
-      groupId: 'group1',
-    })
+    // useECharts({
```
mnischay, line 24: you can probably remove this

packages/react-components/src/components/chart/types.ts
```diff
@@ -93,7 +93,7 @@ export type ChartOptions = {
   fontSettings?: SimpleFontSettings;
   legend?: ChartLegend;
   significantDigits?: number;
-  graphic?: InternalGraphicComponentGroupOption[];
+  graphic?: InternalGraphicComponentGroupOption[]; // This needs to be removed from the public api
```
mnischay, line 96: it should be renamed not removed, this is to allow users to add a TC by default

packages/react-components/src/components/chart/converters/convertStyles.spec.tsx
```diff
@@ -0,0 +1,56 @@
+// const { result } = renderHook(() => useGridSettings(), {
```
mnischay, line 1: you can probably remove this

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -1,88 +1,88 @@
...
+  const { width, height, chartWidth, onResize, minConstraints, maxConstraints } = useResizeableEChart(chartRef, size);
+
+  // apply group to echart
+  useGroupableEChart(chartRef, options.groupId);
 
-  const { series, yAxis } = useMemo(() => {
-    const { series, yAxis } = dataStreams
-      .map(convertSeriesAndYAxis(options as ChartOptions))
-      .reduce(reduceSeriesAndYAxis, { series: defaultSeries, yAxis: defaultYAxis });
-    const { yMax, yMin } = calculateYMaxMin(series);
```
mnischay, line 39: you probably forgot to remove this function definition in the getInfo file

packages/react-components/src/hooks/useECharts/useEChartOptions.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param option - set options on an echarts instance
+ * @param settings - settings to be used when setting an option
+ * @returns void
+ */
+export const useEChartOptions = (
+  chartRef: React.MutableRefObject<ECharts | null>,
+  option: EChartsOption,
+  settings?: SetOptionOpts
+) => {
+  useEffect(() => {
```
mnischay, line 18: curious: do you need the useEffect here ? no matter if any of these changes we need to execute 
```chartRef.current?.setOption(option, settings);
```

packages/react-components/src/components/chart/hooks/useChartSetOptionSettings.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  return useMemo(() => {
+    /**
+     * if the datastreams change update echarts using the replaceMerge stratgey
+     * so that it is ensured that orphaned data points are removed.
+     *
+     * see setOption api for more information on settings
+     * https://echarts.apache.org/en/api.html#echartsInstance.setOption
+     *
+     * NOTE: we can use this hook to refactor trend cursor set option also
```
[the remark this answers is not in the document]
jmbuss: Maybe settings should not be in 1 place, I think even this hook might not be exactly right. but I think we can use a similar strategy in TCs so that all of our hooks only return state instead of calling setOptions internally. I think that would be a cleaner pattern.

packages/react-components/src/components/chart/types.ts
jmbuss: When we were deciding on the API we had said this was not going to be an option? I thought this was only going to be an on the fly feature?

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -1,88 +1,88 @@
...
-  });
-
   const handleContextMenu = (e: ElementEvent) => {
     setContextMenuPos({ x: e.offsetX, y: e.offsetY });
     setShowContextMenu(!showContextMenu);
     e.stop();
   };
 
+  // adapt datastreams into echarts series and yAxis data
+  const { series, yAxis } = useSeriesAndYAxis(dataStreams, { styleSettings: styleSettingsMap, axis: options.axis });
```
mnischay, line 79: we should also separate yAxis and series, looks like they are tightly coupled.

packages/react-components/src/components/chart/hooks/useChartSetOptionSettings.ts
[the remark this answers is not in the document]
mnischay: only return state instead of calling setOptions internally --> thats how TC pattern is , only exception is the delete TC. Again, features should be calling setOptions as an exception, not everytime, which is also the pattern in TC

packages/react-components/src/components/chart/types.ts
mnischay: The final API is obviously decide, but that the intention. not sure what fly feature means, but the intention was that if they want it, they can add.. we may or may not decide to keep this at the time of launch.

PR #1784, opened 2023-08-09
feat(composer): auto query support for Model Shader

sheilaXu:
## Overview
support auto query feature for model shader component

https://github.com/awslabs/iot-app-kit/assets/9315598/09abd434-4035-4768-a65e-5ba08c61474a

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
@@ -17,7 +17,6 @@ import useTagSettings from '../../../../hooks/useTagSettings';
 import {
   DefaultAnchorStatus,
   INavLink,
-  IRuleBasedMap,
   ITagSettings,
   IValueDataBinding,
   KnownComponentType,
@@ -26,21 +25,20 @@ import {
 } from '../../../../interfaces';
 import { IAnchorComponentInternal, ISceneNodeInternal, useStore, useViewOptionState } from '../../../../store';
 import { applyDataBindingTemplate } from '../../../../utils/dataBindingTemplateUtils';
-import { dataBindingValuesProvider, ruleEvaluator } from '../../../../utils/dataBindingUtils';
 import { findComponentByType } from '../../../../utils/nodeUtils';
 import { getSceneResourceInfo } from '../../../../utils/sceneResourceUtils';
 import { colors } from '../../../../utils/styleUtils';
 import { Anchor } from '../../../three';
 import { WidgetSprite, WidgetVisual } from '../../../three/visuals';
 import svgIconToWidgetSprite from '../common/SvgIconToWidgetSprite';
-import useBindingData from '../../../../hooks/useBindingData';
+import useRuleResult from '../../../../hooks/useRuleResult';
 
 export interface AnchorWidgetProps {
   node: ISceneNodeInternal;
   chosenColor?: string;
   defaultIcon: string;
   valueDataBinding?: IValueDataBinding;
-  rule?: IRuleBasedMap;
+  ruleBasedMapId?: string;
   navLink?: INavLink;
 }
 
@@ -52,7 +50,7 @@ export function AsyncLoadedAnchorWidget({
   defaultIcon,
   chosenColor,
   valueDataBinding,
-  rule,
+  ruleBasedMapId,
   navLink,
 }: AnchorWidgetProps): ReactElement {
   const sceneComposerId = useContext(sceneComposerIdContext);
@@ -69,8 +67,11 @@ export function AsyncLoadedAnchorWidget({
   const autoRescale = useMemo(() => {
     return tagSettings.autoRescale;
   }, [tagSettings.autoRescale]);
-  const bindings = useMemo(() => (valueDataBinding ? [valueDataBinding] : undefined), [valueDataBinding]);
-  const bindingData = useBindingData(bindings).data?.at(0);
+  const ruleResult = useRuleResult({
+    ruleMapId: ruleBasedMapId,
+    dataBinding: valueDataBinding,
+    defaultValue: defaultIcon,
+  });
 
   const tagVisible = useViewOptionState(sceneComposerId).componentVisibilities[KnownComponentType.Tag];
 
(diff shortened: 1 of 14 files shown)


packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
```diff
@@ -69,8 +67,11 @@ export function AsyncLoadedAnchorWidget({
   const autoRescale = useMemo(() => {
     return tagSettings.autoRescale;
   }, [tagSettings.autoRescale]);
-  const bindings = useMemo(() => (valueDataBinding ? [valueDataBinding] : undefined), [valueDataBinding]);
-  const bindingData = useBindingData(bindings).data?.at(0);
+  const ruleResult = useRuleResult({
```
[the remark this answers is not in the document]
sheilaXu: we don't plan to add it to release-3.x currently

packages/scene-composer/src/components/three-fiber/AnchorComponent/index.tsx
```diff
@@ -12,15 +11,13 @@ interface IAnchorComponentProps {
...
-  const sceneComposerId = useSceneComposerId();
-  const rule = useStore(sceneComposerId)((state) => state.getSceneRuleMapById(component.ruleBasedMapId));
   return (
     <group name={getComponentGroupName(node.ref, 'TAG')}>
       <AnchorWidget
         node={node}
         defaultIcon={component.icon ?? DefaultAnchorStatus.Info}
         navLink={component.navLink}
-        rule={rule}
+        ruleBasedMapId={component.ruleBasedMapId}
```
divya-sea, line 20: What's the reason for not fetching it from the store? Are we not storing these maps in the store?

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
```diff
@@ -35,7 +36,7 @@ export const DataOverlayDataRow = ({
       if (!binding.valueDataBinding) {
         return;
       }
-      const values: Record<string, unknown> = dataBindingValuesProvider(
+      const values: Record<string, Primitive> = dataBindingValuesProvider(
```
divya-sea, line 39: What does Primitive supports?

packages/scene-composer/src/hooks/useBindingData.ts
```diff
@@ -38,39 +38,41 @@ const useBindingData = (
...
+        },
+      });
 
-            if (results.length > 1 || results[0].dataStreams.length > 1) {
-              log?.warn('Multiple data returned, only use the first one');
-            }
+      provider.subscribe({
+        // TODO: support static data
+        next: (results: TimeSeriesData[]) => {
+          if (isEmpty(results.at(0)?.dataStreams)) {
```
divya-sea, line 56: You can directly check results[0].dataStreams?

packages/scene-composer/src/hooks/useRuleResult.ts
```diff
@@ -1,20 +1,34 @@
...
 
-const useRuleResult = (ruleMapId: string, dataBinding: IValueDataBinding) => {
+import useBindingData from './useBindingData';
+
+const useRuleResult = ({
+  ruleMapId,
+  dataBinding,
+  defaultValue,
+}: {
+  ruleMapId?: string;
```
divya-sea, line 16: you can make them as type safe by having them in an interface.

packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
```diff
@@ -103,13 +104,10 @@ export function AsyncLoadedAnchorWidget({
...
   const visualState = useMemo(() => {
-    const values: Record<string, unknown> =
-      bindingData ?? dataBindingValuesProvider(dataInput, valueDataBinding, dataBindingTemplate);
-    const ruleTarget = ruleEvaluator(defaultIcon, values, rule);
-    const ruleTargetInfo = getSceneResourceInfo(ruleTarget as string);
+    const ruleTargetInfo = getSceneResourceInfo(ruleResult as string);
     // Anchor widget only accepts icon, otherwise, default to Info icon
     return ruleTargetInfo.type === SceneResourceType.Icon ? ruleTargetInfo.value : defaultIcon;
-  }, [rule, dataInput, valueDataBinding, defaultIcon, bindingData]);
+  }, [ruleResult]);
```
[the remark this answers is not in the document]
sheilaXu: visualState is determined by ruleResult, other fields are used to generate the ruleResult

packages/scene-composer/src/components/three-fiber/AnchorComponent/index.tsx
sheilaXu: the `useRuleResult` now used by AnchorWidget will fetch the rule by id

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
sheilaXu: number, string, boolean

divya-sea: So we don't want to support the other primitive data types?

packages/scene-composer/src/hooks/useBindingData.ts
sheilaXu: both results and results[0].dataStreams can be empty array, so I need to check bth. if you are talking about [0] vs at(0), they should be the same?

packages/scene-composer/src/hooks/useRuleResult.ts
sheilaXu: it should be the same to define type inline VS named interface?

packages/scene-composer/src/hooks/useBindingData.ts
divya-sea: Yeah bit confusing as you used some place results.at(0) and in some results[0] both.

packages/scene-composer/src/hooks/useRuleResult.ts
divya-sea: It is more readable and can be reused if you have an interface.

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
sheilaXu: this change is just to make `unknown` a more restricted type. the actually logic to handle the value is unchanged. I believe we cannot handle other data type in anyway. so this won't change the actual functionality

packages/scene-composer/src/hooks/useBindingData.ts
sheilaXu: using at(0) force me to handle the undefined case by throw compiler error, but `results[0].dataStreams` will not throw error when building, but will throw runtime error when results is empty. for the code below, since I've checked they are not empty here, I can just access [0] and impossible to have undefined

packages/scene-composer/src/components/three-fiber/DataOverlayComponent/DataOverlayDataRow.tsx
divya-sea: That means we are assuming here type of a value is known at the compilation time when receiving the data from the an external source. Something to watch for if we haven't tested this scenario.

divya-sea (approved):

PR #1786, opened 2023-08-10
feat(react-components): supporting live mode in echarts

mnischay:
## Overview
This is to add live mode. i.e the charts obey the given viewport

## Verifying Changes
refer to the screen share

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/aef5a294-43e0-418b-a7d4-cb550be0c235

### DIFF
--- a/packages/react-components/src/components/chart/baseChart.tsx
+++ b/packages/react-components/src/components/chart/baseChart.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import {
   useECharts,
   useResizeableEChart,
@@ -9,9 +9,8 @@ import {
 import { ChartOptions } from './types';
 import { useVisualizedDataStreams } from './hooks/useVisualizedDataStreams';
 import { useConvertedOptions } from './converters/convertOptions';
-import { ElementEvent } from 'echarts';
 import { useSeriesAndYAxis } from './converters/convertSeriesAndYAxis';
-import { HotKeys, KeyMap } from 'react-hotkeys';
+import { HotKeys } from 'react-hotkeys';
 import useTrendCursors from './hooks/useTrendCursors';
 import { Resizable } from 'react-resizable';
 import Legend from './legend/legend';
@@ -21,11 +20,9 @@ import { useChartId } from './hooks/useChartId';
 import { useChartSetOptionSettings } from './hooks/useChartSetOptionSettings';
 
 import './chart.css';
-
-const keyMap: KeyMap = {
-  commandDown: { sequence: 'command', action: 'keydown' },
-  commandUp: { sequence: 'command', action: 'keyup' },
-};
+import { useXAxis } from './hooks/useXAxis';
+import { useContextMenu } from './hooks/useContextMenu';
+import { useViewportToMS } from './hooks/useViewportToMS';
 
 /**
  * Developer Notes:
@@ -36,7 +33,10 @@ const keyMap: KeyMap = {
  * 2. get datastreams using useVisualizedDataStreams
  * 3. use datastreams / chart options to compute datastructures
  *    needed to implement various features and adapt them to the echarts api
- * 4. set all of the options in echarts using useEChartOptions
+ * 4. set all  the options in echarts using useEChartOptions
+ * 5. do not make use of setOptions in the individual feature, useEChartOptions should be the only place.
+ *    Exception: when deleting an item or in general when removing some elements you may need to use setOptions
+ *
  */
 
 /**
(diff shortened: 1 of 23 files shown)


mnischay:
The failing test is from scene-composer

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -36,7 +33,10 @@ const keyMap: KeyMap = {
  * 2. get datastreams using useVisualizedDataStreams
  * 3. use datastreams / chart options to compute datastructures
  *    needed to implement various features and adapt them to the echarts api
- * 4. set all of the options in echarts using useEChartOptions
+ * 4. set all  the options in echarts using useEChartOptions
+ * 5. do not make use of setOptions in the individual feature, useEChartOptions should be the only place.
+ *    Exception: when deleting an item or in general when removing some elements you may need to use setOptions
```
jmbuss, line 38: We don't need to use this exception if we correctly set the settings for setOption. I think we should try not to use the echarts ref to do this outside of useEChartOptions

mnischay: right, thats why its an exception :)

packages/react-components/src/components/chart/hooks/useViewportToMS.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  const viewportString = JSON.stringify(viewport);
+  useEffect(() => {
+    let interval: NodeJS.Timer;
+    if (viewport && isDurationViewport(viewport)) {
+      interval = setInterval(() => {
+        setInMS(convertViewportToMs(viewport));
+      }, LIVE_MODE_REFRESH_RATE_MS);
+    } else {
+      setInMS(convertViewportToMs(viewport));
```
jmbuss, line 17: what does milliseconds represent in an absolute viewport? the difference between the start and end?

packages/react-components/src/components/chart/hooks/useXAxis.ts
```diff
@@ -0,0 +1,18 @@
...
+import { XAXisOption } from 'echarts/types/dist/shared';
+import { useMemo } from 'react';
+
+export const useXAxis = (viewportInMs: ViewportInMs, axis?: ChartAxisOptions): XAXisOption => {
+  return useMemo(
+    () => ({
+      show: axis?.showX ?? DEFAULT_X_AXIS.show,
+      type: 'time' as const,
+      splitNumber: 6,
+      interval: 1000 * 60, // 1sec interval
```
jmbuss, line 12: does this number work for all relative durations? Why does it have to be 1 minute?

packages/react-components/src/components/chart/types.ts
```diff
@@ -93,17 +93,24 @@ export type ChartOptions = {
...
+  graphic?: InternalGraphicComponentGroupOption[];
   theme?: string;
   groupId?: string;
   id?: string;
 };
 
+export interface ViewportInMs {
+  end: number;
+  initial: number;
+  widthInMs: number;
```
jmbuss, line 105: Can we move something like this to a util function? I feel like width in milliseconds doesn't really make sense? Assuming this is useful in trendcursors for knowing where on the xaxis to place something?

packages/react-components/src/components/chart/hooks/useViewportToMS.ts
```diff
@@ -0,0 +1,28 @@
...
+export const useViewportToMS = (viewport?: Viewport) => {
+  const [inMS, setInMS] = useState(convertViewportToMs(viewport));
+
+  const viewportString = JSON.stringify(viewport);
+  useEffect(() => {
+    let interval: NodeJS.Timer;
+    if (viewport && isDurationViewport(viewport)) {
+      interval = setInterval(() => {
+        setInMS(convertViewportToMs(viewport));
+      }, LIVE_MODE_REFRESH_RATE_MS);
```
jmbuss, line 15: refresh rate is configurable on the time series data hook, I think we should make those values are in sync since we will want to make that configurable through the viewport selector in the future.

jmbuss (changes requested):

packages/react-components/src/components/chart/hooks/useViewportToMS.ts
mnischay: just converted to ms, initial, end and delta between  them.. all in ms

packages/react-components/src/components/chart/hooks/useXAxis.ts
mnischay: good catch - i was playing around this value, I dont think we should be setting this and let echarts scale it however it does , will remove

packages/react-components/src/components/chart/types.ts
mnischay: the invoking function need not use this, I believe end and initial are probably more used and this is available if needed .. its a simple math, so dont think it will hog up too much processing cycle anyway..

packages/react-components/src/components/chart/hooks/useViewportToMS.ts
mnischay: this is the chart refresh rate which is different from the api fetch frequency right ? 
the api frequency is 5s i think, and this will stay at 1sec is what i gathered from @tracy-french .. not sure, if we want to expose this as a configurable value...

mnischay: With the 5sec delay, for 10sec  viewport, the movement is not smooth, so needs to be 1000ms.

tracy-french (approved):

PR #1787, opened 2023-08-10
feat(react-components): add multiple y axis legend

jmbuss:
## Overview
Adds multiple y axis legend
Adds datastore for highlighting datatreams

## Legal


packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
```diff
@@ -1,11 +1,26 @@
...
+import maxBy from 'lodash.maxby';
+import minBy from 'lodash.minby';
+
 import { ChartAxisOptions, ChartStyleSettingsOptions } from '../types';
 import { convertDataPoint } from './convertDataPoint';
 import { StyleSettingsMap, getChartStyleSettingsFromMap } from './convertStyles';
 import { convertYAxis as convertChartYAxis } from './convertAxis';
+import { ChartStyleSettingsWithDefaults } from '../utils/getStyles';
+
+const EMPHASIZE_SCALE_CONSTANT = 2;
```
mnischay, line 14: we can move all the constants are in the echarts constants file ?

packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
```diff
@@ -1,11 +1,26 @@
...
+const DEEMPHASIZE_OPACITY = 0.25;
+
+export type YAxisLegendOption = { datastream: DataStream; value: DataPoint; color: string; significantDigits: number };
```
mnischay, line 17: we can move all the types to the types.ts files ?  all the types are moved/created in types.ts in general

packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
```diff
@@ -77,13 +114,23 @@ const convertYAxis = ({ color, yAxis }: ChartStyleSettingsOptions): YAXisCompone
...
+export const convertSeriesAndYAxis = (styles: ChartStyleSettingsWithDefaults) => (datastream: DataStream) => {
+  const series = convertSeries(datastream, styles);
   const yAxis = convertYAxis(styles);
 
+  const { color, significantDigits } = styles;
+  const { data } = datastream;
+
+  const toYAxisLegend = yAxisLegendGenerator({ datastream, color, significantDigits });
+
+  const yMax = maxBy(data, dataValue);
```
mnischay, line 126: not sure if this is the same yMax we were using for TC , we do not need yMin and yMax anymore

packages/react-components/src/components/chart/converters/convertStyles.ts
```diff
@@ -37,6 +46,9 @@ export const getChartStyleSettingsFromMap =
  * @returns [StyleSettingsMap, (datastream: DataStream) => ChartStyleSettingsOptions]
  */
 export const useChartStyleSettings = (datastreams: DataStream[], chartOptions: ConvertChartOptions) => {
+  const highlightedDataStreams = useDataStore((state) => state.highlightedDataStreams);
+  const isDataStreamHighlighted = useDataStore((state) => state.isDataStreamHighlighted);
```
mnischay, line 50: usually, we do not use store to process data, like I am guessing this is not a action , this is a post-process on an action, so this may be better suited to be utility function rather than store action ?

packages/react-components/src/components/chart/eChartsConstants.ts
```diff
@@ -80,3 +80,5 @@ export const TREND_CURSOR_LINE_MARKERS_GRAPHIC_INDEX = 3;
 export const CHART_RESIZE_INITIAL_FACTOR = 0.75;
 export const CHART_RESIZE_MIN_FACTOR = 0.3;
 export const CHART_RESIZE_MAX_FACTOR = 0.75;
+
+export const MULTI_Y_AXIS_LEGEND_WIDTH = 172;
```
mnischay, line 84: curious : is this a constant , i.e do we want it to always occupy a fixed width ?

packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
```diff
@@ -4,28 +4,66 @@ import { TableProps } from '@cloudscape-design/components/table/interfaces';
...
   series: SeriesOption[];
 }) => {
   const legendColumnDefinition = {
     id: 'Legends',
     header: <div className='base-chart-legend-col-header'>Properties</div>,
-    cell: (e: { lineColor: string; name: string }) => (
-      <div className='base-chart-legend-row-data-container'>
+    cell: (e: { datastream: DataStream; lineColor: string; name: string }) => {
+      const { datastream, lineColor, name } = e;
+      const highlightDataStream = useDataStore((state) => state.highlightDataStream);
```
mnischay, line 28: curious: why do we want to store the inner workings of a component in zustand/store ? this data is always used and referred within the base component.. so we could just use state ? unless this is shared across multiple basecharts..

packages/react-components/src/components/chart/legend/colorIcon.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export const ColorIcon = ({ color }: { color: string }) => (
+  <div style={{ height: '100%', display: 'flex', alignItems: 'center' }}>
+    <div
+      style={{
+        height: '4px',
```
mnischay, line 7: worth using the design tokens ? if not, may be we could move to a .css file ?

packages/react-components/src/store/highlightedDataStreamsSlice.ts
```diff
@@ -0,0 +1,28 @@
...
+  highlightDataStream: (datastream?: DataStream) => void;
+  unHighlightDataStream: (datastream?: DataStream) => void;
+  isDataStreamHighlighted: (datastream?: DataStream) => boolean;
+}
+
+export const createHighlightedDataStreamsSlice: StateCreator<HighlightedDataSteamsState> = (set, get) => ({
+  highlightedDataStreams: [],
+  highlightDataStream: (datastream?: DataStream) =>
+    set((state) => {
+      if (!datastream) return { highlightedDataStreams: state.highlightedDataStreams };
```
mnischay, line 18: you can just update only when if state is true ? OR may be the action is not even called if the datastreams is undefined ? 
i.e make `highlightDataStream: (datastream: DataStream)`

packages/react-components/src/components/chart/store/store.ts
```diff
@@ -0,0 +1,11 @@
+import { create } from 'zustand';
+import { devtools } from 'zustand/middleware';
+import { createHighlightedDataStreamsSlice, HighlightedDataSteamsState } from './highlightedDataStreams';
+
+export type StateData = HighlightedDataSteamsState;
+export const createChartStore = () =>
+  create<StateData>()(
+    devtools((...args) => ({
```
mnischay, line 8: do you know how this devtools will work with the global store devtools ?

packages/react-components/src/components/chart/store/highlightedDataStreams.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface HighlightedDataSteamsState extends HighlightedDataStreamsData {
+  highlightDataStream: (datastream?: DataStream) => void;
+  unHighlightDataStream: (datastream?: DataStream) => void;
+}
+
+export const createHighlightedDataStreamsSlice: StateCreator<HighlightedDataSteamsState> = (set) => ({
+  highlightedDataStreams: [],
+  highlightDataStream: (datastream?: DataStream) =>
+    set((state) => {
+      if (!datastream) return { highlightedDataStreams: state.highlightedDataStreams };
```
mnischay, line 17: this should be if ```(!datastream) return state ;``` right ?

packages/react-components/src/components/chart/converters/convertStyles.ts
```diff
@@ -61,11 +74,16 @@ export const useChartStyleSettings = (datastreams: DataStream[], chartOptions: C
   }, [datastreamDeps]);
 
   return useMemo(() => {
+    // Use to see if we should emphasize / de-emphasize trends
+    const shouldUseEmphasis = highlightedDataStreams.length > 0;
+
     const map = datastreams.reduce<StyleSettingsMap>((styleMap, datastream) => {
```
mnischay, line 80: the reduce function arent most easy to read, mind adding some comments?

packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
```diff
@@ -4,28 +4,68 @@ import { TableProps } from '@cloudscape-design/components/table/interfaces';
...
+      const highlightDataStream = useChartStore((state) => state.highlightDataStream);
+      const unHighlightDataStream = useChartStore((state) => state.unHighlightDataStream);
+      const highlightedDataStreams = useChartStore((state) => state.highlightedDataStreams);
+      const isDataStreamHighlighted = isDataStreamInList(highlightedDataStreams);
+
+      const toggleHighlighted = () => {
+        if (isDataStreamHighlighted(datastream)) {
+          unHighlightDataStream(datastream);
+        } else {
+          highlightDataStream(datastream);
```
mnischay, line 38: the input is ```datastream: DataStream``` so its always defined, in that case , the ignature for ```highlightDataStream``` can be ```highlightDataStream(datastream" DataStream);``` right ? i.e we can avoid if else is zustand state

packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
```diff
@@ -4,28 +4,68 @@ import { TableProps } from '@cloudscape-design/components/table/interfaces';
...
+          onClick={toggleHighlighted}
+        >
+          <div
+            className={`base-chart-legend-row-line-ind ${
+              isDataStreamHighlighted(datastream) ? 'base-chart-legend-row-line-ind-highlighted' : ''
+            }`}
+            style={{
+              backgroundColor: lineColor,
+            }}
+          ></div>
```
mnischay, line 58: />

packages/react-components/src/components/chart/store/useChartStore.ts
```diff
@@ -0,0 +1,6 @@
+import { useContext } from 'react';
+// import { useStore } from 'zustand';
```
mnischay, line 2: remove ?

packages/react-components/src/components/chart/utils/getStyles.ts
```diff
@@ -19,6 +29,7 @@ export const getDefaultStyles = (
     lineStyle: 'solid',
     lineThickness: 2,
     yAxis: undefined,
-    significantDigits: 4,
+    significantDigits: significantDigits ?? 4,
```
mnischay, line 32: probably worth using same constant across ?there is already Y_AXIS_INTERPOLATED_VALUE_PRECISION. you can update the name and just have one default precision.

mnischay: Also, just for my reference... why 4 ?

mnischay (changes requested):
just some super minor changes requested, Also, mind attaching a screen capture/shot ?

packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
```diff
@@ -94,7 +137,12 @@ const addYAxisIndex = <T extends SeriesOption>(series: T, yAxisIndex = 0): T =>
 
 type SeriesAndYAxis = ReturnType<ReturnType<typeof convertSeriesAndYAxis>>;
 
-type ReducedSeriesAndYAxis = { series: SeriesAndYAxis['series'][]; yAxis: NonNullable<SeriesAndYAxis['yAxis']>[] };
+type ReducedSeriesAndYAxis = {
+  series: SeriesAndYAxis['series'][];
+  yAxis: NonNullable<SeriesAndYAxis['yAxis']>[];
+  yMins: NonNullable<SeriesAndYAxis['yMin']>[];
+  yMaxs: NonNullable<SeriesAndYAxis['yMax']>[];
```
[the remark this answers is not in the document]
jmbuss: no, we need extra data mapped for the feature to work correctly.

packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
```diff
@@ -39,33 +50,55 @@ const convertStep = (
...
   yAxis && {
-    show: false,
+    /**
+     * showing the axis only to ensure that the horizontal
+     * mark lines are visible
+     *
+     * axis label refers to the numbers at each mark line
+     */
+    show: true,
+    axisLabel: { show: false },
```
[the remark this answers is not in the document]
jmbuss: no, it's not possible to show numbers in echarts for multiple y-axis. We don't know how many custom axis will be defined and would not have the space to show a unique axis for each. This is why we have the legend menus

packages/react-components/src/components/chart/store/highlightedDataStreams.ts
jmbuss: Yes, this is the same thing. But you're right, probably better to change it to just be state

packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
jmbuss: this isn't the only place we are using highlight / unhighlight. Other locations have an undefined.

packages/react-components/src/components/chart/utils/getStyles.ts
jmbuss: also, user's can set the precision they want at a chart level and property level. So this value is dynamic, but 4 is the default if unset.

mnischay: right, lets just create a const for this fallback precision value, so we can just use a constant everywhere

mnischay (approved):
Looks good to me, but there are merge conflicts

PR #1790, opened 2023-08-10
fix(Tag style): custom color to support icon rules

divya-sea:
## Overview
This change contains the fix to rule with custom icon color.
1. Verify tag color changes if rule meets the criteria.
2. Verify tag color does not change if rule does not meet criteria.
3. Verify when rule does not meet, it shows default custom color for tags.
* unit tests

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/54058998/ccd834c7-a22b-4249-b842-0325607cdcf7

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
@@ -42,6 +42,8 @@ export interface AnchorWidgetProps {
   navLink?: INavLink;
 }
 
+type overrideCustomColorType = (value: string | undefined) => void;
+
 // Adds the custom objects to React Three Fiber
 extend({ Anchor, WidgetVisual, WidgetSprite });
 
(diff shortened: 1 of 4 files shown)


packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
```diff
@@ -104,11 +106,13 @@ export function AsyncLoadedAnchorWidget({
 
   // Evaluate visual state based on data binding
   const visualState = useMemo(() => {
```
haweston, line 108: Is this variable only consumed in testing? It seems like it's also only the icon state now as it has no color information in the actual return value.

divya-sea: If I understand correctly your comment, you are asking about custom color info for the icon.  If it is Info then it will be iottwinmaker-common-icon:Info. For custom color it will be iottwinmaker-common-icon:Custom-#ffffff. Color information will be part of the targetIcon value.

haweston: Ah, okay. Wasn't clear from just scanning variable names. That was one part.

haweston: Other question is do we use the const visualState anywhere?

divya-sea: Yes we pass this visualState part of anchor widget prop.

haweston: Okay, missed expanding one code section, found it.

haweston (approved):

PR #1791, opened 2023-08-11
fix(3DKG): fix height flexibility for KG component

MO-Elmu:
## Overview
This change will allow height flexibility if app developer used flexible container. The height of the KG component will expand and shrink according to the container resize. Width flexibility was working fine, it turns polaris container doesn't offer height flexibility and the fitHeight property is not quite what i wanted.
### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


examples/react-app/src/components/KnowledgeGraph.scss
```diff
@@ -0,0 +1,6 @@
+.KnowledgeGraph {
+  min-height: 680px;
```
mumanity, line 2: Suggest converting these sizes to use `rem` so that they're properly responsive. `px` tends to be monitor-specific.

mumanity: Could also implement `vh` & `vw` if you want the size based on the browser window.

packages/react-components/src/components/knowledge-graph/styles.scss
```diff
@@ -0,0 +1,16 @@
...
+  display: flex;
+  flex-direction: column;
+}
+.search-bar {
+  min-width: 500px;
+  flex: 0;
+  margin: 10px;
+}
+.action-buttons {
+  min-width: 300px;
```
mumanity, line 13: This should use `rem`.

mumanity (changes requested):

examples/react-app/src/components/KnowledgeGraph.scss
MO-Elmu: Two things I would like to clarify, This change has nothing to do with the PR purpose just while testing my change using the react example I realized the KG component in the example is styles with a class that didn't exist so I added the class which has the same CSS as the video player and the Scene viewer both use px.
The other thing is I can't change this in isolation if the rest of the components are using PX this will mess up the whole app if the user change the default browser size from 16 to any value the knowledge Graph will be the only component that will adjust coz its using rem.

I don't disagree with rem but it should be part of a comprehensive evaluation and migration (or using both rem & px) to all the components in the App so we don't end up with inconsistent behavior. This argument actually apply to the below comments as well.

packages/react-components/src/components/knowledge-graph/styles.scss
MO-Elmu: same here, this is a small div/component in the KG UI the other div like canvas search bar .. etc all use px adjusting this alone means as the default value of 16px grow or shrink this is the only div that will grow and shrink while the other stay the same (coz the are using px). by the way this one is also not part of the main purpose of the PR they were inline styling just moved them to stylesheet file. I am not opposed to rem but I think partial rem usage while many other inter related div use px is dangerous. I think we should create an asana task so this will be a comprehensive task that move everything to rem (or use both) with extensive testing. And if i go down this rabbit hole :) the scope of this PR will be very big. I can see the change will be straight forward if its a font-size styling but I am bit reluctant to do this  partially for width, height ..etc.

Great discussion [here ](https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css)about using rem vs px, but in general I am not opposed to using rem but it should be part of comprehensive migration task with careful testing.

PR #1794, opened 2023-08-13
feat(dashboard): line symbol

shlokj:
## Overview
This PR is for the line symbol widget. It allows the user to drag and drop a line. The line can be manipulated using the anchors at its ends.

Currently, the line is limited to its bounding box: the bouding box has to be resized to allow a larger line. This is not the final state - in a future PR, the bounding box will be invisible except for during multiple selection and the line will be able to be placed anywhere regardless of it.

## Verifying Changes

Refer to the screen recording below:

https://github.com/awslabs/iot-app-kit/assets/34567765/3d885d29-2d1b-4168-af0b-52e529949029

## Legal

### DIFF
--- a/packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
+++ b/packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
@@ -1,0 +1,137 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { DndProvider } from 'react-dnd';
+import { HTML5Backend } from 'react-dnd-html5-backend';
+import LineWidgetComponent from './component';
+import { Provider } from 'react-redux';
+import { configureDashboardStore } from '~/store';
+import { DashboardState } from '~/store/state';
+import { LineProperties, LineWidget } from '../types';
+
+const mockWidgetPartial = {
+  id: 'mock-line-widget',
+  type: 'line-symbol',
+  x: 0,
+  y: 0,
+  z: 1,
+  width: 7,
+  height: 4,
+};
+
+const TestComponent = (object: { widget: LineWidget; isSelected: boolean }) => {
+  const { widget, isSelected } = object;
+
+  const initialState: Partial<DashboardState> = {
+    dashboardConfiguration: {
+      widgets: [widget],
+      viewport: { duration: '5m' },
+    },
+    selectedWidgets: isSelected ? [widget] : [],
+  };
+
+  return (
+    <Provider store={configureDashboardStore(initialState)}>
+      <DndProvider backend={HTML5Backend}>
+        <LineWidgetComponent {...widget} />
+      </DndProvider>
+    </Provider>
+  );
+};
+
+describe('LineWidgetComponent', () => {
+  describe('Rendering', () => {
+    [
+      {
+        properties: {
+          lineStyle: 'solid',
+          color: 'black',
+          thickness: 5,
+          start: { x: 20, y: 20 },
+          end: { x: 180, y: 180 },
+        } as LineProperties,
+        isSelected: true,
+      },
+      {
+        properties: {
+          lineStyle: 'dashed',
+          color: 'black',
+          thickness: 5,
+          start: { x: 30, y: 20 },
+          end: { x: 180, y: 190 },
+        } as LineProperties,
+        isSelected: true,
+      },
+      {
+        properties: {
+          lineStyle: 'dotted',
+          color: 'black',
+          thickness: 5,
+          start: { x: 25, y: 5 },
+          end: { x: 100, y: 120 },
+        } as LineProperties,
+        isSelected: true,
+      },
+      {
+        properties: {
+          lineStyle: 'solid',
+          color: 'red',
+          thickness: 5,
+          start: { x: 20, y: 20 },
+          end: { x: 180, y: 180 },
+        } as LineProperties,
+        isSelected: true,
+      },
+      {
+        properties: {
+          lineStyle: 'dashed',
+          color: 'blue',
+          thickness: 10,
+          start: { x: 180, y: 175 },
+          end: { x: 20, y: 30 },
+        } as LineProperties,
+        isSelected: true,
+      },
+      {
+        properties: {
+          lineStyle: 'dotted',
+          color: 'green',
+          thickness: 15,
+          start: { x: 180, y: 175 },
+          end: { x: 20, y: 30 },
+        } as LineProperties,
+        isSelected: true,
+      },
+    ].forEach((configuration) => {
+      const widget = { ...mockWidgetPartial, properties: { ...configuration.properties } };
+      const isSelected = configuration.isSelected;
+      it(`should render correctly with ${JSON.stringify(configuration)} while ${
+        isSelected ? 'selected' : 'not selected'
+      }`, () => {
+        const { container } = render(<TestComponent widget={widget} isSelected={isSelected} />);
+        expect(container).toMatchSnapshot();
+      });
+    });
+  });
+
+  describe('Anchor visibility', () => {
+    const widget = {
+      ...mockWidgetPartial,
+      properties: {
+        lineStyle: 'solid',
+        color: 'black',
+        thickness: 5,
+        start: { x: 180, y: 175 },
+        end: { x: 20, y: 30 },
+      },
+    } as LineWidget;
+
+    it('should display anchors when the widget is selected', async () => {
+      const { getAllByTestId } = render(<TestComponent widget={widget} isSelected={true} />);
+      expect(getAllByTestId('line-anchor')).toHaveLength(2);
+    });
+    it('should not display anchors when the widget is not selected', () => {
+      const { queryAllByTestId } = render(<TestComponent widget={widget} isSelected={false} />);
+      expect(queryAllByTestId('line-anchor')).toHaveLength(0);
+    });
+  });
+});
--- a/packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
+++ b/packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
@@ -1,0 +1,103 @@
+import React, { CSSProperties } from 'react';
+import './component.css';
+import { LineWidget } from '~/customization/widgets/types';
+import { SVG_STROKE_DASHED, SVG_STROKE_DOTTED, SVG_STROKE_SOLID } from '../constants';
+import { LineAnchor } from './lineAnchor/component';
+import { useIsSelected } from '~/customization/hooks/useIsSelected';
+import { useWidgetActions } from '~/customization/hooks/useWidgetActions';
+import { DashboardWidget } from '~/types';
+import { XYCoord } from 'react-dnd';
+
+const ANCHOR_HALF_SIDE_LENGTH_PX = 10;
+
+const LineWidgetComponent: React.FC<LineWidget> = (widget) => {
+  const isSelected = useIsSelected(widget);
+
+  const {
+    lineStyle = 'solid',
+    color = 'black',
+    thickness = 5,
+    start = {
+      x: 25,
+      y: 200,
+    },
+    end = {
+      x: 375,
+      y: 200,
+    },
+  } = widget.properties;
+
+  let strokeString = SVG_STROKE_SOLID;
+
+  if (lineStyle === 'dashed') {
+    strokeString = SVG_STROKE_DASHED;
+  } else if (lineStyle === 'dotted') {
+    strokeString = SVG_STROKE_DOTTED;
+  }
+
+  // we need to subtract ANCHOR_HALF_SIDE_LENGTH_PX because otherwise the line ends match the anchors' top-left points
+  // whereas we want them to match their centers
+  const startAnchorStyle: CSSProperties = {
+    position: 'absolute',
+    left: `${start.x - ANCHOR_HALF_SIDE_LENGTH_PX}px`,
+    top: `${start.y - ANCHOR_HALF_SIDE_LENGTH_PX}px`,
+  };
+
+  const endAnchorStyle: CSSProperties = {
+    position: 'absolute',
+    left: `${end.x - ANCHOR_HALF_SIDE_LENGTH_PX}px`,
+    top: `${end.y - ANCHOR_HALF_SIDE_LENGTH_PX}px`,
+  };
+
+  const { update: updateWidget } = useWidgetActions<DashboardWidget>();
+
+  const updateAnchor = (anchorType: 'start' | 'end', newPoint: XYCoord) => {
+    const updatedProperties = {
+      ...widget.properties,
+      [anchorType]: {
+        x: newPoint.x,
+        y: newPoint.y,
+      },
+    };
+    const newWidget = { ...widget, properties: updatedProperties };
+    updateWidget(newWidget);
+  };
+
+  return (
+    <div className='fit-full'>
+      {isSelected && (
+        <LineAnchor
+          style={startAnchorStyle}
+          point={{ x: widget.properties.start.x, y: widget.properties.start.y }}
+          dimensions={{ height: widget.height, width: widget.width }}
+          updateWidget={(newPoint) => {
+            updateAnchor('start', newPoint);
+          }}
+        />
+      )}
+      <svg className='fit-full' pointerEvents='none'>
+        <line
+          x1={start.x}
+          y1={start.y}
+          x2={end.x}
+          y2={end.y}
+          stroke={color}
+          strokeWidth={thickness}
+          stroke-dasharray={strokeString}
+        />
+      </svg>
+      {isSelected && (
+        <LineAnchor
+          style={endAnchorStyle}
+          point={{ x: widget.properties.end.x, y: widget.properties.end.y }}
+          dimensions={{ height: widget.height, width: widget.width }}
+          updateWidget={(newPoint) => {
+            updateAnchor('end', newPoint);
+          }}
+        />
+      )}
+    </div>
+  );
+};
+
+export default LineWidgetComponent;
--- a/packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
+++ b/packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
@@ -1,0 +1,87 @@
+import React, { CSSProperties, useEffect, useRef } from 'react';
+import './component.css';
+import { XYCoord, useDrag } from 'react-dnd';
+import { useGridSettings } from '~/components/actions/useGridSettings';
+
+/**
+ * Compute the new position of the widget based on initial position, offset difference, and grid settings.
+ *
+ * @param {XYCoord} initialPosition - The initial position of the widget.
+ * @param {XYCoord} offsetDifference - The offset difference based on the drag event.
+ * @param {number} height - The height of the widget.
+ * @param {number} width - The width of the widget.
+ * @param {number} cellSize - The size of a cell in the grid.
+ * @return {XYCoord} - The new position (x, y) for the widget.
+ */
+const computeNewPosition = (
+  initialPosition: XYCoord,
+  offsetDifference: XYCoord,
+  height: number,
+  width: number,
+  cellSize: number
+): XYCoord => {
+  const { x: initialX, y: initialY } = initialPosition;
+  const { x: offsetX, y: offsetY } = offsetDifference;
+
+  const widthPixels = Math.round(width * cellSize);
+  const heightPixels = Math.round(height * cellSize);
+
+  const newX = Math.max(0, Math.min(initialX + offsetX, widthPixels));
+  const newY = Math.max(0, Math.min(initialY + offsetY, heightPixels));
+
+  return { x: newX, y: newY };
+};
+
+/**
+ * Custom hook to handle drag-and-drop for LineWidget anchors.
+ *
+ * This hook initializes drag functionality and updates the widget position
+ * whenever a drag event occurs on either anchor.
+ *
+ * @param {LineWidget} widget - The current LineWidget object that is being manipulated.
+ * @param {'start'|'end'} anchorType - Specifies which anchor ('start' or 'end') is being dragged.
+ * @return {ConnectDragSource} - A ref to be attached to the drag component.
+ */
+const useDragAndUpdate = (
+  point: XYCoord,
+  dimensions: { height: number; width: number },
+  updateWidget: (point: XYCoord) => void
+) => {
+  const initialPointRef = useRef({
+    x: 0,
+    y: 0,
+  });
+
+  const { cellSize } = useGridSettings();
+
+  const [{ diff, isDragging }, ref] = useDrag({
+    type: 'LineAnchor',
+    item: () => {
+      initialPointRef.current = point;
+      // we need to return an empty object at the minimum, even if no props are returned
+      return {};
+    },
+    collect: (monitor) => {
+      const offsetDifference = monitor.getDifferenceFromInitialOffset();
+      return { diff: offsetDifference, isDragging: monitor.isDragging() };
+    },
+  });
+
+  useEffect(() => {
+    if (isDragging && diff) {
+      const newPoint = computeNewPosition(initialPointRef.current, diff, dimensions.height, dimensions.width, cellSize);
+      updateWidget(newPoint);
+    }
+  }, [diff?.x, diff?.y]);
+  return ref;
+};
+
+export const LineAnchor: React.FC<{
+  style: CSSProperties;
+  point: XYCoord;
+  dimensions: { height: number; width: number };
+  updateWidget: (point: XYCoord) => void;
+}> = ({ style, point, dimensions, updateWidget }) => {
+  const dragRef = useDragAndUpdate(point, dimensions, updateWidget);
+  return <div data-testid='line-anchor' ref={dragRef} style={style} className='line-anchor' />;
+};
--- a/packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
+++ b/packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import LineWidgetComponent from './component';
+import LineSymbolIcon from './icon';
+import type { DashboardPlugin } from '~/customization/api';
+import { LineWidget } from '../types';
+
+export const linePlugin: DashboardPlugin = {
+  install: ({ registerWidget }) => {
+    registerWidget<LineWidget>('line', {
+      render: (widget) => <LineWidgetComponent {...widget} />,
+      componentLibrary: {
+        name: 'Line Symbol',
+        icon: LineSymbolIcon,
+      },
+      properties: () => ({
+        start: {
+          x: 25,
+          y: 200,
+        },
+        end: {
+          x: 375,
+          y: 200,
+        },
+      }),
+      initialSize: {
+        width: 400,
+        height: 400,
+      },
+    });
+  },
+};
--- a/packages/dashboard/src/customization/widgets/types.ts
+++ b/packages/dashboard/src/customization/widgets/types.ts
@@ -97,6 +97,19 @@ export type RectangleProperties = {
   borderThickness?: number;
 };
 
+type Point = {
+  x: number;
+  y: number;
+};
+
+export type LineProperties = {
+  start: Point;
+  end: Point;
+  lineStyle?: 'solid' | 'dashed' | 'dotted';
+  color?: string;
+  thickness?: number;
+};
+
 type ChartPropertiesUnion =
   | KPIProperties
   | StatusProperties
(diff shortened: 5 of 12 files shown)


packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { useWidgetActions } from '~/customization/hooks/useWidgetActions';
+import { DashboardWidget } from '~/types';
+import { LineAnchor } from './lineAnchor/component';
+import { useIsSelected } from '~/customization/hooks/useIsSelected';
+
+const LineWidgetComponent: React.FC<LineWidget> = (widget) => {
+  const { update } = useWidgetActions<DashboardWidget>();
+  const isSelected = useIsSelected(widget);
+
+  const fitFull = {
```
mnischay, line 13: nit: seems like a constant which need not be created every time. i.e could be outside the component

mnischay: also, this could be just a class? instead of a style variable

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const linePlugin: DashboardPlugin = {
+  install: ({ registerWidget }) => {
+    registerWidget<LineWidget>('line', {
+      render: (widget) => <LineWidgetComponent {...widget} />,
+      componentLibrary: {
+        name: 'Line',
+        icon: LineIcon,
+      },
+      properties: () => ({
+        lineStyle: 'solid',
```
mnischay, line 16: super nit: probably worth creating a const for this value object

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,101 @@
...
+        updateWidget(updatedWidget);
+      }
+    }
+  }, [diff?.x, diff?.y]);
+  return (
+    <div
+      ref={ref}
+      style={{
+        ...style,
+        width: '20px',
```
mnischay, line 92: please move the styling to a css file

packages/react-components/src/components/iot-app-kit-config/defaultValues.ts
```diff
@@ -2,7 +2,7 @@ import { FEATURE_FLAGS } from './types';
 
 const DEFAULT_FEATURE_CONFIG = {
   [FEATURE_FLAGS.ENABLE_E_CHARTS]: false,
-  [FEATURE_FLAGS.ENABLE_SYMBOL_LIBRARY]: false,
+  [FEATURE_FLAGS.ENABLE_SYMBOL_LIBRARY]: true,
```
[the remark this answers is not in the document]
mnischay: Also, please start using the ```getConfigValue``` from the react-component ... instead of the iot-app-config HOC/provider, we will be removing it soon

packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
```diff
@@ -0,0 +1,67 @@
...
+    width: '100%',
+    height: '100%',
+  };
+
+  const {
+    lineStyle = 'solid',
+    color = 'black',
```
mnischay, line 20: just curious: is it worth to just use the cloudscape design tokens ?

jmbuss (changes requested):

packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
corteggiano: +1 to cloudscape tokens, here are the token values
https://cloudscape.design/foundation/visual-foundation/design-tokens/

packages/dashboard/src/components/dragLayer/itemTypes.ts
```diff
@@ -3,4 +3,5 @@ export enum ItemTypes {
   Component = 'Component',
   ResourceExplorerAssetProperty = 'ResourceExplorerAssetProperty',
   ResourceExplorerAlarm = 'ResourceExplorerAlarm',
+  LineAnchor = 'LineAnchor',
```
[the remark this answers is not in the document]
shlokj: done

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,101 @@
...
+      return {
+        dotType,
+      };
+    },
+    collect: (monitor) => {
+      const offsetDifference = monitor.getDifferenceFromInitialOffset();
+      return { diff: offsetDifference, isDragging: monitor.isDragging() };
+    },
+  });
+  useEffect(() => {
```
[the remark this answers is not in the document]
shlokj: putting that code in collect causes far too many renders and the app crashes - i actually tried that before putting it in a useEffect

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { DashboardWidget } from '~/types';
+import { useGridSettings } from '~/components/actions/useGridSettings';
+
+export const LineAnchor: React.FC<{
+  style: CSSProperties;
+  dotType: 'start' | 'end';
+  widget: LineWidget;
+  updateWidget: (widget: DashboardWidget) => UpdateWidgetsAction;
+}> = ({ style, dotType, widget, updateWidget }) => {
+  const initialPositionRef = useRef({
```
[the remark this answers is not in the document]
shlokj: done

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,101 @@
...
+    },
+    collect: (monitor) => {
+      const offsetDifference = monitor.getDifferenceFromInitialOffset();
+      return { diff: offsetDifference, isDragging: monitor.isDragging() };
+    },
+  });
+  useEffect(() => {
+    if (isDragging && diff) {
+      let updatedWidget;
+      if (dotType === 'start') {
```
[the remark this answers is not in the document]
shlokj: done

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+        updatedWidget = {
+          ...widget,
+          properties: {
+            ...widget.properties,
+            endX: newEndX,
+            endY: newEndY,
+          },
+        };
+        console.log(updatedWidget);
```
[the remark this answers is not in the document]
shlokj: done

packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
shlokj: done - moved it to a css file and used it as a class

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
shlokj: done

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
```diff
@@ -0,0 +1,30 @@
...
+        color: 'black',
+        thickness: 5,
+        startX: 25,
+        startY: 25,
+        endX: 375,
```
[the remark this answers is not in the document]
shlokj: yep, navita said it should be horizontal. made the change.

packages/dashboard/src/customization/widgets/types.ts
```diff
@@ -97,6 +97,16 @@ export type RectangleProperties = {
   borderThickness?: number;
 };
 
+export type LineProperties = {
+  lineStyle?: 'solid' | 'dashed' | 'dotted';
+  color?: string;
+  thickness?: number;
+  startX?: number;
```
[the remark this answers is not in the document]
shlokj: you're right. i created the type and made it required.

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
shlokj: i looked at all other dashboard widgets and it's done this way. any merit to creating a const here?

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,101 @@
...
+          properties: {
+            ...widget.properties,
+            startX: newStartX,
+            startY: newStartY,
+          },
+        };
+      } else if (dotType == 'end') {
+        const newEndX = Math.max(
+          0,
+          Math.min((initialPositionRef.current.initialEndX ?? 0) + diff.x, Math.floor(widget.width * cellSize))
```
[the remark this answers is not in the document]
shlokj: sorry, your question seems a bit ambiguous. 

are you suggesting i use `Math.round` instead of `Math.floor`? minor difference, but i think that's a good idea - will implement.

or are you asking whether it already is an integer? yeah, the start/end coordinates are always whole numbers because diff only has integers and `Math.round` (previously `Math.floor`) returns integers.

packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  const endDotStyle: CSSProperties = {
+    position: 'absolute',
+    left: `${endX}px`,
+    top: `${endY}px`,
+  };
+
+  return (
+    <div style={fitFull}>
```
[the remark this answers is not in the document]
shlokj: moved `fitFull`. i don't see the merit in moving `startDotStyle` and `endDotStyle` because we would only be able to move `position: 'absolute'`. let me know what you think.

packages/react-components/src/components/iot-app-kit-config/defaultValues.ts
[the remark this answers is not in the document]
shlokj: left that feature flag true just so i can test the feature. followed up with you on slack regarding your second comment.

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,96 @@
...
+    y: 0,
+  });
+  const initialEndPointRef = useRef({
+    x: 0,
+    y: 0,
+  });
+
+  const { cellSize } = useGridSettings();
+
+  function computeNewPosition(initialPosition: XYCoord, offsetDifference: XYCoord) {
```
jmbuss, line 28: let's move this outside of the component so it is not created for every render.

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const { x: initialX, y: initialY } = initialPosition;
+    const { x: offsetX, y: offsetY } = offsetDifference;
+
+    const widthPixels = Math.round(widget.width * cellSize);
+    const heightPixels = Math.round(widget.height * cellSize);
+
+    const newX = Math.max(0, Math.min(initialX + offsetX, widthPixels));
+    const newY = Math.max(0, Math.min(initialY + offsetY, heightPixels));
+
+    return { x: newX, y: newY } as XYCoord;
```
jmbuss, line 38: is there a way to write this so we don't need to cast?

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,96 @@
...
+    },
+    collect: (monitor) => {
+      const offsetDifference = monitor.getDifferenceFromInitialOffset();
+      return { diff: offsetDifference, isDragging: monitor.isDragging() };
+    },
+  });
+  useEffect(() => {
+    if (isDragging && diff) {
+      let updatedWidget;
+      if (anchorType === 'start') {
```
jmbuss, line 64: is there a way to abstract these blocks? they look the same per each point.

jmbuss: Could we also move the drag and effect hook into a single hook since they are coupled together? This should make the code easier to read.

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { UpdateWidgetsAction } from '~/store/actions';
+import { XYCoord, useDrag } from 'react-dnd';
+import { DashboardWidget } from '~/types';
+import { useGridSettings } from '~/components/actions/useGridSettings';
+import './component.css';
+
+export const LineAnchor: React.FC<{
+  style: CSSProperties;
+  anchorType: 'start' | 'end';
+  widget: LineWidget;
```
jmbuss, line 12: does the anchor need to know about the entire widget or can we pair this down to just a point and updatePoint?

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
```diff
@@ -0,0 +1,34 @@
...
+  install: ({ registerWidget }) => {
+    registerWidget<LineWidget>('line', {
+      render: (widget) => <LineWidgetComponent {...widget} />,
+      componentLibrary: {
+        name: 'Line Symbol',
+        icon: LineSymbolIcon,
+      },
+      properties: () => ({
+        lineStyle: 'solid',
+        color: 'black',
```
jmbuss, line 17: I think we should only define the properties that are required here so that we are not specifying defaults multiple times.

jmbuss (changes requested):
Nice work! One general comment is that I would like to see some tests for this component. The movement functions specifically since they hold all of the logic.

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
shlokj: `return { x: newX, y: newY };`

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
shlokj: done

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
shlokj: i think it does, but i'm not sure. how else would we `updateWidget`?

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
shlokj: cool, keeping only the points then.

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
shlokj: > is there a way to abstract these blocks? they look the same per each point.
done

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+          thickness: 15,
+          start: { x: 180, y: 175 },
+          end: { x: 20, y: 30 },
+        } as LineProperties,
+        isSelected: true,
+      },
+    ].forEach((configuration) => {
+      const widget = { ...mockWidgetPartial, properties: { ...configuration.properties } };
+      const isSelected = configuration.isSelected;
+      it(`should render with ${JSON.stringify(widget)} correctly. The widget ${
```
jmbuss, line 109: Is this line readable in the console? Can we make the title of the test more succinct?

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+import LineWidgetComponent from './component';
+import { Provider } from 'react-redux';
+import { DashboardStore, configureDashboardStore } from '~/store';
+import { DashboardState } from '~/store/state';
+import { LineProperties, LineWidget } from '../types';
+import { act } from 'react-dom/test-utils';
+
+const mockWidgetPartial = {
+  id: 'mock-line-widget',
+  type: 'line',
```
jmbuss, line 14: I think this is the line chart type? Do the types clash?

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+        lineStyle: 'solid',
+        color: 'black',
+        thickness: 5,
+        start: { x: 180, y: 175 },
+        end: { x: 20, y: 30 },
+      },
+    } as LineWidget;
+
+    it('should display anchors when the widget is selected', async () => {
+      // (useIsSelected as jest.Mock).mockReturnValue(true); // this makes it work, but we shouldn't need it
```
jmbuss, line 131: don't forget to cleanup comments / loggers / etc.!

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+        thickness: 5,
+        start: { x: 50, y: 50 },
+        end: { x: 100, y: 100 },
+      },
+    } as LineWidget;
+
+    const renderWithStore = (initialState: Partial<DashboardState>) => {
+      const store = configureDashboardStore(initialState);
+
+      const TestComponentWithStore = () => (
```
jmbuss, line 158: is this different than test component in some way? or is it possible to use that?

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+      const endingPosition = {
+        clientX: 100,
+        clientY: 120,
+      };
+
+      expect(store.getState().dashboardConfiguration.widgets[0].properties.start).toEqual({ x: 50, y: 50 });
+
+      act(() => {
+        fireEvent.dragStart(lineStartAnchor, startingPosition);
+        fireEvent.dragEnter(lineStartAnchor, endingPosition);
```
jmbuss, line 197: are all of these events required? for instance dragEnter / dragOver & drop / dragEnd seem to be doing the same thing

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+        fireEvent.dragOver(lineStartAnchor, endingPosition);
+        fireEvent.drop(lineStartAnchor, endingPosition);
+        fireEvent.dragEnd(lineStartAnchor, endingPosition);
+      });
+
+      console.log(lineStartAnchor);
+
+      act(() => {
```
jmbuss, line 205: take a look at the docs for act - https://github.com/threepointone/react-act-examples/blob/master/sync.md - might help you refactor this test.

packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  if (lineStyle === 'dashed') {
+    strokeString = SVG_STROKE_DASHED;
+  } else if (lineStyle === 'dotted') {
+    strokeString = SVG_STROKE_DOTTED;
+  }
+
+  const startAnchorStyle: CSSProperties = {
+    position: 'absolute',
+    left: `${start.x - 10}px`,
```
jmbuss, line 42: what is the 10 for? let's make a note about it and maybe even move it into some constants file

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,109 @@
+import React, { CSSProperties, useEffect, useRef } from 'react';
+import './component.css';
+import { LineWidget } from '~/customization/widgets/types';
+import { XYCoord, useDrag } from 'react-dnd';
+import { DashboardWidget } from '~/types';
+import { useGridSettings } from '~/components/actions/useGridSettings';
+import { useWidgetActions } from '~/customization/hooks/useWidgetActions';
+
+function computeNewPosition(
```
jmbuss, line 9: we should add jsdoc comments to these functions. Also any reason to mix regular / arrow functions?

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,109 @@
...
+      updateWidget(updatedWidget);
+    }
+  }, [diff?.x, diff?.y]);
+  return ref;
+};
+
+export const LineAnchor: React.FC<{
+  style: CSSProperties;
+  anchorType: 'start' | 'end';
+  widget: LineWidget;
```
jmbuss, line 97: does the anchor need the entire widget to function? I think it would be easier for it to just require a point

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
```diff
@@ -0,0 +1,31 @@
...
+import LineSymbolIcon from './icon';
+import type { DashboardPlugin } from '~/customization/api';
+import { LineWidget } from '../types';
+
+export const linePlugin: DashboardPlugin = {
+  install: ({ registerWidget }) => {
+    registerWidget<LineWidget>('line', {
+      render: (widget) => <LineWidgetComponent {...widget} />,
+      componentLibrary: {
+        name: 'Line Symbol',
```
jmbuss, line 12: does this fit under the icon? I feel like the text might be too large?

packages/dashboard/testing/mocks.ts
```diff
@@ -176,6 +177,52 @@ export const MOCK_RECTANGLE_WIDGET: RectangleWidget = {
...
+      x: 375,
+      y: 200,
+    },
+    lineStyle: 'solid',
+    color: 'black',
+    thickness: 5,
+  },
+};
+
+export const MOCK_LINE_WIDGET_2: LineWidget = {
```
jmbuss, line 203: do we need both of these mocks? they are the same except for the id?

jmbuss (changes requested):

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+      },
+    } as LineWidget;
+
+    const renderWithStore = (initialState: Partial<DashboardState>) => {
+      const store = configureDashboardStore(initialState);
+
+      const TestComponentWithStore = () => (
+        <Provider store={store}>
+          <DndProvider backend={HTML5Backend}>
+            <LineWidgetComponent {...(store.getState().dashboardConfiguration.widgets[0] as LineWidget)} />
```
jmbuss, line 161: Be careful on this line, the widget will never be updated when the store is updated. you would need to use the redux selector for that to happen

jmbuss (changes requested):

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
shlokj: it doesn't get cropped, but it pushes other symbols a little to the side

packages/dashboard/testing/mocks.ts
shlokj: i may have added that to test something and ended up forgetting. removed it.

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
shlokj: good point. i replaced `widget` with `configuration` since that is more relevant. also shortened the title.

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
shlokj: the line chart has `type: 'line-chart'`

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
shlokj: yeah - as discussed on slack, we needed to use `useSelector` for the state to actually update.

anyway, as discussed with @diehbria, we will be skipping these tests for now (removing them from this PR entirely) and moving them to a PR after the other components have been merged in. i will mark all of these tests, related to firing the drag-and-drop event, as resolved for now and come back to them in the future.

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
shlokj: i saw an example that did it this way. will check again in a future pr.

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
shlokj: done

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
shlokj: the anchor uses `update` from `useWidgetActions`. we wouldn't be able to update the widget if we pass in only a point.

packages/dashboard/src/customization/widgets/lineSymbol/component.test.tsx
shlokj: resolving since we will come back to this in a future pr

packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
shlokj: i added comments explaining why we have it. i also took it out into a constant, but it didn't seem worth creating a separate file, so i added `const ANCHOR_HALF_SIDE_LENGTH_PX = 10;` at the top of this file, right after the imports.

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
shlokj: added jsdoc comments to all functions (including the hook) in this file. no reason in particular, happened with the flow - made them arrow functions.

packages/dashboard/src/customization/widgets/lineSymbol/plugin.tsx
shlokj: This is similar to what we see for Timeline, except just more

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
jmbuss: Right, that is the correct way to update the widget, but it looks like the anchor is taking on too many responsibilities. For example, the hook initializes a ref for start and end point but only ever uses 1 of those. In memory this means we have extra refs that we don't need. It also handles updating the entire widget. I think it would be better if the line anchor only tracked 1 endpoint, and we can instead pass a function to update that endpoint on the widget. This would keep all the widget concerns in the line symbol component and the anchor concerns in the line anchor component.

jmbuss (changes requested):

packages/dashboard/src/customization/widgets/lineSymbol/component.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const newWidget = { ...widget, properties: updatedProperties };
+    updateWidget(newWidget);
+  };
+
+  return (
+    <div className='fit-full'>
+      {isSelected && (
+        <LineAnchor
+          style={startAnchorStyle}
+          anchorType='start'
```
jmbuss, line 71: is anchor type still required now that it handles only 1 point?

packages/dashboard/src/customization/widgets/lineSymbol/lineAnchor/component.tsx
```diff
@@ -0,0 +1,98 @@
...
+  style: CSSProperties;
+  anchorType: 'start' | 'end';
+  point: XYCoord;
+  dimensions: { height: number; width: number };
+  updateWidget: (point: XYCoord) => void;
+}> = ({ style, anchorType, point, dimensions, updateWidget }) => {
+  const dragRef = useDragAndUpdate(point, dimensions, anchorType, updateWidget);
+  return (
+    <div
+      id='line-symbol-widget'
```
jmbuss, line 91: We should use class instead since this doesn't uniquely id 1 html element. There's always 2.

PR #1798, opened 2023-08-14
fix(scene composer): Color picker bug fixes

divya-sea:
* bug fixes for color picker
## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPicker.tsx
+++ b/packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPicker.tsx
@@ -1,7 +1,7 @@
 import { Button, FormField, Icon, Input, InputProps, SpaceBetween, TextContent } from '@awsui/components-react';
 import { NonCancelableCustomEvent } from '@awsui/components-react/internal/events';
-import React, { useCallback, useState } from 'react';
-import { ChromePicker, CirclePicker } from 'react-color';
+import React, { useCallback, useEffect, useState } from 'react';
+import { CirclePicker, ColorResult, SketchPicker } from 'react-color';
 import { useIntl } from 'react-intl';
 
 import { IColorPickerProps } from '../interface';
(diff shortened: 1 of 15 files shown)


packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPicker.tsx
```diff
@@ -69,22 +86,62 @@ export const ColorPicker = ({ color, onSelectColor, label }: IColorPickerProps):
...
+      setHexCodeError(''); // Clear any existing error message
+      setNewColor(color.hex);
+      onSelectColor(color.hex);
+    },
+    [color, onSelectColor, onUpdateCustomColors],
+  );
+
+  const handleHexCodeChange = useCallback(
+    (event: NonCancelableCustomEvent<InputProps.ChangeDetail>) => {
+      setNewColor(event.detail.value);
```
haweston, line 105: Do we want to set the color if it's not valid hex code?

packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPicker.tsx
```diff
@@ -34,30 +41,40 @@ export const ColorPicker = ({ color, onSelectColor, label }: IColorPickerProps):
...
-  }, []);
+
+    return () => {
+      document.removeEventListener('click', handleOutsideClick);
+    };
+  }, [showPicker, handleOutsideClick]);
+
+  const checkIfCustomColor = (color: any) => {
+    if (!Object.values(palleteColors).includes(color)) {
+      setCustomInternalColors(customInternalColors.concat(color));
```
haweston, line 70: What's the experience here? Are we tracking all previous attempted custom colors?

divya-sea: Experience is we provide basic colors that will be displayed by default to all. But of they chose from the StechPicker and that hex code is part of the basic default color then we don't want to add those colors to part of the custom colors. This requirement is from UX. So this condition will prevent adding redundant colors to custom color list.

packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPicker.tsx
divya-sea: First, we want to set it to the state variable as this is coming as external prop. Then we are validating whether it is a valid color or not. In the same method we are validating.

haweston: in handleHexCodeChange it doesn't look like an external prop yet. it's an internal event.detail.value that we haven't validated yet?

divya-sea: const [newColor, setNewColor] = useState<string>(color);
Color is a prop and I am using a single state variable to update the state of the component. So initially will have a value that is coming from prop. If user change the hex code then we are checking whther its a valid or not then resetting it again.
const handleHexCodeChange = useCallback(
    (event: NonCancelableCustomEvent<InputProps.ChangeDetail>) => {
      setNewColor(event.detail.value);
}

haweston (approved):

PR #1799, opened 2023-08-15
feat(react-component): adding config service to toggle echarts

mnischay:
## Overview
adding a simple config service which helps with toggling between echarts and synchro charts

## Verifying Changes
refer to screen capture

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/1012248e-d8b3-4481-9418-7a714bf0f6bb


packages/react-components/src/store/config.ts
```diff
@@ -0,0 +1,12 @@
+import { StateCreator } from 'zustand/esm';
+import Cookies from 'js-cookie';
+
+export type Flags = 'useEcharts';
+export interface ConfigState {
+  config: Record<Flags, boolean>;
+}
+export const createConfigSlice: StateCreator<ConfigState> = () => ({
+  config: {
+    useEcharts: !!Cookies.get('USE_ECHARTS'),
```
tracy-french, line 10: Could we use local storage instead of cookies? We have to get consent from users to use cookies.

tracy-french (changes requested):

packages/react-components/src/store/config.ts
mnischay: Fair call, updating to localStorage

tracy-french (approved):
Looks good! Thank you @mnischay!

PR #1809, opened 2023-08-17
feat(dashboard): sync echarts viewport

corteggiano:
## Overview

sync viewports across widgets in dashboard through the useViewport hook

before:
view

after:
TODO add after gif

## Legal


packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -190,76 +190,80 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
...
-              )}
-            </SpaceBetween>
-          </Box>
-        </div>
-        <div className='display-area' ref={(el) => setViewFrameElement(el || undefined)}>
-          <ReadOnlyGrid {...grid}>
-            <Widgets {...widgetsProps} />
-          </ReadOnlyGrid>
-          <WebglContext viewFrame={viewFrame} />
+      <TimeSync>
```
corteggiano, line 193: this file's only change is wrapping the components with the <timesync> context

packages/react-components/src/components/chart/hooks/useDataZoomListeners.ts
```diff
@@ -0,0 +1,40 @@
...
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        //@ts-ignore
+        currentDataZoom.start > 0 ? (newDataZoom.startValue = currentDataZoom.startValue) : (newDataZoom.start = 0);
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        //@ts-ignore
+        currentDataZoom.end < 100 ? (newDataZoom.endValue = currentDataZoom.endValue) : (newDataZoom.end = 100);
+
+        setViewport(
+          { start: new Date(currentDataZoom.startValue), end: new Date(currentDataZoom.endValue) },
+          'echarts-gesture'
```
jmbuss, line 29: Should we move these to some string constant?

packages/react-components/src/components/chart/hooks/useDataZoomListeners.ts
```diff
@@ -0,0 +1,40 @@
...
+      // this will prevent the data from moving if we are zoomed in
+      chart?.on('dataZoom', function () {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        //@ts-ignore
+        const currentDataZoom = chart?.getOption().dataZoom[0];
+
+        const newDataZoom = {};
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        //@ts-ignore
+        currentDataZoom.start > 0 ? (newDataZoom.startValue = currentDataZoom.startValue) : (newDataZoom.start = 0);
```
jmbuss, line 22: should we move this to some other function which handles updating the dataZoom?

packages/react-components/src/components/chart/eChartsConstants.ts
```diff
@@ -19,6 +19,7 @@ export const DEFAULT_Y_AXIS: YAXisComponentOption = {
   type: 'value',
   axisLabel: { formatter: format('.2s') },
   position: 'left',
+  min: (val) => val.min - 0.5,
```
jmbuss, line 22: can we add a note for the .5? Is it for padding?

packages/react-components/src/components/chart/hooks/useDataZoomListeners.ts
```diff
@@ -0,0 +1,40 @@
+import { useEffect } from 'react';
+import { useViewport } from '../../../hooks/useViewport';
+import { EChartsType, getInstanceByDom } from 'echarts';
+
+export const useDataZoomListeners = (ref: React.RefObject<HTMLDivElement>) => {
+  const { setViewport } = useViewport();
+
+  useEffect(() => {
+    if (ref.current) {
+      const chart: EChartsType | undefined = getInstanceByDom(ref.current);
```
jmbuss, line 10: is it better to get the ref from the dom? I think we also have the chartRef available from the useEChart hook

jmbuss (commented):
Nice! left a few comments!

packages/react-components/src/components/chart/hooks/useDataZoomListeners.ts
corteggiano: yeah, ill add them to a file in the hook's folder

corteggiano: made a constant string for this value!

packages/react-components/src/components/chart/hooks/useDataZoomListeners.ts
corteggiano: that's basically what this whole file is for, just setting up event handlers for dataZoom

jmbuss (approved):

PR #1815, opened 2023-08-18
feat(TM-source): add property value query

sheilaXu:
## Overview
add property value query and support it in data binding provider

https://github.com/awslabs/iot-app-kit/assets/9315598/af86648f-d8c5-4a4b-b684-447d093db33c

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/source-iottwinmaker/src/initialize.ts
+++ b/packages/source-iottwinmaker/src/initialize.ts
@@ -18,16 +18,18 @@ import { S3SceneLoader } from './scene-module/S3SceneLoader';
 import { SceneMetadataModule } from './scene-module/SceneMetadataModule';
 import { KGDataModule } from './knowledgeGraph-module/KGDataModule';
 import { VideoDataImpl } from './video-data/VideoData';
-import { ErrorDetails, TimeSeriesDataModule } from '@iot-app-kit/core';
+import { DataBase, ErrorDetails, Query, TimeSeriesDataModule, DataRequest } from '@iot-app-kit/core';
 import { TwinMakerTimeSeriesDataProvider } from './time-series-data/provider';
 import { createDataSource } from './time-series-data/data-source';
 import { TwinMakerMetadataModule } from './metadata-module/TwinMakerMetadataModule';
 import type { Credentials, CredentialProvider } from '@aws-sdk/types';
 import type { VideoDataProps } from './types';
-import type { TwinMakerDataStreamQuery, TwinMakerQuery } from './time-series-data/types';
+import type { TwinMakerDataStreamQuery } from './time-series-data/types';
 import type { TimeSeriesDataQuery, TimeSeriesDataRequest } from '@iot-app-kit/core';
 import { TwinMakerErrorCode } from './common/error';
 import { createEntityPropertyBindingProvider } from './data-binding-provider/createEntityPropertyBindingProvider';
+import { TwinMakerHistoryQuery, TwinMakerPropertyValueQuery } from './common/queryTypes';
+import { TwinMakerPropertyValueDataProvider } from './property-value/provider';
 
 const SOURCE = 'iottwinmaker';
 
@@ -129,11 +131,20 @@ export const initialize = (
   const secretsManagerClient: SecretsManagerClient =
     authInput.secretsManagerClient ?? secretsManagersdk(inputWithCred.awsCredentials, inputWithCred.awsRegion);
 
-  // For caching TwinMaker API calls
+  // For caching TwinMaker metadata API calls
   const cachedQueryClient = new QueryClient({
     defaultOptions: {
       queries: {
         staleTime: Infinity,
+      },
+    },
+  });
+
+  // For refreshing TwinMaker property value API calls
+  const refreshingQueryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        staleTime: 1000,
       },
     },
   });
@@ -143,7 +154,9 @@ export const initialize = (
     createDataSource(twinMakerMetadataModule, twinMakerClient)
   );
 
-  const timeSeriesDataQuery: (query: TwinMakerQuery) => TimeSeriesDataQuery = (query: TwinMakerQuery) => ({
+  const timeSeriesDataQuery: (query: TwinMakerHistoryQuery) => TimeSeriesDataQuery = (
+    query: TwinMakerHistoryQuery
+  ) => ({
     toQueryString: () =>
       JSON.stringify({
         source: SOURCE,
@@ -162,15 +175,36 @@ export const initialize = (
       }),
   });
 
+  const propertyValueQuery = (query: TwinMakerPropertyValueQuery): Query<DataBase[], DataRequest> => ({
+    toQueryString: () =>
+      JSON.stringify({
+        source: SOURCE,
+        queryType: 'property-value',
+        query,
+      }),
+    build: (_sessionId: string, params: DataRequest) =>
+      new TwinMakerPropertyValueDataProvider(refreshingQueryClient, twinMakerClient, {
+        queries: [
+          {
+            workspaceId,
+            ...query,
+          },
+        ],
+        request: params,
+      }),
+  });
+
   return {
     query: {
-      timeSeriesData: (query: TwinMakerQuery): TimeSeriesDataQuery => timeSeriesDataQuery(query),
+      timeSeriesData: timeSeriesDataQuery,
+      propertyValue: propertyValueQuery,
     },
     s3SceneLoader: (sceneId: string) => new S3SceneLoader({ workspaceId, sceneId, twinMakerClient, s3Client }),
     valueDataBindingProviders: (onError?: (errorCode: TwinMakerErrorCode, errorDetails?: ErrorDetails) => void) => ({
       TwinMakerEntityProperty: createEntityPropertyBindingProvider({
         metadataModule: twinMakerMetadataModule,
         timeSeriesDataQuery,
+        propertyValueQuery,
         onError,
       }),
     }),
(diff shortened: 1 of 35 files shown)


packages/source-iottwinmaker/src/utils/propertyValueUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+  return JSON.stringify(value);
+};
+
+/**
+ * Convert the TwinMaker DataType into AppKit DataType
+ *
+ * @param tmDataType the TwinMaker DataType to be converted.
+ * @returns the converted AppKit DataType
+ */
+export const toDataType = (tmDataType: TMDataType): DataType | undefined => {
```
divya-sea, line 48: I remember when I was working on Athena data connector, if customer created a database using List/Map( advance  data type), Console was throwing an error. How do we want to handle here - https://docs.aws.amazon.com/iot-twinmaker/latest/apireference/API_DataType.html

packages/source-iottwinmaker/src/utils/propertyValueUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @returns Returns false if the value is nullish, otherwise true.
+ */
+export const isDefined = <T>(value: T | null | undefined): value is T => !isNil(value);
+
+/**
+ * Extracts the value out of a TwinMaker property's DataValue
+ *
+ * NOTE: Currently we treat booleans and objects as strings.
+ */
+export const toValue = (dataValue: DataValue): Primitive | undefined => {
```
divya-sea, line 20: Maybe we can rename to `propertyDataValue`?

divya-sea: What's the reason for extracting only propertyValues?

packages/source-iottwinmaker/src/property-value/provider.ts
```diff
@@ -0,0 +1,76 @@
...
+      queryFn: async () => {
+        try {
+          const promises: Promise<DataStreamBase<Primitive>[]>[] = [];
+          queries.forEach((query) => {
+            promises.push(getPropertyValueByEntity({ client: tmClient, query: query }));
+          });
+          return [{ dataStreams: (await Promise.all(promises)).flat() }];
+        } catch (_err) {
+          // TODO: return error
+          return [];
```
divya-sea, line 43: Maybe return an error object instead of returning the empty array?

packages/source-iottwinmaker/src/property-value/provider.ts
```diff
@@ -0,0 +1,76 @@
...
+export class TwinMakerPropertyValueDataProvider implements Provider<DataBase[]> {
+  public tmClient: IoTTwinMakerClient;
+  public queryClient: QueryClient;
+
+  public observer: QueryObserver<DataBase[], unknown, DataBase[], DataBase[], TwinMakerStaticDataQuery[]>;
+
+  private _unsubscribes: (() => void)[] = [];
+
+  constructor(
+    queryClient: QueryClient,
```
divya-sea, line 20: Instead of instantiating the tmClient and queryClient inside the constructor, Consider them passing as dependencies to the class for better flexibility.

packages/source-iottwinmaker/src/property-value/provider.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { IoTTwinMakerClient } from '@aws-sdk/client-iottwinmaker';
+import { getPropertyValueByEntity } from './client/getPropertyValueByEntity';
+import { TwinMakerStaticDataQuery } from './types';
+
+/**
+ * Provider for TwinMaker time series data
+ */
+export class TwinMakerPropertyValueDataProvider implements Provider<DataBase[]> {
+  public tmClient: IoTTwinMakerClient;
```
divya-sea, line 12: Constructor takes tmClient and queryClient. Maybe we can add input validations to ensure these parameters are provided and have expected type?

packages/source-iottwinmaker/src/property-value/client/getPropertyValueByEntity.ts
```diff
@@ -0,0 +1,60 @@
...
+export const getPropertyValueByEntity = async ({
+  client,
+  query,
+}: {
+  client: IoTTwinMakerClient;
+  query: TwinMakerStaticDataQuery;
+}): Promise<DataStreamBase[]> => {
+  const results: DataStreamBase[] = [];
+  let nextToken: string | undefined = undefined;
+  do {
```
divya-sea, line 16: Maybe you can add a try catch block here and handle potential failures?

packages/source-iottwinmaker/src/property-value/client/getPropertyValueByEntity.ts
```diff
@@ -0,0 +1,60 @@
+import { GetPropertyValueCommand, GetPropertyValueResponse, IoTTwinMakerClient } from '@aws-sdk/client-iottwinmaker';
+import { DataType, DataStreamBase } from '@iot-app-kit/core';
+import { TwinMakerStaticDataQuery } from '../types';
+import { toValue } from '../../utils/propertyValueUtils';
+import { isNumber } from 'lodash';
+
+export const getPropertyValueByEntity = async ({
```
divya-sea, line 7: add a one liner what is this function is about?

packages/source-iottwinmaker/src/property-value/client/getPropertyValueByEntity.ts
```diff
@@ -0,0 +1,60 @@
...
+  client,
+  query,
+}: {
+  client: IoTTwinMakerClient;
+  query: TwinMakerStaticDataQuery;
+}): Promise<DataStreamBase[]> => {
+  const results: DataStreamBase[] = [];
+  let nextToken: string | undefined = undefined;
+  do {
+    const response: GetPropertyValueResponse = await client.send(
```
divya-sea, line 17: Looks like return type of this GetPropertyValueResponse and gets the property values for a component, component type, entity, or workspace. But not about the history of a time series property value? Are you handling that separately?

packages/source-iottwinmaker/src/property-value/provider.ts
sheilaXu: error handling is my next task

packages/source-iottwinmaker/src/property-value/provider.ts
sheilaXu: they are parameters passed in right now

packages/source-iottwinmaker/src/property-value/client/getPropertyValueByEntity.ts
divya-sea: We should also test with large response data to test the performance of the fetching propertyValue and calculate the average time to parse and fetch results. That depends on p90/p99 of API.

packages/source-iottwinmaker/src/property-value/provider.ts
sheilaXu: they are required parameters with correct types, so the input validation is already there

packages/source-iottwinmaker/src/property-value/client/getPropertyValueByEntity.ts
sheilaXu: the failures will be thrown to caller and caught there.

packages/source-iottwinmaker/src/utils/propertyValueUtils.ts
sheilaXu: AppKit data type doesn't support list or map currently, so I convert them to strings for now. when we have the use case to handle list/map/table, we can update here to return them in proper format

packages/source-iottwinmaker/src/property-value/provider.ts
```diff
@@ -0,0 +1,76 @@
...
+      queries: TwinMakerStaticDataQuery[];
+    }
+  ) {
+    this.queryClient = queryClient;
+    this.tmClient = tmClient;
+    this.observer = new QueryObserver(queryClient, {
+      queryKey: queries,
+      queryFn: async () => {
+        try {
+          const promises: Promise<DataStreamBase<Primitive>[]>[] = [];
```
divya-sea, line 36: promise is a generic word used, you could rename it better readability?

packages/source-iottwinmaker/src/utils/propertyValueUtils.ts
sheilaXu: this is a function moved from another file and follow the name with SW source. the purpose of this function is just to get the actual value from the `DataValue` type

packages/source-iottwinmaker/src/utils/propertyValueUtils.ts
divya-sea: Why do we want to convert to string instead of saying not supported types? You can exclude them when you are reading the property and its data type?

packages/source-iottwinmaker/src/initialize.ts
```diff
@@ -138,12 +140,23 @@ export const initialize = (
     },
   });
 
+  // For refreshing TwinMaker property value API calls
+  const refreshingQueryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        staleTime: 1,
```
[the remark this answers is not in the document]
sheilaXu: I'll update to 1s if not issue found

packages/source-iottwinmaker/src/property-value/client/getPropertyValueByEntity.ts
sheilaXu: ok

packages/source-iottwinmaker/src/property-value/client/getPropertyValueByEntity.ts
sheilaXu: I have a separate task to test with a large amount of data bindings

packages/source-iottwinmaker/src/property-value/provider.ts
sheilaXu: it's probably not too bad and short enough to just be here

packages/source-iottwinmaker/src/utils/propertyValueUtils.ts
sheilaXu: they can still be displayed in say overlay as string.

divya-sea (approved):

PR #1817, opened 2023-08-22
fix(react-component): updating TC method to use echarts's methods

mnischay:
## Overview
updating the calculateTimeStamp method to use echarts's for obtaining timestamp of click
1. adding echarts to dashboard's bar and scatter charts
2. updating the calculateTimeStamp method to use echarts's for obtaining timestamp of click
3. using chartRef instead of ref to within trendcursors
4. adding interpolation for different step types

## Verifying Changes
refer to screen capture

## Legal


packages/dashboard/src/customization/widgets/barChart/component.tsx
```diff
@@ -33,6 +34,27 @@ const BarChartWidgetComponent: React.FC<BarChartWidget> = (widget) => {
   const aggregation = getAggregation(queryConfig);
 
   const significantDigits = widgetSignificantDigits ?? dashboardSignificantDigits;
+  const grid = useSelector((state: DashboardState) => state.grid);
+  const size = { width: grid.cellSize * widget.width, height: grid.cellSize * widget.height };
+
+  if (showEcharts) {
+    return (
+      <Chart
```
mnischay, line 42: we should probably have a generic widget component and this component might not be even required...

packages/react-components/src/components/chart/utils/getInfo.ts
```diff
@@ -1,9 +1,9 @@
 import { DurationViewport, Viewport } from '@iot-app-kit/core/src';
```
corteggiano, line 1: is this whole file doing interpolation calculations? if so could we rename it to something less generic... getInfo doesn't tell us much about what it does

mnischay: Sure thing, any suggestions ? , it is all the auxiliary functions we have for charts / TCs...

packages/dashboard/src/customization/widgets/barChart/component.tsx
corteggiano: we could split each out into a separate file in the utils folder, or move the TC/interpolation helpers closer to where they're used

packages/react-components/src/components/chart/utils/getInfo.ts
corteggiano: we could split each out into a separate file in the utils folder, or move the TC/interpolation helpers closer to where they're used

corteggiano (approved):
looks good! lets go TCs 🥳

PR #1844, opened 2023-08-30
feat(scene composer): added icon picker changes

divya-sea:
## Overview
* Added Icon picker changes where now display ~1900 icons added using font awesome library.
* Added font awesome library to package.json.
* Added icon styles.
* Unit tests

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/54058998/6bb04a4e-316c-4d80-8718-00efb48a34eb

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/interfaces/components.ts
```diff
@@ -49,6 +51,7 @@ export interface IAnchorComponent extends ISceneComponent {
   offset?: Vector3;
   chosenColor?: string;
   customColors?: string[];
+  customIcon?: IconDefinition;
```
haweston, line 54: So this will show up in our save data for the scene. Is there a way to abstract the type so we're not directly dependent on a 3rd party type in our save format?

divya-sea: Yes. I will create an internal type as it makes sense to use our own type in case if they change their definition.

packages/scene-composer/src/interfaces/components.ts
```diff
@@ -49,6 +50,7 @@ export interface IAnchorComponent extends ISceneComponent {
   offset?: Vector3;
   chosenColor?: string;
   customColors?: string[];
+  customIcon?: IIconDefinition;
```
sheilaXu, line 53: the `IIconDefinition` has too much info to be saved into scene file. do we really need to support customized icon height width? I thought we will just need the icon name/id so we know which icon to render, then our code will handle sizing stuff

packages/scene-composer/src/components/panels/scene-components/tag-style/IconPicker/IconPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const [newIcon, setNewIcon] = useState<IconDefinition>(selectedIcon ?? faInfo);
+  const [showPicker, setShowPicker] = useState<boolean>(false);
+  const iconsPerPage = 8; // Number of icons per page
+  const defaultRows = 3; // Number of rows to display by default
+  const icons: IconDefinition[] = Object.values(fas); // Array of all Font Awesome Solid icons
+  const totalPages = Math.ceil(icons.length / iconsPerPage); // Calculate the total number of pages
+  const [currentPage] = useState<number>(1);
+  const [numRows, setNumRows] = useState(defaultRows);
+  const generateRandomString = Math.random().toString(16).slice(2);
+  const [randomDomId] = useState<string>(generateRandomString);
```
sheilaXu, line 25: looks like this component will need to be fully rerendered during each render cycle because all ids will be changed. is there any other identifier to use so we only rerender when needed?

packages/scene-composer/src/interfaces/components.ts
divya-sea: No. I also thought `faInfo` is a name but its an object of type IconDefinition. Icondefinition has prefix, name and icon. Their api accepts an icon of type Icondefinition. And each IconDefinition has different configuration that means different height/width every time user selects on the page. Icon prop has an of configuration of that icon including path to svg. We really can't skip any of this as it is same as saving databinding object. I don't think this will introduce any issues.

packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPickerUtils/SvgParserHelper.tsx
```diff
@@ -30,3 +46,20 @@ export const colorPickerPreviewSvg = (color: string): JSX.Element => {
     </svg>
   );
 };
+
+function adjustIconSize(element: Element, iconWidth: number | undefined, iconHeight: number | undefined) {
```
sheilaXu, line 50: comments for what this helper is doing with some examples?

packages/scene-composer/src/components/panels/scene-components/tag-style/IconPicker/IconPicker.tsx
divya-sea: I don't think so. I did not want to change the ID of the UI elements for every render but I wanted to use a unique one for different instances of icon picker. State variable won't change for every render unless we update it with set state. I think this is best way to get an random ID without updating a state and changing the active state of the picker.

packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPickerUtils/SvgParserHelper.tsx
divya-sea: next pr

packages/scene-composer/src/interfaces/components.ts
haweston: Okay this is addressed.

sheilaXu (changes requested):
Please revisit the Tag interface that will be stored in scene file and only keep the minimum fields needed to identify the icon. 

@divya-sea - We have addressed the issue with Tag interface. I guess we can close this comment.

packages/scene-composer/src/components/panels/scene-components/tag-style/IconPicker/IconPicker.tsx
sheilaXu: oh, `Math.random()` is called during every render so new `generateRandomString` will be generated, but it won't update the state after initialize

divya-sea: Yes. we are not setting this anywhere. This actually good wrt to performance as well.

packages/scene-composer/src/interfaces/components.ts
sheilaXu: the icon has type 
```
export type IconProp = IconName | [IconPrefix, IconName] | IconLookup;
export interface IconLookup {
  prefix: IconPrefix;
  iconName: IconName;
}
```
so prefix + name is good enough

haweston (approved):

packages/scene-composer/src/interfaces/components.ts
```diff
@@ -49,6 +50,7 @@ export interface IAnchorComponent extends ISceneComponent {
   offset?: Vector3;
   chosenColor?: string;
   customColors?: string[];
```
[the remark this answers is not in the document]
sheilaXu: if this is the recently used colors, then it should be stored at scene file level

packages/scene-composer/src/interfaces/components.ts
divya-sea: In this approach, I will have to internally search every time for this IconName and get the IconDefinition. Because I don't think IconName give us theIconDefinition out of the box.

packages/scene-composer/src/interfaces/components.ts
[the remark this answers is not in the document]
divya-sea: It is not recently used color. According to UX mock-ups we should display customcolors ( not part of basic color pallete) in a separate container those are picked from Sktech picker. This is a good option to have as an API as well if customer has a list of preferred color they provide here that will be displayed and they can easily picked from that list. 

We have not implemented recently used colors/icons yet. I am going to display chosenColor and chosenIcon reading the tag component from document and display them below `default icons` as `used in scene`.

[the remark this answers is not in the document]
divya-sea: Sync-up with PM and UX team to get the consensus on saving them at the scene level.

packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPickerUtils/SvgParserHelper.tsx
divya-sea: Done

packages/scene-composer/src/interfaces/components.ts
```diff
@@ -1,3 +1,4 @@
+import { IIconLookup } from '../components/panels/scene-components/tag-style/IconPicker/IconPickerUtils/IconPickerUtils';
```
sheilaXu, line 1: better to just define `IIconLookup` in this file since it's the type for a data in tag scene component

packages/scene-composer/src/components/panels/scene-components/tag-style/IconPicker/IconPickerUtils/IconPickerUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+//     string, // unicode
+//     IIconPathData, // svgPathData
+//   ];
+// }
+// export type IIconPathData = string | string[];
+
+// export interface IIconDefinition1 {
+//   prefix: string;
+//   iconName: string;
+// }
```
sheilaXu, line 17: clean up

packages/scene-composer/src/interfaces/components.ts
divya-sea: done

sheilaXu (approved):

PR #1846, opened 2023-08-30
fix(scene tags): fix anchor stems not triggering onWidgetClick

haweston:
## Overview
Problem: Tags don't consistently triggering onWidgetClicked even though they do onSelectionChange reliably.  
Cause: an Anchor is actually a group with two parts in it: the anchor (circle) and a LineSegment (the white stem ). Only the anchor has onClick eventing.
Solution:  Adding the same onClick handler to the line segment as well allows the onWidgetClick to trigger no matter what part of the anchor is clicked.  

## Verifying Changes
1. Story book scene composer scene with an entity bound tag
2. Move camera to where you can see the tag AND the white stem
3. Click on both the tag and the white stem and observe that both the selectionChanged and widgetClicked events fire in the story book action log.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/__snapshots__/AnchorWidget.spec.tsx.snap
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/__snapshots__/AnchorWidget.spec.tsx.snap
@@ -13,7 +13,9 @@ exports[`AnchorWidget should render correctly 1`] = `
       }
     }
   >
-    <lineSegments>
+    <lineSegments
+      onClick={[Function]}
+    >
       <lineBasicMaterial
         color="#ffffff"
       />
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/__snapshots__/AnchorWidget.spec.tsx.snap
```diff
@@ -13,7 +13,9 @@ exports[`AnchorWidget should render correctly 1`] = `
       }
     }
   >
-    <lineSegments>
```
divya-sea, line 16: I have added a test to mock onWidgetClick by mocking the event. Maybe we can add similarly for line segment too? 
https://github.com/awslabs/iot-app-kit/blob/main/packages/scene-composer/src/augmentations/components/three-fiber/anchor/__tests__/AnchorWidget.spec.tsx#L241

haweston: Yep, let me add that since we already have it set up.

divya-sea (approved):

PR #1848, opened 2023-08-31
fix(propertyName): added copy button and textArea for long propertyNa…

raymfang:
## Overview
Customers are facing issues with long propertyNames: 1) not being able to copy them over to rule expressions easily, and 2) not being able to read their rule expressions due to the long names taking up the entire text input field.

To resolve this, a copy button for propertyName has been added, and the rule expression input is now a textArea, providing customers with more space to type and view their rule expressions.

## Verifying Changes

Run scene composer in storybook, and click on a tag. Assign an entity, component, and property to the tag. Notice that the copy button appears when a property is selected in the dropdown menu.

Open the rules panel and notice that the rule expression is in an expandable text area.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/SceneRulesPanel.tsx
+++ b/packages/scene-composer/src/components/panels/SceneRulesPanel.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useContext, useState } from 'react';
-import { AttributeEditor, Box, Button, FormField, Input, SpaceBetween } from '@awsui/components-react';
+import { AttributeEditor, Box, Button, FormField, Input, SpaceBetween, Textarea } from '@awsui/components-react';
 import { useIntl } from 'react-intl';
 
 import { useSceneDocument } from '../../store';
@@ -16,10 +16,9 @@ interface ISceneRuleMapExpandableInfoSectionProps {
   ruleMap: IRuleBasedMapInternal;
 }
 
-const SceneRuleMapExpandableInfoSection: React.FC<React.PropsWithChildren<ISceneRuleMapExpandableInfoSectionProps>> = ({
-  ruleBasedMapId,
-  ruleMap,
-}: ISceneRuleMapExpandableInfoSectionProps) => {
+export const SceneRuleMapExpandableInfoSection: React.FC<
+  React.PropsWithChildren<ISceneRuleMapExpandableInfoSectionProps>
+> = ({ ruleBasedMapId, ruleMap }: ISceneRuleMapExpandableInfoSectionProps) => {
   const sceneComposerId = useContext(sceneComposerIdContext);
   const { removeSceneRuleMapById, updateSceneRuleMapById } = useSceneDocument(sceneComposerId);
   const [newRule, setNewRule] = useState<IRuleStatementInternal | undefined>(undefined);
@@ -82,7 +81,8 @@ const SceneRuleMapExpandableInfoSection: React.FC<React.PropsWithChildren<IScene
           {
             label: intl.formatMessage({ defaultMessage: 'Expression', description: 'Input field label' }),
             control: (item, itemIndex) => (
-              <Input
+              <Textarea
+                data-testid='rule-expression'
                 value={item.expression}
                 placeholder={intl.formatMessage({
                   defaultMessage: 'e.g. value > 0',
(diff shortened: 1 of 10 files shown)


packages/scene-composer/src/components/panels/scene-components/common/ValueDataBindingBuilder.tsx
```diff
@@ -193,14 +204,53 @@ export const ValueDataBindingBuilder: React.FC<IValueDataBindingBuilderProps> =
...
+                        size='small'
+                        position='top'
+                        triggerType='custom'
+                        dismissButton={false}
+                        content={<StatusIndicator type='success'>{selectedOption.label + ' copied'} </StatusIndicator>}
+                      >
+                        <Button
+                          variant='icon'
+                          iconName='copy'
+                          ariaLabel='Copy property name'
```
divya-sea, line 221: This needs to be localized.

packages/scene-composer/src/components/panels/scene-components/common/ValueDataBindingBuilder.tsx
```diff
@@ -193,14 +204,53 @@ export const ValueDataBindingBuilder: React.FC<IValueDataBindingBuilderProps> =
...
+                          onClick={() => {
+                            navigator.clipboard.writeText(selectedOption!.label ? selectedOption!.label : '');
```
divya-sea, line 223: Return value of `window.clipboard.writeText` is a promise. so you need to cover them with async and await.

packages/scene-composer/src/components/panels/SceneRulesPanel.tsx
```diff
@@ -82,7 +82,7 @@ const SceneRuleMapExpandableInfoSection: React.FC<React.PropsWithChildren<IScene
           {
             label: intl.formatMessage({ defaultMessage: 'Expression', description: 'Input field label' }),
             control: (item, itemIndex) => (
-              <Input
+              <Textarea
```
divya-sea, line 85: What's the reason to moving from Input to TextArea?

raymfang: Customers would have to paste in a very long propertyName (e.g. 'Property1111-2222-3333-4444-5555') into the rule expression and it would take up all of the text input field, preventing them from seeing the rest of their rule expression (e.g. "== 50").

packages/scene-composer/src/components/panels/scene-components/common/ValueDataBindingBuilder.scss
```diff
@@ -0,0 +1,8 @@
+.copy-select-flex {
+    display: flex;
+    flex-wrap: nowrap;
+}
+
+.full-width {
```
sheilaXu, line 6: prefix class name with `tm-` to avoid conflict with others

packages/scene-composer/src/components/panels/scene-components/common/ValueDataBindingBuilder.tsx
```diff
@@ -193,14 +204,53 @@ export const ValueDataBindingBuilder: React.FC<IValueDataBindingBuilderProps> =
...
-              <Select
-                data-testid='value-data-binding-builder-select'
-                selectedOption={selectedOption}
-                onChange={onSelectChange(definition.fieldName, index)}
-                options={options}
-                disabled={disabled}
-                placeholder={intl.formatMessage({ defaultMessage: 'Select an option', description: 'placeholder' })}
-              />
+              {definition.fieldName === 'propertyName' && (
+                <div className='copy-select-flex'>
```
divya-sea, line 208: Please refer to this https://cloudscape.aws.dev/components/space-between/?tabId=playground. You can use this to place them in the same line. And also you don't need to have two separate select as you can toggle copy button based on the definition.

packages/scene-composer/src/components/panels/scene-components/common/ValueDataBindingBuilder.scss
divya-sea: He does not need this. He should be able to place components inline with space between.

packages/scene-composer/src/components/panels/SceneRulesPanel.tsx
divya-sea: Is there a limit on the length of the property name?

divya-sea: And also you should write unit test case to test your feature with different values.

divya-sea: This needs to addressed later :)

divya-sea (approved):

PR #1860, opened 2023-09-04
feat(dashboard): triangle symbol

shlokj:
## Overview

Allows users to drag and drop a triangle from the component bar. This PR also has organizational changes for the symbol library.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/34567765/12549519-08eb-47c0-99db-55416ab5a1b5

## Legal


packages/dashboard/src/customization/widgets/triangleSymbol/component.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const fitFull = {
+    width: widthPx,
+    height: heightPx,
+  };
+
+  const halfBorderThickness = borderThickness / 2;
+
+  return (
+    <div style={fitFull}>
```
jmbuss, line 29: looks like we are sharing some common code here with other symbol widgets. What do you think about reorganizing the component structure so that we can make use of the same css classes.

packages/dashboard/src/customization/widgets/triangleSymbol/plugin.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import TriangleWidgetComponent from './component';
+import TriangleIcon from './icon';
+import type { DashboardPlugin } from '~/customization/api';
+import { TriangleWidget } from '../types';
+
+export const trianglePlugin: DashboardPlugin = {
+  install: ({ registerWidget }) => {
+    registerWidget<TriangleWidget>('triangle', {
```
jmbuss, line 9: To go along with my comment about the organization, a plugin doesn't only need to be 1 component. We could have a symbol library plugin which registers all of the symbol widgets instead of just 1.

packages/dashboard/src/customization/widgets/types.ts
```diff
@@ -110,6 +110,13 @@ export type LineProperties = {
   thickness?: number;
 };
 
+export type TriangleProperties = {
+  borderStyle?: 'solid' | 'dashed' | 'dotted';
```
jmbuss, line 114: I wonder if we can specify something a little more generic regarding symbol stylings? Are these the same properties included on rectangle and line?

packages/dashboard/src/customization/widgets/triangleSymbol/component.tsx
jmbuss: Another piece of common code might be width/height in pixels. looks like line anchor uses that calculation too.

packages/dashboard/src/customization/widgets/types.ts
shlokj: i think the line is a little different from other symbols. for example, it doesn't have a fill, and it will not have a rotation field if it is ever added to other symbols in the future. we should keep the line separate.

but i do agree that other shapes, like the rectangle, triangle, and ellipse can have the same set of properties.

packages/dashboard/src/customization/widgets/triangleSymbol/component.tsx
shlokj: i think it's a good idea to generalize all shapes (a line is not a shape for this purpose). i will create a directory for shapes within widgets

packages/dashboard/src/customization/widgets/triangleSymbol/plugin.tsx
shlokj: i don't think we can have a single plugin for all widgets, since the `WidgetRegistrationOptions` argument to `registerWidget` allows us to register only a single widget in the component palette. we could, however, put all the plugins in a single file and return them in an array, and we can use `forEach` in api.ts to register them. does that sound good?

packages/dashboard/src/customization/widgets/types.ts
jmbuss: Cool, ya I think that's a good idea!

packages/dashboard/src/customization/widgets/triangleSymbol/plugin.tsx
jmbuss: What if you called register widget multiple times? We could have something like symbol plugin that calls register widget 4+ times for: line, rectangle, ellipse, triangle, etc.

packages/dashboard/src/customization/widgets/triangleSymbol/component.tsx
jmbuss: Nice, ya I like that idea, also, it might be useful to have some more utility hooks like useSizeInPixels for instance which would do that calculation of converting grid size to pixels using cell size. I know I saw that code in multiple spots.

shlokj: sounds good. would we want to have that function/hook in the symbol directory or at a higher level? because we might end up using it somewhere else later on too - it's a nice thing to have

packages/dashboard/src/customization/widgets/triangleSymbol/plugin.tsx
shlokj: right, that's a better idea. i will try to do that

packages/dashboard/src/customization/widgets/triangleSymbol/component.tsx
shlokj: added it to the symbol library for now but it can be moved anytime

packages/dashboard/src/customization/widgets/symbolLibrary/plugin.tsx
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    registerWidget<ShapeWidget>('rectangle', {
+      render: (widget) => <RectangleWidgetComponent {...widget} />,
+      componentLibrary: {
+        name: 'Rectangle',
+        icon: RectangleIcon,
+      },
+      properties: () => ({
+        borderStyle: 'solid',
```
jmbuss, line 42: do we need to specify these here? I think they have defaults in the component

jmbuss (commented):
Looks great! I think this a really good example of the capabilities of the plugin system and how we can expand to use HMIs

Left two minor comments.

packages/dashboard/src/customization/widgets/symbolLibrary/plugin.tsx
shlokj: removed properties for rectangle and triangle entirely

PR #1874, opened 2023-09-06
feat(echarts): allow scroll left and right

corteggiano:
## Overview
- allow scrolling left/right on echarts, component will load more data as viewport changes
- better syncing of components
- improve file naming
scroll

## Legal


packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,64 @@
...
+// this function will prevent the data from moving if we are zoomed in
+const onDataZoomEvent = (chart: EChartsType, setViewport: (viewport: Viewport, lastUpdatedBy?: string) => void) => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //@ts-ignore
+  const currentDataZoom = chart.getOption().dataZoom[0];
+
+  const newDataZoom = {};
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //@ts-ignore
+  currentDataZoom.start > 0 ? (newDataZoom.startValue = currentDataZoom.startValue) : (newDataZoom.start = 0);
```
mnischay, line 19: `newDataZoom.startValue = currentDataZoom.start > 0 ? currentDataZoom.startValue : 0;`

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,64 @@
...
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //@ts-ignore
+  currentDataZoom.end < 100 ? (newDataZoom.endValue = currentDataZoom.endValue) : (newDataZoom.end = 100);
```
mnischay, line 22: same as above

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,64 @@
...
+import { DEFAULT_DATA_ZOOM } from '../eChartsConstants';
+import { ViewportInMs } from '../types';
+import { Viewport } from '@iot-app-kit/core';
+
+// known bug with zooming into live data https://github.com/apache/echarts/issues/11679
+// this function will prevent the data from moving if we are zoomed in
+const onDataZoomEvent = (chart: EChartsType, setViewport: (viewport: Viewport, lastUpdatedBy?: string) => void) => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //@ts-ignore
+  const currentDataZoom = chart.getOption().dataZoom[0];
```
mnischay, line 14: we should save the option we pass if we need to update it later, reading option from echarts is not the suggested way

mnischay, line 14: The dataTypes is provided with echarts, we shouldnt be using //ts-ignore

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,64 @@
...
+  currentDataZoom.start > 0 ? (newDataZoom.startValue = currentDataZoom.startValue) : (newDataZoom.start = 0);
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //@ts-ignore
+  currentDataZoom.end < 100 ? (newDataZoom.endValue = currentDataZoom.endValue) : (newDataZoom.end = 100);
+
+  setViewport(
+    { start: new Date(currentDataZoom.startValue), end: new Date(currentDataZoom.endValue) },
+    ECHARTS_GESTURE
+  );
+  chart.setOption(
```
mnischay, line 28: do we need to set this instantly ? can this hook return this object so all the options is set in the set options hook

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,64 @@
...
+    { lazyUpdate: true }
+  );
+};
+
+export const useDataZoom = (chartRef: MutableRefObject<EChartsType | null>, viewportInMs: ViewportInMs) => {
+  const { lastUpdatedBy, setViewport } = useViewport();
+  const [isScrolling, setIsScroling] = useState(false);
+
+  // handle live mode
+  useEffect(() => {
```
mnischay, line 41: do we need useEffect here ? given the if condition is gating the update anyway

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,64 @@
...
+        { lazyUpdate: true }
+      );
+    }
+  }, [chartRef, viewportInMs, isScrolling]);
+
+  // create listener for DataZoom events so that all charts can sync up
+  useEffect(() => {
+    const chart = chartRef.current;
+    if (chart) {
+      chart.on('dataZoom', function () {
```
mnischay, line 58: you also need to unsubscribe from this 

return () => chart.off('dataZoom', handler)

mnischay (changes requested):

packages/react-components/src/components/chart/hooks/useDataZoom.ts
corteggiano: nope, we have to set either `startValue` or just `start` depending on what the current dataZoom is

packages/react-components/src/components/chart/hooks/useDataZoom.ts
corteggiano: nope, we have to set either `endValue` or just `end` depending on what the current dataZoom is

packages/react-components/src/components/chart/hooks/useDataZoom.ts
corteggiano: so maintain a local state with the currentDataZoom ?

packages/react-components/src/components/chart/hooks/useDataZoom.ts
corteggiano: yes this needs to be done instantly as it is a user scroll event which we are overriding

packages/react-components/src/components/chart/hooks/useDataZoom.ts
corteggiano: the useEffect is so that the dataZoom is only updated if the `viewportInMs` value changes

packages/react-components/src/components/chart/hooks/useDataZoom.ts
mnischay: sure, i am not questioning the logic... when using tertiary op, it should return a value rather than assigning the value ? also, I am guessing you have ts-ignore for this ? ts-ignore should almost never be used

packages/react-components/src/components/chart/hooks/useDataZoom.ts
mnischay: Sure, if needed

mnischay: in that case probably worth to def an type / interface ?

packages/react-components/src/components/chart/hooks/useDataZoom.ts
mnischay: hmmmmm, mind taking at the useTrendCursors hook ??

packages/react-components/src/components/chart/hooks/useDataZoom.ts
mnischay: but this will get triggered when either  isScrolling and or chartRef has changed as well right ? when you remove the chart.setoptions, and datazoom is a state,, you will be setting state here... it is not the best to NOT setState inside of a useEffect, https://www.youtube.com/watch?v=bGzanfKVFeU

packages/react-components/src/components/chart/hooks/useDataZoom.ts
corteggiano: ended up not needing this at all! less code wooo

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,55 @@
...
+import { DEFAULT_DATA_ZOOM } from '../eChartsConstants';
+import { ViewportInMs } from '../types';
+import { Viewport } from '@iot-app-kit/core';
+
+// known bug with zooming into live data https://github.com/apache/echarts/issues/11679
+// this function will prevent the data from moving if we are zoomed in
+const onDataZoomEvent = (chart: EChartsType, setViewport: (viewport: Viewport, lastUpdatedBy?: string) => void) => {
+  if (chart) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    //@ts-ignore
```
mnischay, line 14: we still need this ?

packages/react-components/src/components/chart/hooks/useDataZoom.ts
mnischay: isnt this the same as locally storing the state? i.e datazoom will not change outside of the chart right ? this will be updated based on the viewport anyway ... ?

packages/react-components/src/components/chart/hooks/useDataZoom.ts
mnischay: I am very positive that this need not be called here ....

packages/react-components/src/components/chart/hooks/useDataZoom.ts
corteggiano: oh we actually dont! good call

packages/react-components/src/components/chart/hooks/useDataZoom.ts
mnischay: https://github.com/awslabs/iot-app-kit/blob/main/packages/react-components/src/components/chart/baseChart.tsx#L40

packages/react-components/src/hooks/useECharts/useGroupableEchart.ts
```diff
@@ -10,10 +11,12 @@ import type { ECharts } from 'echarts';
  * @returns void
  */
 export const useGroupableEChart = (chartRef: React.MutableRefObject<ECharts | null>, groupId?: string) => {
+  const { group } = useViewport();
```
mnischay, line 14: Curious: what is this group ? 

previously this was dedicated to "group" charts, i.e have sync mode turned ON

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -0,0 +1,54 @@
...
+        { lazyUpdate: true }
+      );
+    }
+  }, [chartRef, viewportInMs, isScrolling]);
+
+  // create listener for DataZoom events so that all charts can sync up
+  useEffect(() => {
+    const chart = chartRef?.current;
+    if (chart) {
+      chart.on('dataZoom', function () {
```
mnischay, line 44: use arrow function ?

packages/react-components/src/hooks/useECharts/useGroupableEchart.ts
corteggiano: yeah, so we're still relying on this group as the source of truth since we have a mix of synchro-charts and echarts 

once we are solely using echarts we can stop relying on the useViewport group!

PR #1875, opened 2023-09-06
feat(dashboard): add widget tile

jmbuss:
## Overview
Adds a tile overlay to line / scatter / bar / status / timeline widgets which displays a title and a remove button. The title overlay is used in the plugin so that it is configurable by any extensibility options.
image (10)
image (9)
image (8)

## Legal

### DIFF
--- a/packages/dashboard/src/components/widgets/tile/index.ts
+++ b/packages/dashboard/src/components/widgets/tile/index.ts
@@ -1,0 +1,3 @@
+import TileWidget from './tile';
+
+export default TileWidget;
--- a/packages/dashboard/src/components/widgets/tile/tile.css
+++ b/packages/dashboard/src/components/widgets/tile/tile.css
@@ -1,0 +1,22 @@
+.widget-tile {
+  position: relative;
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+}
+
+.widget-tile-body {
+  flex: 1;
+}
+
+.widget-tile-resize-handle {
+  position: absolute;
+  bottom: 0;
+  right: 4px;
+}
+
+.horizontal-divider {
+  width: 100%;
+  height: 2px;
+}
--- a/packages/dashboard/src/components/widgets/tile/tile.spec.tsx
+++ b/packages/dashboard/src/components/widgets/tile/tile.spec.tsx
@@ -1,0 +1,100 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+
+import wrapper from '@cloudscape-design/components/test-utils/dom';
+
+import { act, render } from '@testing-library/react';
+
+import { screen } from '@testing-library/dom';
+
+import WidgetTile from './tile';
+import { configureDashboardStore } from '~/store';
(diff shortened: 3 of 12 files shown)


packages/dashboard/src/components/widgets/tile/index.tsx
tracy-french, line 1: We don't have a defined convention yet, so not saying we need to make the change and is my opinion, yet I am not a fan of keeping the component in an index.tsx file. I would rather have a tile.tsx file which is exported from an index.ts file.

Depending on your setup, it can be very difficult to track the directory of every file you have open. If an IDE is not performing file/directory uniquification (meaning two tabs of index.tsx will have their directory name prepended to make the name unique, component1/index.tsx and component2/index.tsx.

It also makes fuzzy search take a few seconds longer each time. 

Little things where someone needs to go a step further to know their context than necessary.

packages/dashboard/src/components/widgets/tile/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import SpaceBetween from '@cloudscape-design/components/space-between';
+import Button from '@cloudscape-design/components/button';
+import { CancelableEventHandler, ClickDetail } from '@cloudscape-design/components/internal/events';
+import {
+  colorBorderDividerDefault,
+  borderRadiusBadge,
+  colorBackgroundContainerContent,
+} from '@cloudscape-design/design-tokens';
+import Icon from '@cloudscape-design/components/icon';
```
tracy-french, line 13: Nit: spacing and organization.

packages/dashboard/src/components/widgets/tile/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+ */
+const WidgetTile: React.FC<WidgetTileProps> = ({ children, widget, title, removeable }) => {
+  const isReadOnly = useSelector((state: DashboardState) => state.readOnly);
+
+  const isRemoveable = !isReadOnly && removeable;
+  const isResizable = !isReadOnly;
+  const headerVisible = !isReadOnly || title;
+
+  return (
+    <div
```
tracy-french, line 58: Can we include some ARIA handling? This element is too generic for screen readers to understand.

tracy-french (changes requested):

packages/dashboard/src/components/widgets/tile/index.tsx
tracy-french: This is what I am requesting changes on.

jmbuss: What aria handling do you think is required for this element? The Box cloudscape component for example does not apply any aria attributes (atleast from my inspection), and is what I was originally using until more customized styling was required on the container. I can use a different html element, but by default Box is also just a div? I don't know if any other element can be more descriptive?

packages/dashboard/src/components/widgets/tile/index.tsx
jmbuss: 👍 Fair enough, I've changed the naming convention.

tracy-french (approved):

PR #1893, opened 2023-09-11
feat(scene composer): icon picker rule changes

divya-sea:
## Overview
This contains following changes,
1. Added icon rule changes.
2. Moved customColors under properties.
3. Added unit tests.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/54058998/c4344955-ef79-4a9b-a733-2c6eb0ff4e1e

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/package.json
+++ b/packages/scene-composer/package.json
@@ -150,6 +150,7 @@
     "@react-three/postprocessing": "2.6.2",
     "@react-three/test-renderer": "^8.1.3",
     "@tanstack/react-query": "^4.29.15",
+    "@testing-library/user-event": "^14.4.3",
     "@tweenjs/tween.js": "^20.0.3",
     "3d-tiles-renderer": "0.3.16",
     "arraybuffer-to-string": "1.0.2",
--- a/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
+++ b/packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
@@ -1,7 +1,7 @@
+import { IconLookup, findIconDefinition } from '@fortawesome/fontawesome-svg-core';
 import { ThreeEvent, extend } from '@react-three/fiber';
 import React, { ReactElement, useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
 import * as THREE from 'three';
-import { IconLookup, findIconDefinition } from '@fortawesome/fontawesome-svg-core';
 
 import {
   CustomIconSvgString,
@@ -25,6 +25,7 @@ import {
   SceneResourceType,
   SelectedAnchor,
 } from '../../../../interfaces';
+import { IIconLookup } from '../../../../models/SceneModels';
 import { IAnchorComponentInternal, ISceneNodeInternal, useStore, useViewOptionState } from '../../../../store';
 import { applyDataBindingTemplate } from '../../../../utils/dataBindingTemplateUtils';
 import { findComponentByType } from '../../../../utils/nodeUtils';
@@ -33,7 +34,6 @@ import { colors } from '../../../../utils/styleUtils';
 import { Anchor } from '../../../three';
 import { WidgetSprite, WidgetVisual } from '../../../three/visuals';
 import svgIconToWidgetSprite from '../common/SvgIconToWidgetSprite';
-import { IIconLookup } from '../../../../models/SceneModels';
 
 export interface AnchorWidgetProps {
   node: ISceneNodeInternal;
@@ -44,8 +44,6 @@ export interface AnchorWidgetProps {
   ruleBasedMapId?: string;
   navLink?: INavLink;
 }
-
-type overrideCustomColorType = (value: string | undefined) => void;
 
 // Adds the custom objects to React Three Fiber
 extend({ Anchor, WidgetVisual, WidgetSprite });
@@ -95,13 +93,8 @@ export function AsyncLoadedAnchorWidget({
 
   const [_parent, setParent] = useState<THREE.Object3D | undefined>(getObject3DFromSceneNodeRef(node.parentRef));
   const [overrideCustomColor, setOverrideCustomColor] = useState<string | undefined>(undefined);
-  const selectedIconDefinition = findIconDefinition({
-    prefix: customIcon?.prefix,
-    iconName: customIcon?.iconName,
-  } as IconLookup);
-  const newCustomIcon = selectedIconDefinition?.icon[4] as string;
-  const newIconWidth = selectedIconDefinition?.icon[0];
-  const newIconHeight = selectedIconDefinition?.icon[1];
+  const [overrideCustomIcon, setOverrideCustomIcon] = useState<IIconLookup | undefined>(undefined);
+
   const baseScale = useMemo(() => {
     // NOTE: For Fixed Size value was [0.05, 0.05, 1]
     const defaultScale = autoRescale ? [0.05, 0.05, 1] : [0.5, 0.5, 1];
@@ -113,16 +106,54 @@ export function AsyncLoadedAnchorWidget({
   useEffect(() => {
     setParent(node.parentRef ? getObject3DFromSceneNodeRef(node.parentRef) : undefined);
   }, [node.parentRef]);
-
   // Evaluate visual state based on data binding
   const visualState = useMemo(() => {
-    getCustomIconColor(ruleResult, setOverrideCustomColor);
-    const ruleTargetInfo = getSceneResourceInfo(ruleResult as string);
     // Anchor widget only accepts icon, otherwise, default to Info icon
-    return ruleTargetInfo.type === SceneResourceType.Icon ? ruleTargetInfo.value : defaultIcon;
+    const ruleTargetInfo = getSceneResourceInfo(ruleResult.target as string);
+    if (ruleTargetInfo.type === SceneResourceType.Icon) {
+      // check overrideCustomColor is defined but not same as rule targetMetadata color
+      if (overrideCustomColor && overrideCustomColor !== ruleResult.targetMetadata?.color) {
+        setOverrideCustomColor(ruleResult.targetMetadata?.color);
+      }
+      // check overrideCustomColor is undefined but targetMetadata color is defined
+      if (!overrideCustomColor && ruleResult.targetMetadata?.color) {
+        setOverrideCustomColor(ruleResult.targetMetadata?.color);
+      }
+
+      // check overrideCustomIcon is defined but not same as rule targetMetadata icon
+      if (
+        overrideCustomIcon &&
+        (overrideCustomIcon.iconName !== ruleResult.targetMetadata?.iconName ||
+          overrideCustomIcon.prefix !== ruleResult.targetMetadata?.iconPrefix)
+      ) {
+        setOverrideCustomIcon({
+          prefix: ruleResult.targetMetadata?.iconPrefix,
+          iconName: ruleResult.targetMetadata?.iconName,
+        } as IIconLookup);
+      }
+
+      // check targetMetadata icon is defined
+      if (ruleResult.targetMetadata && (ruleResult.targetMetadata.iconName || ruleResult.targetMetadata.iconPrefix)) {
+        setOverrideCustomIcon({
+          prefix: ruleResult.targetMetadata?.iconPrefix,
+          iconName: ruleResult.targetMetadata?.iconName,
+        } as IIconLookup);
+      }
+      return ruleTargetInfo.value;
+    }
+    return defaultIcon;
   }, [ruleResult]);
 
   const visualRuleCustomColor = overrideCustomColor !== undefined ? overrideCustomColor : chosenColor;
+  const visualCustomIcon =
+    overrideCustomIcon !== undefined && overrideCustomIcon.iconName !== undefined ? overrideCustomIcon : customIcon;
+  const selectedIconDefinition = findIconDefinition({
+    prefix: visualCustomIcon?.prefix,
+    iconName: visualCustomIcon?.iconName,
+  } as IconLookup);
+  const newCustomIcon = selectedIconDefinition?.icon[4] as string;
+  const newIconWidth = selectedIconDefinition?.icon[0];
+  const newIconHeight = selectedIconDefinition?.icon[1];
   const defaultVisualMap = useMemo(() => {
     // NOTE: Due to the refactor from a Widget Visual (SVG to Mesh) to a Widget Sprite (SVG to Sprite)
     //  we need a new way of showing selection. This is done by showing a transparent larger image BEHIND the
@@ -149,19 +180,19 @@ export function AsyncLoadedAnchorWidget({
     return iconStrings.map((iconString, index) => {
       const iconStyle = keys[index];
       if (iconStyle === 'Custom' && (visualRuleCustomColor || newCustomIcon)) {
-        const modifiedIconStyle = visualRuleCustomColor || newCustomIcon;
+        let modifiedIconStyle = iconStyle;
+        if (visualRuleCustomColor) {
+          modifiedIconStyle = `${modifiedIconStyle}-${visualRuleCustomColor}`;
+        }
+        if (newCustomIcon) {
+          modifiedIconStyle = `${modifiedIconStyle}-${newCustomIcon}`;
+        }
         const modifiedSvg = modifySvg(iconString, visualRuleCustomColor, newCustomIcon, newIconWidth, newIconHeight);
-        return svgIconToWidgetSprite(
-          modifiedSvg,
-          iconStyle,
-          modifiedIconStyle ? `${iconStyle}-${modifiedIconStyle}` : iconStyle,
-          isAlwaysVisible,
-          !autoRescale,
-        );
+        return svgIconToWidgetSprite(modifiedSvg, iconStyle, modifiedIconStyle, isAlwaysVisible, !autoRescale);
       }
       return svgIconToWidgetSprite(iconString, iconStyle, iconStyle, isAlwaysVisible, !autoRescale);
     });
-  }, [autoRescale, CustomIconSvgString, visualRuleCustomColor, newCustomIcon]);
+  }, [autoRescale, CustomIconSvgString, visualRuleCustomColor, visualCustomIcon, newCustomIcon]);
 
   const isAnchor = (nodeRef?: string) => {
     const node = getSceneNodeByRef(nodeRef);
@@ -197,6 +228,7 @@ export function AsyncLoadedAnchorWidget({
     navLink,
     visualRuleCustomColor,
     newCustomIcon,
+    visualCustomIcon,
   ]);
 
   const onClick = useCallback(
@@ -294,19 +326,6 @@ export const AnchorWidget: React.FC<AnchorWidgetProps> = (props: AnchorWidgetPro
     </React.Suspense>
   );
 };
-
-/**
- * This method sets the custom icon color if it is returned from the rule.
- * @param ruleTarget
- * @param setOverrideCustomColor
- */
-function getCustomIconColor(ruleTarget: string | number, setOverrideCustomColor: overrideCustomColorType) {
-  const ruleColor =
-    typeof ruleTarget === 'string' && ruleTarget.includes('Custom-')
-      ? ruleTarget.split(':')[1].split('-')[1]
-      : undefined;
-  setOverrideCustomColor(ruleColor);
-}
 
 /**
  * This method parse the svg string and fill the color
--- a/packages/scene-composer/src/components/panels/SceneRulesPanel.spec.tsx
+++ b/packages/scene-composer/src/components/panels/SceneRulesPanel.spec.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
 import { render } from '@testing-library/react';
+import React from 'react';
 
 import { SceneRulesPanel } from './SceneRulesPanel';
 
@@ -35,7 +35,6 @@ jest.mock('../../store/Store', () => {
 describe('SceneRulesPanel returns expected elements.', () => {
   it('SceneRulesPanel returns expected elements.', async () => {
     const { container } = render(<SceneRulesPanel />);
-
     expect(container).toMatchSnapshot();
   });
 });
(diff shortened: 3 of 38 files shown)


sheilaXu:
from the screen recording I saw a couple of issues:
1. after you set the data binding and rule id for the tag component, it disappears for a while (~0.07 minute) until the rule has the icon target set. this is not a good experience and the tag component should just ignore the rule when it's not for icon instead of disappearing
2. the first rule expression in the rules panel displays an `info` icon, but when the rule is applied, the tag still displays the `3` icon from the default style. I know you display the `info` icon as default value for icon picker, but it appears to me like the icon from the rule is not applied to the tag. I suggest to not display default icon in the picker to avoid confusion

packages/scene-composer/src/utils/dataBindingUtils.ts
```diff
@@ -160,25 +160,22 @@ export const ruleEvaluator = (
...
+        return { target: statement.target, targetMetadata: statement.targetMetadata };
       }
     } catch (error) {
       // It might be a user mistake for one rule, but the next rule may work, we should continue
       LOG.error('Failed to evaluate rule', error as Error);
     }
   }
-
-  return defaultState;
+  return { target: defaultState, targetMetadata: undefined };
```
sheilaXu, line 180: can you add a return type to this function so the consumers know clearly the type of the value returned? right now I can see 2 returns `{ target: xxx, targetMetadata: yyy}` but the rest return different types. it will be hard to handle the results in this way.
also instead of `{ target: xxx, targetMetadata: yyy}`, I suggest `{ result: xxx, targetMetadata: yyy}` so it makes more sense for cases returning defaultState or number.

packages/scene-composer/src/interfaces/dataTypes.ts
```diff
@@ -34,6 +34,11 @@ export interface ITransformConstraint {
 }
 
 export type { IValueDataBinding, ITwinMakerEntityDataBindingContext } from '@iot-app-kit/source-iottwinmaker';
+export type TargetMetadata<T> = Record<string, T> & {
+  color: string;
+  iconPrefix: string;
+  iconName: string;
```
sheilaXu, line 40: I guess they should all be optional

packages/scene-composer/src/interfaces/dataTypes.ts
```diff
@@ -34,6 +34,11 @@ export interface ITransformConstraint {
 }
 
 export type { IValueDataBinding, ITwinMakerEntityDataBindingContext } from '@iot-app-kit/source-iottwinmaker';
+export type TargetMetadata<T> = Record<string, T> & {
```
sheilaXu, line 37: maybe remove `Record<string, T> &`  part and make this type more specific, so we have better typing support to avoid typos. 
I've seen problem with several other fields typed as `Record<string, xxx>` that we can hardly know the values being stored. so I think we should start to type with the clear keys and extend in the future when we need more keys, instead of starting with a generic map type

packages/scene-composer/src/components/panels/scene-rule-components/SceneRuleTargetEditor.tsx
```diff
@@ -92,23 +102,47 @@ export const SceneRuleTargetEditor: React.FC<ISceneRuleTargetEditorProps> = ({
...
+                }}
+                selectedIcon={icon as IconLookup}
+                iconPickerLabel={formatMessage({ defaultMessage: 'Icon', description: 'Icon' })}
+                iconFilterText={formatMessage({ defaultMessage: 'Find icons', description: 'Find icons' })}
+                iconFilterTextAriaLabel={formatMessage({
+                  defaultMessage: 'Filter icons',
+                  description: 'Filter icons',
+                })}
+              />
+            </>
```
sheilaXu, line 145: all of those are specific to icon target, so they should be in `SceneRuleTargetIconEditor` component instead of here. and all states related to chosenColor and customIcon should all be move into `SceneRuleTargetIconEditor`

packages/scene-composer/src/components/panels/scene-components/tag-style/IconPicker/IconPicker.tsx
```diff
@@ -108,14 +107,14 @@ export const IconPicker = ({
                 />
                 <div style={IconGridLayout}>
                   {filteredIcons.map((icon, index) => {
-                    return <FontAwesomeIcon key={index} icon={icon} onClick={() => handleIconClick(icon)} />;
+                    return <FontAwesomeIcon key={index + 1} icon={icon} onClick={() => handleIconClick(icon)} />;
```
sheilaXu, line 110: why is `+ 1`?

packages/scene-composer/src/components/panels/SceneRulesPanel.spec.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
+import wrapper from '@awsui/components-react/test-utils/dom';
```
sheilaXu, line 3: are there unused imports?

packages/scene-composer/src/components/panels/scene-components/AnchorComponentEditor.tsx
```diff
@@ -216,8 +219,11 @@ export const AnchorComponentEditor: React.FC<IAnchorComponentEditorProps> = ({
                   chosenColor: pickedColor,
                 });
               }}
-              onUpdateCustomColors={(chosenCustomColors) => onUpdateCallback({ customColors: chosenCustomColors })}
-              customColors={anchorComponent.customColors}
+              onUpdateCustomColors={(chosenCustomColors) =>
+                setSceneProperty(KnownSceneProperty.TagCustomColors, chosenCustomColors)
+              }
+              // onUpdateCallback({ customColors: chosenCustomColors })}
```
sheilaXu, line 225: clean up?

packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
```diff
@@ -113,16 +106,57 @@ export function AsyncLoadedAnchorWidget({
...
+      // check overrideCustomColor is undefined but targetMetadata color is defined
+      if (!overrideCustomColor && ruleResult.targetMetadata?.color) {
+        setOverrideCustomColor(ruleResult.targetMetadata?.color);
+      }
+
+      // check overrideCustomIcon is defined but not same as rule targetMetadata icon
+      if (
+        overrideCustomIcon &&
+        overrideCustomIcon.iconName !== ruleResult.targetMetadata?.iconName &&
+        overrideCustomIcon.prefix !== ruleResult.targetMetadata?.iconPrefix
```
sheilaXu, line 127: this should be `||` because icon prefix can be unchanged, and only icon name is changed. we should also update in this case

packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
```diff
@@ -113,16 +106,57 @@ export function AsyncLoadedAnchorWidget({
   useEffect(() => {
     setParent(node.parentRef ? getObject3DFromSceneNodeRef(node.parentRef) : undefined);
   }, [node.parentRef]);
-
   // Evaluate visual state based on data binding
   const visualState = useMemo(() => {
-    getCustomIconColor(ruleResult, setOverrideCustomColor);
-    const ruleTargetInfo = getSceneResourceInfo(ruleResult as string);
+    const checkTargetExists = typeof ruleResult === 'object' && 'target' in ruleResult;
+    const checkTargetMetadataExists = typeof ruleResult === 'object' && 'targetMetadata' in ruleResult;
```
sheilaXu, line 112: with the property return type defined for evaluateRule function, we don't need those check

packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
```diff
@@ -113,16 +106,57 @@ export function AsyncLoadedAnchorWidget({
...
+        overrideCustomIcon.prefix !== ruleResult.targetMetadata?.iconPrefix
+      ) {
+        setOverrideCustomIcon({
+          prefix: ruleResult.targetMetadata?.iconPrefix,
+          iconName: ruleResult.targetMetadata?.iconName,
+        } as IconLookup);
+      }
+
+      // check overrideCustomIcon is undefined but targetMetadata icon is defined
+      if (ruleResult.targetMetadata && ruleResult.targetMetadata.iconName && ruleResult.targetMetadata.iconPrefix) {
```
sheilaXu, line 136: from the comment, you also want condition `!overrideCustomIcon`?

packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
```diff
@@ -113,16 +106,57 @@ export function AsyncLoadedAnchorWidget({
...
+          prefix: ruleResult.targetMetadata?.iconPrefix,
+          iconName: ruleResult.targetMetadata?.iconName,
+        } as IconLookup);
+      }
+    }
+    const ruleTargetInfo = getSceneResourceInfo(
+      checkTargetExists ? (ruleResult.target as string) : (ruleResult as string),
+    );
     // Anchor widget only accepts icon, otherwise, default to Info icon
     return ruleTargetInfo.type === SceneResourceType.Icon ? ruleTargetInfo.value : defaultIcon;
```
sheilaXu, line 147: all of above logic should run only when `ruleTargetInfo.type === SceneResourceType.Icon` is true. otherwise return default icon

packages/scene-composer/src/interfaces/dataTypes.ts
divya-sea: In our case T is color,icon and iconName and in the future if you all together want to create TargetMetadata of different type then this record still holds good. I guess we can keep as it is but internally color, icon and iconName we can specify a type. I would like to keep this generic type as it is but T type can be specified internally.

packages/scene-composer/src/components/panels/scene-rule-components/SceneRuleTargetEditor.tsx
divya-sea: They are specific to Icon and logic is within it. There are two reasons why we need to add them here. 
1. Because of the current grid pattern. There is an outer most grid for targets.
2. There is grid within target components itself
If we have move them the inside icon component then we will have change the UX design as there is a outer grid structure which set the icon component to render from colSpan 4 If you add them part of the icon component. 
I would like to keep as it is for this release unless there is an urgent change in UI requirement.

packages/scene-composer/src/utils/dataBindingUtils.ts
divya-sea: There are total 5 return results here. We can return targetMetadata if the type of target is icon. But outermost return will return the default target state if nothing is matched but I just added undefined because we need this to evaluate visual state.

packages/scene-composer/src/augmentations/components/three-fiber/anchor/AnchorWidget.tsx
divya-sea: yes. ad also filtered the icons in anchor component as well since we have support only for icon.

packages/scene-composer/src/components/panels/scene-components/AnchorComponentEditor.tsx
divya-sea: done

packages/scene-composer/src/components/panels/SceneRulesPanel.spec.tsx
divya-sea: done

packages/scene-composer/src/utils/dataBindingUtils.ts
divya-sea: Returning a RuleReturnResult type.

packages/scene-composer/src/components/panels/SceneRulesPanel.spec.tsx
sheilaXu: this doesn't seem to be addressed

packages/scene-composer/package.json
```diff
@@ -150,6 +150,7 @@
     "@react-three/postprocessing": "2.6.2",
     "@react-three/test-renderer": "^8.1.3",
     "@tanstack/react-query": "^4.29.15",
+    "@testing-library/user-event": "^14.4.3",
```
sheilaXu, line 153: this new dependency doesn't seem to be used anywhere. please remove it if so

packages/scene-composer/src/components/panels/scene-components/AnchorComponentEditor.tsx
sheilaXu: change not committed here as well

packages/scene-composer/src/components/panels/scene-components/tag-style/ColorPicker/ColorPicker.tsx
```diff
@@ -70,12 +72,15 @@ export const ColorPicker = ({
...
-    if (!Object.values(palleteColors).includes(color)) {
-      setCustomInternalColors(customInternalColors.concat(color));
-      onUpdateCustomColors?.([...new Set(customInternalColors)]);
-    }
-  };
+  const checkIfCustomColor = useCallback(
+    (color): void => {
+      if (!Object.values(palleteColors).includes(color) && color !== colors.infoBlue) {
+        setCustomInternalColors(customInternalColors.concat(color));
+        onUpdateCustomColors?.([...new Set(customInternalColors)]);
```
sheilaXu, line 79: `customInternalColors` here is not updated to have the new color yet since setState is async. you should have a local array with the new value

packages/scene-composer/src/hooks/useRuleResult.ts
```diff
@@ -1,13 +1,18 @@
...
 import { useSceneDocument, useDataStore } from '../store';
 import { useSceneComposerId } from '../common/sceneComposerIdContext';
 import { dataBindingValuesProvider, ruleEvaluator } from '../utils/dataBindingUtils';
 
 import useBindingData from './useBindingData';
 
+export interface RuleReturnResult {
+  target: string | number;
+  targetMetadata?: TargetMetadata<string>;
+}
```
sheilaXu, line 14: this type should be defined in dataBindingUtils file because `useRuleResult` imports from `dataBindingUtils`. if you add it here, `dataBindingUtils` imports from `useRuleResult` and creates the circular dependency

packages/scene-composer/src/components/three-fiber/MotionIndicatorComponent/__tests__/__snapshots__/MotionIndicatorComponentSnap.spec.tsx.snap
```diff
@@ -89,7 +89,7 @@ exports[`MotionIndicatorComponent should render with correct speed and values fr
     <div
       data-testid="linear-plane"
     >
-      [{"speed":3,"foregroundColor":0,"backgroundColor":0,"config":{"numOfRepeatInY":2,"defaultForegroundColor":"red","defaultBackgroundColor":"blue","defaultSpeed":6,"backgroundColorOpacity":1},"scale":[1,2,3]},{}]
+      [{"speed":0,"foregroundColor":0,"backgroundColor":0,"config":{"numOfRepeatInY":2,"defaultForegroundColor":"red","defaultBackgroundColor":"blue","defaultSpeed":6,"backgroundColorOpacity":1},"scale":[1,2,3]},{}]
```
sheilaXu, line 92: why is this changed? something seems to be wrong

packages/scene-composer/src/interfaces/dataTypes.ts
sheilaXu: I don't think a set of metadata will always have the same type, so the generic type here will not be super useful here. and also if we need more kinds of metadata, we should defined in here to make it more clear. so the generic `Record<>` part will not be useful as well

packages/scene-composer/src/components/panels/scene-rule-components/SceneRuleTargetIconEditor.tsx
```diff
@@ -38,38 +42,43 @@ export const SceneRuleTargetIconEditor: React.FC<ISceneRuleTargetIconEditorProps
...
-            'Specifies the localized string that describes an option as being selected. This is required to provide a good screen reader experience',
-        })}
-      />
-      {isCustomStyle ? (
-        <DecodeSvgString
-          selectedColor={chosenColor ?? colors.customBlue}
-          iconString={iconString}
-          width='32px'
-          height='32px'
+    <>
```
sheilaXu, line 48: this layer doesn't seem to be needed

packages/scene-composer/src/components/panels/scene-rule-components/SceneRuleTargetEditor.tsx
```diff
@@ -92,23 +102,47 @@ export const SceneRuleTargetEditor: React.FC<ISceneRuleTargetEditorProps> = ({
...
+              <IconPicker
+                onSelectIconChange={(pickedIcon) => {
+                  onChange(convertToIotTwinMakerNamespace(targetInfo.type, targetInfo.value), {
+                    ...targetMetadata,
+                    iconPrefix: pickedIcon.prefix,
+                    iconName: pickedIcon.iconName,
+                  } as TargetMetadata<string>);
+                  setIcon(pickedIcon as IconLookup);
+                }}
+                selectedIcon={icon as IconLookup}
```
sheilaXu, line 137: do you need to cast to `IconLookup` here? can we use `IIconLookup` for any internal component, then only cast to `IconLookup` when passing it to fontawsome?

packages/scene-composer/package.json
divya-sea: We can use it in the future instead of fireEvent. This is better version of fireEvent.

packages/scene-composer/src/interfaces/dataTypes.ts
divya-sea: You meant to they will have the same type? If it is not of same then, T holds good. All you have to do is 
```
color?: string;
 iconPrefix?: string;
 iconName?: string;
opacity?: Object;
meta?: string[];
```
If we string just to Record<string, string>, the above case does not work. That time again we will have to create a new type.

packages/scene-composer/src/hooks/useRuleResult.ts
divya-sea: Yeah I wanted to know what is the best place to declare this type.

packages/scene-composer/src/components/three-fiber/MotionIndicatorComponent/__tests__/__snapshots__/MotionIndicatorComponentSnap.spec.tsx.snap
divya-sea: I can run them again.check what changed this.

packages/scene-composer/src/components/panels/scene-rule-components/SceneRuleTargetIconEditor.tsx
divya-sea: yes We don't need this.

packages/scene-composer/src/components/panels/scene-rule-components/SceneRuleTargetEditor.tsx
divya-sea: No we don't

packages/scene-composer/package.json
sheilaXu: we can add it in the future when it's actually in use. I don't see a reason to add a library in advance unless you know you will use it in next change

divya-sea: we should not be using firevent actually. It is out dated. We should be using this and replacing fireevent.

packages/scene-composer/src/components/three-fiber/MotionIndicatorComponent/MotionIndicatorComponent.tsx
```diff
@@ -72,8 +72,10 @@ const MotionIndicatorComponent: React.FC<IMotionIndicatorComponentProps> = ({
         dataBindingTemplate,
       );
 
-    const result = ruleEvaluator(0, values, speedRule) as number;
-    return result >= 0 ? result : 0;
+    const result = ruleEvaluator(0, values, speedRule);
```
sheilaXu, line 75: shouldn't it be `ruleEvaluator(0, values, speedRule).target`?

packages/scene-composer/src/interfaces/dataTypes.ts
sheilaXu: so whenever you add new metadata, you will add the field explicitly here. there isn't a case we will use `Record<string, T>`. it just make us unsure about the exact metadata can be in the metadata

packages/scene-composer/src/components/three-fiber/MotionIndicatorComponent/MotionIndicatorComponent.tsx
divya-sea: We are returning returning the RuleReturnResult and there is only single value in it. This is also one way of getting the returned result?

packages/scene-composer/src/interfaces/dataTypes.ts
sheilaXu: I've recently changed some types defined as `Record<xx, yy>` to be explicit interfaces, because I'm seeing problem for us to know the data stored without searching in the entire code base

packages/scene-composer/src/components/three-fiber/MotionIndicatorComponent/MotionIndicatorComponent.tsx
sheilaXu: but `ruleEvaluator` returns `{ target: xx, targetMetadata: yy}`, why would `typeof result === 'number'` be true?

divya-sea: I removed targetMetadata from this. That is not required here. If (evalResult) type is number  we are returning Number(evalResult) so here in this rule evaluator speed is a number.

sheilaXu (approved):

PR #1898, opened 2023-09-11
feat(first Person): proof of using pointer lock for a first person view

haweston:
## Overview
Add an option for a first person camera view and navigation mode using a pointer lock model.

## Verifying Changes
Storybook
select the First Person story.  This should auto set the First Person feature flag.
Your camera controls should have a third option added to them of 'Pointer Lock' that's in addition to the existing orbit and pan modes.

Selecting this button and then using the mouse to click the canvas will put you into pointer lock mode.  to leave the mode click Esc like the pop-up message in Chrome mentions.

Once in pointer lock mode navigation works as:
Rotation camera view: left mouse click and drag mouse
move forward: W key or Up arrow key
move backward: S key or down arrow key
move left: A key or left arrow key
move right: D key or right arrow key
move down: V key
move up: space bar

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/EditorCamera.tsx
+++ b/packages/scene-composer/src/components/three-fiber/EditorCamera.tsx
@@ -9,7 +9,7 @@ import useLogger from '../../logger/react-logger/hooks/useLogger';
 import {
   Layers,
   ROOT_OBJECT_3D_NAME,
-  DEFAULT_CAMERA_CONTROLS_OPTIONS,
+  DEFAULT_ORBIT_CAMERA_CONTROLS_OPTIONS,
   DEFAULT_CAMERA_POSITION,
   DEFAULT_CAMERA_TARGET,
   DEFAULT_TWEEN_DURATION,
@@ -22,8 +22,11 @@ import { CameraControlImpl, TweenValueObject } from '../../store/internalInterfa
 import useActiveCamera from '../../hooks/useActiveCamera';
 import { getSafeBoundingBox } from '../../utils/objectThreeUtils';
 import { getMatterportSdk } from '../../common/GlobalSettings';
+import { MapControls as MapControlsImpl, OrbitControls as OrbitControlsImpl } from '../../three/OrbitControls';
+import { PointerLockControls as PointerLockControlsImpl } from '../../three/PointerLockControls';
+import { isOrbitOrPanControl } from '../../utils/controlUtils';
 
-import { MapControls, OrbitControls } from './controls';
+import { MapControls, OrbitControls, PointerLockControls } from './controls';
 
 export const EditorMainCamera = forwardRef<Camera>((_, forwardedRef) => {
   const log = useLogger('EditorMainCamera');
@@ -34,7 +37,8 @@ export const EditorMainCamera = forwardRef<Camera>((_, forwardedRef) => {
   const matterportSdk = getMatterportSdk(sceneComposerId);
   const scene = useThree((state) => state.scene);
   const setThree = useThree((state) => state.set);
-  const makeDefault = cameraControlsType === 'orbit' || cameraControlsType === 'pan';
+  const makeDefault =
+    cameraControlsType === 'orbit' || cameraControlsType === 'pan' || cameraControlsType === 'pointerLock';
 
   const cameraControlsImplRef = useRef<CameraControlImpl>();
   const cameraRef = useRef<Camera>();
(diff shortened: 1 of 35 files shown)


packages/scene-composer/src/components/three-fiber/EditorCamera.tsx
```diff
@@ -243,7 +269,12 @@ export function findBestViewingPosition(
     Number.isFinite(objectBoundingBox.max.z)
   ) {
     // normal case
-    const camera = controls.object as THREE.PerspectiveCamera;
+    let camera;
```
mumanity, line 272: Nit-pick: This would be a great opportunity to use a ternary... unless you need it to have a third case of undefined. 

```
const camera = controls instanceof OrbitControlsImpl || controls instanceof MapControlsImpl
  ? controls.object as THREE. PerspectiveCamera
  : controls.camera

File: packages/scene-composer/src/components/three-fiber/controls/PointerLockControls.tsx [hunk 2]
```diff
@@ -0,0 +1,189 @@
...
+
+        controls.moveRight(-newVelocity.x * delta);
+        controls.moveForward(-newVelocity.z * delta);
+        controls.moveUp(-newVelocity.y * delta);
+
+        velocity.current = newVelocity;
+      }
+    });
+
+    const onKeyDown = useCallback((event: KeyboardEvent) => {
```
mumanity, line 72: Is there a reason that this isn't setup as a reusable util? We need to make sure that any keyboard management we're introducing is consistent & has no collisions with other features.

packages/scene-composer/src/components/three-fiber/EditorCamera.tsx
haweston: Undefined is already guarded by the if above so I'll swap to a ternary.

packages/scene-composer/src/components/three-fiber/controls/PointerLockControls.tsx
haweston: It's kind of straight port of the example code from three-stblib and drei.... which is what we also did for the other camera controls. We had a design doc a while back about doing a key binding control utility like your talking about but it didn't get implemented yet.

packages/scene-composer/src/utils/controlUtils.ts
```diff
@@ -0,0 +1,36 @@
+import { CameraControlsType } from '../interfaces';
+
+export const isPointerLockControl = (controlType: CameraControlsType): boolean => {
+  if (controlType === 'pointerLock') {
```
mumanity, line 4: i <3 ternary
```
return controlType === 'pointerLock'
 ? true
 : false

Comment (haweston) [reply packages/scene-composer/src/utils/controlUtils.ts:4]: Okay converting this and the below to one line ternary

File: packages/scene-composer/src/components/three-fiber/controls/PointerLockControls.tsx (hunk 2 above)
Comment (haweston) [reply packages/scene-composer/src/components/three-fiber/controls/PointerLockControls.tsx]: Done a basic stripped done version of the the key mapping that we'll revisit later.

File: packages/scene-composer/src/hooks/useKeyPress.ts [hunk 4]
```diff
@@ -0,0 +1,92 @@
+/* istanbul ignore file */
+// Credit original Source: https://codesandbox.io/s/rob7b8?file=/src/hooks/useKeyPress.tsx:0-2069
+// Heavily modified for my needs
```
mumanity, line 3: Nit-pick: Probably don't want to keep this line as-is in the comments. Suggest clarifying what "my needs" are instead.

packages/scene-composer/src/hooks/useKeyPress.ts
```diff
@@ -0,0 +1,92 @@
...
+
+import { Key } from './useKeyMap';
+
+function isKey(key: string | number | Key): key is Key {
+  return !!(key as Key).key;
+}
+
+interface KeyPressArgs {
+  enabled?: boolean;
+  onKeyPressed?: (delta?: number) => void; // called every frame while key is held down.
```
mumanity, line 15: Nit-pick: Probably don't need this level of commenting as it is standard javascript behaviors for these actions.

packages/scene-composer/tests/components/three-fiber/EditorCamera.spec.tsx
```diff
@@ -70,6 +79,11 @@ import { EditorMainCamera, findBestViewingPosition } from '../../../src/componen
 import { useStore } from '../../../src/store';
 
 import { useThree } from '@react-three/fiber';
+
+import { OrbitControls } from '../../../src/three/OrbitControls';
+//auto mock for instanceof check passing
+//jest.mock('../../../src/three/OrbitControls');
```
mumanity, line 85: Nit-pick: Should probably just delete unless there's a TODO associated with these. If that's the case, would like to see the TODO comment inline with these.

mumanity (approved):

packages/scene-composer/src/hooks/useKeyPress.ts
haweston: Removed

packages/scene-composer/src/hooks/useKeyPress.ts
haweston: removed

packages/scene-composer/tests/components/three-fiber/EditorCamera.spec.tsx
haweston: Removed

PR #1901, opened 2023-09-12
feat(react-components): adding TrendCursor Sync to dashboard

mnischay:
## Overview

- Adding TrendCursorSync component to dashbaord.
- cleaning up some of the old code / approach

## Legal

https://github.com/awslabs/iot-app-kit/assets/135036199/7d5ac3f5-855d-4e04-adfa-1bb6895a0576


packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -65,7 +68,7 @@ const BaseChart = ({ viewport, queries, size = { width: 500, height: 500 }, ...o
     useResizeableEChart(chartRef, size);
 
   // apply group to echarts
-  useGroupableEChart(chartRef, options.groupId);
+  useGroupableEChart(chartRef, group);
```
corteggiano, line 71: `group ?? options.groupId` in case the viewport group is nonexistent?

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -95,7 +98,7 @@ const BaseChart = ({ viewport, queries, size = { width: 500, height: 500 }, ...o
     series,
     chartId,
     viewportInMs,
-    groupId: options.groupId,
+    groupId: group,
```
corteggiano, line 101: same as above ^

packages/react-components/src/components/chart/baseChart.tsx
mnischay: groupId will not be part of options i.e props anymore. if viewport does not exist, we are simply not in sync mode. and the undefined value of the group is handled in the hook

mnischay (commented):
The test fail is from scene-composer i.e TM team

`Failed:    @iot-app-kit/scene-composer#lint`

corteggiano (approved):

Issue #1872, opened 2023-09-03
Update table widget - Add fixed height and width for the table widget


nadeswal:
Current state: When a user adds a table widget, the table is barely visible and the third column is hidden because of it's size. The outline of the table widget is bigger than then the table inside it and that needs to be fixed.

Preferred UX: The table widget should have a fixed width: 100px and height: 60 px  when the use drag and drops the widget. i.e initial size (we can play with the size and pick one that suits better) that is filled throughout it's selection boundaries. 

Let's use the empty state design provided by the cloudscape. Link: [https://cloudscape.design/components/table/?tabId=playground&example=empty-state](url) Again, let's not add search box.

Change the "no resource" text to "No data to display"
Change the "create resource" text in the button to "add asset properties"

diehbria:
issue has been resolved

PR #1904, opened 2023-09-13
Add fixed height and width for the table widget #1872

Priyanshu44:
## Overview
Please note in this CR we have just added the empty state design whereas width & height fix will be taken care in another CR .

## Verifying Changes

image


Priyanshu44:
Wrong commit

PR #1905, opened 2023-09-13
feat: add fixed height and width for the table widget

Priyanshu44:
Added empty table state design with "No data to display" content.

image

Please note we have just added empty state design for the table whereas width & height will be taken care in another CR.

### DIFF
--- a/packages/react-components/src/components/table/table.tsx
+++ b/packages/react-components/src/components/table/table.tsx
@@ -36,7 +36,7 @@ export const Table = ({
   significantDigits?: number;
   paginationEnabled?: boolean;
   pageSize?: number;
-} & Pick<TableBaseProps, 'resizableColumns' | 'sortingDisabled' | 'stickyHeader'>) => {
+} & Pick<TableBaseProps, 'resizableColumns' | 'sortingDisabled' | 'stickyHeader' | 'empty'>) => {
   const { dataStreams, thresholds: queryThresholds } = useTimeSeriesData({
     viewport: passedInViewport,
     queries,
--- a/packages/dashboard/src/customization/widgets/table/component.tsx
+++ b/packages/dashboard/src/customization/widgets/table/component.tsx
@@ -3,6 +3,8 @@ import { useSelector } from 'react-redux';
 
 import { Table, TableColumnDefinition } from '@iot-app-kit/react-components';
 
+import EmptyTableComponent from './emptyTableComponent';
+
 import { computeQueryConfigKey } from '../utils/computeQueryConfigKey';
 import type { DashboardState } from '~/store/state';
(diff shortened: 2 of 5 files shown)


packages/dashboard/src/customization/widgets/table/emptyTableState.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Box, Button, SpaceBetween } from '@cloudscape-design/components';
+import type { FunctionComponent } from 'react';
+
+const EmptyTableState: FunctionComponent = () => {
```
mnischay, line 5: This is not a "state" so a better name would be "EmptyTable" or even "EmptyTableComponent"

packages/dashboard/src/customization/widgets/table/emptyTableState.tsx
```diff
@@ -0,0 +1,16 @@
...
+  return (
+    <Box data-testid='emptyStateTableDisplay' margin={{ vertical: 'xs' }} textAlign='center' color='inherit'>
+      <SpaceBetween size='m'>
+        <b data-testid='default-msg'>No data to display</b>
+        <Button>Add asset properties</Button>
```
mnischay, line 10: i guess, we discussed about this, no button required here

packages/react-components/src/components/table/table.tsx
```diff
@@ -32,7 +32,8 @@ export const Table = ({
   viewport?: Viewport;
   styles?: StyleSettingsMap;
   significantDigits?: number;
-} & Pick<TableBaseProps, 'resizableColumns'>) => {
+} & Pick<TableBaseProps, 'resizableColumns'> &
```
mnischay, line 35: I think you can do `Pick<Props, 'email' | 'password'>`

mnischay (changes requested):

packages/dashboard/src/customization/widgets/table/emptyTableState.tsx
Priyanshu44: resolved

packages/dashboard/src/customization/widgets/table/emptyTableState.tsx
Priyanshu44: I will confirm with Navita and will update accordingly

Priyanshu44: Resolved

corteggiano (approved):

mnischay (approved):

corteggiano:
hello, please run `npm run fix` at the root to fix the lint errors so we can get this change in!

PR #1908, opened 2023-09-13
feat(dashboard): update collapsible side panels

hwandersman:
## Overview
Updated the side panels to expand only when clicking the panel icon and collapse when clicking the arrow button. 

Stories:
* As a user, I can collapse the resource explorer and configuration panel by clicking the collapse arrow at the top of the panel.
* As a user, I don't want the panels to collapse and expand based on the selection state.

## Verifying Changes
Tested with storybook, see the video below.

https://github.com/awslabs/iot-app-kit/assets/87385528/416b4d21-64a7-47bd-b503-3c28001a90a8

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
+++ b/packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
@@ -1,19 +1,95 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
 
 import './index.css';
-import { colorBackgroundButtonPrimaryDefault, spaceContainerHorizontal } from '@cloudscape-design/design-tokens';
+import {
+  colorBackgroundButtonPrimaryDefault,
+  colorBorderDividerDefault,
+  spaceContainerHorizontal,
+  spaceStaticXxxs,
+} from '@cloudscape-design/design-tokens';
+import Box, { BoxProps } from '@cloudscape-design/components/box';
+import Button from '@cloudscape-design/components/button';
+import { IconProps } from '@cloudscape-design/components/icon';
+import { DEFAULT_COLLAPSED_SIDE_PANE_WIDTH } from '../resizablePanes/constants';
 
-export function CollapsiblePanel(props: { icon: string; dataCy: string }) {
-  return (
+type CollapsiblePanelProps = {
+  isPanelCollapsed: boolean;
+  panelWidth: number;
+  onCollapsedPanelClick: () => void;
+  panelContent: ReactNode;
+  icon: string;
+  side: 'right' | 'left';
+  headerText: string;
+};
+
+export function CollapsiblePanel(props: CollapsiblePanelProps) {
+  // Used to set the borderRight or borderLeft prop of the pane
+  const borderSide = props.side === 'right' ? 'Left' : 'Right';
+
+  // Used to set the side of the icon in the top level box
+  const iconSide = borderSide.toLowerCase() as BoxProps.Float;
+
+  // Used to set the iconName of the collapsible button
+  const iconName = `angle-${props.side}` as IconProps.Name;
+
+  const Divider = () => (
+    <div className='collapsible-panel-vertical-divider' style={{ backgroundColor: colorBorderDividerDefault }} />
+  );
+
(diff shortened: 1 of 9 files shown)


mnischay (commented):
ignore this

packages/dashboard/src/components/internalDashboard/index.css
```diff
@@ -51,4 +51,36 @@
   border-radius: 50%;
   width: 40px;
   height: 40px;
+}
+
+.collapsible-panel-vertical-divider {
+  width: 2px;
```
mnischay, line 57: can we utilize cloudscape spacing tokens ?

mnischay (changes requested):
Code changes LGTM, but I doubt we want this behaviour, can you please check with Navita ?

packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
```diff
@@ -1,19 +1,95 @@
...
+  onCollapsedPanelClick: () => void;
+  panelContent: ReactNode;
+  icon: string;
+  side: 'right' | 'left';
+  headerText: string;
+};
+
+export function CollapsiblePanel(props: CollapsiblePanelProps) {
+  // Used to set the borderRight or borderLeft prop of the pane
+  const borderSide = props.side === 'right' ? 'Left' : 'Right';
```
jmbuss, line 27: if we made the borderSide lowercase would we need to cast the iconside?

packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
```diff
@@ -1,19 +1,95 @@
...
+} from '@cloudscape-design/design-tokens';
+import Box, { BoxProps } from '@cloudscape-design/components/box';
+import Button from '@cloudscape-design/components/button';
+import { IconProps } from '@cloudscape-design/components/icon';
+import { DEFAULT_COLLAPSED_SIDE_PANE_WIDTH } from '../resizablePanes/constants';
 
-export function CollapsiblePanel(props: { icon: string; dataCy: string }) {
-  return (
+type CollapsiblePanelProps = {
+  isPanelCollapsed: boolean;
```
jmbuss, line 16: is it possible to track the collapsed state in this component?

packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
hwandersman: `borderSide` has to better capitalized because the value it's used for must be `borderRight` or `borderLeft`. All lower case doesn't work. I'm just reusing the fact that this is the flipped side for the `iconSide` variable.

jmbuss (commented):
Nice work! left a couple comments which aren't really blockers.

packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
hwandersman: What do you mean? Like have this component manage the panel collapsed state? 
The resizablePanes component manages the left and right collapsible panels. It needs to manage this state for its own styling, which is then passed to the collapsible panel components.

jmbuss:
> Code changes LGTM, but I doubt we want this behaviour, can you please check with Navita ?

We did sync with UX on this, it is the best path forward to fix a nasty usability bug.

packages/dashboard/src/components/internalDashboard/index.css
hwandersman: I don't believe there are any divider components in cloudscape. Fitting it in to match the mock is difficult with the default padding sizes of SpaceBetween or Box.

packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
jmbuss: Oh, right, I forgot that special styling exists in the root component.

jmbuss (approved):

PR #1909, opened 2023-09-13
feat(composer): add a11y color picker

jacksonvogt:
## Overview
New, accessible color picker built on top of [react-colorful](https://omgovich.github.io/react-colorful/). 

Level access (the a11y auditor) found 10 a11y issues with the current color picker from [react-color](https://casesandberg.github.io/react-color/), and they were unable to be remediated. 

I've added this to all four instances of the color picker in the scene composer: lights, motion indicator, rules, and custom tags. 

Notes
- The save/cancel flow is a little awkward for lights, since there is no swatch to control expanding/contracting the color picker. I think that we should add a swatch in the lights panel to unify the color picker experience
- The available width is smaller in rules which stacks some of the options
- As discussed with Brenton, this component should pop out and not be in the page flow, which it currently does not do. But since it is accessible as is, I'd like to leave that as a backlog task. Thoughts?

Check out a ~1 min demo video here: https://drive.corp.amazon.com/documents/jackvogt@/a11y%20remediation/Color%20Picker/Color%20Picker%20demo.mov.

I've tested manually in storybook and connected locally to the console, and have 96% unit test coverage.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/package.json
```diff
@@ -159,6 +159,7 @@
     "path-browserify": "1.0.1",
     "postprocessing": "6.28.5",
     "react-color": "^2.19.3",
```
divya-sea, line 161: If we are not using react-color which is lacking the a11ly. Then better to remove from package.json?

jacksonvogt: I left the CirclePicker from react-color in tag-style/ColorPicker/ColorPicker.tsx since it is accessible via keyboard. I just changed the 'custom color' selection from the SketchPicker to this custom one.

divya-sea: I see in this new color library, we would have to explicitly click on the color picker to save the selected color. We want to customer to just click on the color and that should reflect as chosenColor.  Is it possible to make save optional?

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
```diff
@@ -0,0 +1,168 @@
...
+    } else {
+      hexInputFieldRef.current?.focus();
+    }
+  };
+
+  return (
+    <ColorPickerWrapper>
+      <form onSubmit={(e) => e.preventDefault()}>
+        <div className='color-picker' data-testid='react-colorful-picker'>
+          <HexColorPicker color={internalColor} onChange={debounce(handleColorfulChange, 100)} />
```
divya-sea, line 90: What's the purpose of debounce here?

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+export const CustomColorPicker: React.FC<CustomColorPickerProps> = ({
+  color,
+  onSubmit,
+  onCancel,
+}: CustomColorPickerProps) => {
+  const { formatMessage } = useIntl();
+  const hexInputFieldRef = useRef<InputProps.Ref>(null);
+  const [saveAttempted, setSaveAttempted] = useState(false);
```
divya-sea, line 44: I still feel we should make this save optional as it makes easier for customer to just select a color from the picker. We need to check with @brenton

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
```diff
@@ -0,0 +1,168 @@
...
+  RgbToHex,
+  hexToRgb,
+  isValidHex,
+  validateHex,
+  validateRgbValue,
+  FixedWidthInput,
+} from './CustomColorPickerHelpers';
+
+export const DEFAULT_COLOR_PICKER_COLOR = {
+  hex: '#ffffff',
```
divya-sea, line 23: This color should come from polaris supported color - https://cloudscape.aws.dev/foundation/visual-foundation/data-vis-colors/. You will have to replace all hard-coded color with polaris supported color.

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
jacksonvogt: Without it, moving the sliders around was jerky and not smooth and it felt uncomfortable to control

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
jacksonvogt: Sounds good, looks like he should be back tomorrow. I'll ping him on slack and see what he thinks, and I'll include you in the chat

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
jacksonvogt: Yes good point, will fix. I used mostly polaris colors but forgot this one🤦🏻‍♂️

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPickerHelpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+};
+
+//Converts a string in decimal form to a string in hex with two digits
+/**
+ * Formats a number in a string to hexidemal in a string
+ *
+ * @param value - A string containing only numeric characters
+ * @returns A string containing the hexidecimal equivalent of the input
+ */
+const decimalToHex = (value: string) => {
```
divya-sea, line 73: A few cases to consider,
1. What happens when we pass negative numbers?
2. What happens when pass large number?

jacksonvogt: As it stands, this function is only called by RgbToHex (which I will rename to rgbToHex), which itself is only called after the rgb input has been validated by validateRgbValue. This ensures that decimalToHex will only be called with a string containing a number between 0-255. However this implementation may change in the future, and your questions could lead to some bad errors. 

Do you think updating the comments above the function to say that the input must be a string of a positive number between 0-255 is enough? Or would it be best if I added some checks to the function?

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const onSubmitAttempt = () => {
+    setSaveAttempted(true);
+    //The hex input is the only field that could conceivably be invalid
+    const valid = isValidHex(hexInputValue);
+    if (valid) {
+      onSubmit(internalColor);
+    } else {
+      hexInputFieldRef.current?.focus();
+    }
```
jkiernan12, line 83: Thoughts on handling errors during user input (in handleHexChange)? Kind of like how Polaris tag editor handles errors immediately during user input. I know for forms we often wait until submission, but for something smaller like this, it might not be bad to show they have a bad hex immediately

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
```diff
@@ -0,0 +1,168 @@
...
+              </div>
+            </FormField>
+            <SpaceBetween size='xs' direction='horizontal'>
+              <FormField label={formatMessage({ defaultMessage: 'R', description: 'Abbreviation for red' })}>
+                <div role='status'>
+                  <FixedWidthInput
+                    width='42px'
+                    value={rgbInputValue.r}
+                    onChange={(e) => handleRGBChange({ r: validateRgbValue(e.detail.value) })}
+                    data-testid='red-input'
```
jkiernan12, line 122: Would using a number input here work? Polaris implementation [here](https://cloudscape.aws.dev/components/input/?tabId=playground&props=N4IgbghgNgrgpiAXCAEnKUD2ACA6pgJygBMQAaECGAF0wGFMBbAByjmoWWoPnJAEsAdsxoBZTMU4hBMRnAL8Axn2oBPZlJmMARvL4BnDsyQgQAXyA). You wouldn't have to worry about validating `onlyDigits` and I think there are some benefits to browsers knowing the type of input expected

jkiernan12 (commented):
Really nice job -- just a couple of questions/comments

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
jacksonvogt: @TheEvilDev The scene-components/tag-style color picker component is already called ColorPicker, but uses this one. Should we rename that one to TagColorPicker or similar?

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
jacksonvogt: Yeah that's a good point. Like you said, polaris team recommends waiting until after a first submission until displaying errors which is why I did it this way, but for a small form like this it might be better to validate every change. I am planning to ask Brenton about the save/cancel flow anyway, I'll see what he thinks about this as well.

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPicker.tsx
jacksonvogt: Good point, will change

packages/scene-composer/tests/components/panels/CustomColorPicker/CustomColorPicker.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+import wrapper from '@awsui/components-react/test-utils/dom';
+
+import {
+  CustomColorPicker,
+  DEFAULT_COLOR_PICKER_COLOR,
+} from '../../../../src/components/panels/CustomColorPicker/CustomColorPicker';
+
+jest.mock('@awsui/components-react', () => ({
+  ...jest.requireActual('@awsui/components-react'),
+}));
```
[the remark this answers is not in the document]
jacksonvogt: Good to know, will fix. I was wondering why this code was used all over the tests since it seems odd (and thought I remembered something from last summer but I wasn't sure...)

packages/scene-composer/src/components/panels/CustomColorPicker/CustomColorPickerHelpers.tsx
jacksonvogt: I didn't know about three's `ColorRepresentation` but it looks really clean 😀. Will switch over to that

PR #1931, opened 2023-09-15
feat(composer): add keyboard nav to floating toolbar

jkiernan12:
## Overview
Lack of keyboard navigation is an a11y issue. This PR adds keyboard nav to the floating toolbar component. Tabbing through the scene composer will now focus toolbar items. The menu can be operated with tab, shift + tab, enter, and escape.

Unit tests have been added for new functionality.

Video demo: https://drive.corp.amazon.com/view/johnfk@/cr-uploads/keyboard_nav_scene_composer.mov

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/components/toolbars/common/ToolbarItem.tsx
```diff
@@ -29,12 +29,56 @@ export function ToolbarItem<T extends ToolbarItemOptions>({
...
+    return null;
+  };
 
   function handleItemClick(item: T) {
     type === 'mode-select' && setSelectedItem(item);
-    onClick?.(item);
+    onSelect?.(item);
   }
 
+  const handleItemKeyDown = (item: T, e, focusIndex?: number) => {
```
jacksonvogt, line 58: Should this be a useCallback since we are passing it down? Could very well not be, I'm no expert on useCallbacks

jacksonvogt (commented):
Looks good! Might be best to get a look or two from the TM team as well as from me, so I'll leave approval to them

packages/scene-composer/src/components/toolbars/common/ItemContainer.tsx
```diff
@@ -90,6 +92,16 @@ export const ItemContainer = React.forwardRef<HTMLDivElement, ToolbarItemContain
...
     );
 
+    const keyDown = useCallback(
+      (e) => {
+        e.stopPropagation();
+        if (!item.isDisabled || e.key !== 'Enter') {
+          onItemKeyDown?.(e);
+        }
+      },
+      [type, item.uuid, onItemKeyDown],
```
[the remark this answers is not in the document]
jkiernan12: Nope -- will update and fix

packages/scene-composer/src/components/toolbars/common/ToolbarItem.tsx
jkiernan12: Yeah, since the component it's being passed to is memoized we can wrap this in useCallback -- same with `handleItemClick` above

Issue #105, opened 2022-05-17
Refactor related-table tests to use create root


diehbria:
**Is your feature request related to a problem? Please describe.**
`ReactDOM.render` is no longer supported in react 18, but is utilized in the related table tests.

This causes the following warning
```
 Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot
```
**Describe the solution you'd like**
Follow guide https://reactjs.org/link/switch-to-createroot

square-li:
fixed by https://github.com/awslabs/iot-app-kit/pull/151

sheilaXu:
This fix is reverted in https://github.com/awslabs/iot-app-kit/pull/150 because it's incompatible with react v17 and we want to continue with v17 for the reason mentioned in the PR.

PR #1941, opened 2023-09-19
fix(composer): create existing matterport tag as entities

sheilaXu:
## Overview
For existing matterport scenes that has already synced tags into scene json file, create them as entities when dynamic scene is enabled and the sync button is clicked 

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
+++ b/packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
@@ -83,9 +83,15 @@ export const MatterportTagSync: React.FC = () => {
     if (tags) {
       for (const [key, value] of tags) {
         if (oldTagMap[key]) {
-          await handleUpdateMatterportTag(layerId, oldTagMap[key].nodeRef, oldTagMap[key].node, value);
+          await handleUpdateMatterportTag({
+            layerId,
+            sceneRootId: rootId,
+            ref: oldTagMap[key].nodeRef,
+            node: oldTagMap[key].node,
+            item: value,
+          });
         } else {
-          await handleAddMatterportTag(layerId, rootId, key, value);
+          await handleAddMatterportTag({ layerId, sceneRootId: rootId, id: key, item: value });
         }
         delete oldTagMap[key];
       }
@@ -102,9 +108,15 @@ export const MatterportTagSync: React.FC = () => {
         }
 
         if (oldTagMap[key]) {
-          await handleUpdateMatterportTag(layerId, oldTagMap[key].nodeRef, oldTagMap[key].node, value);
+          await handleUpdateMatterportTag({
+            layerId,
+            sceneRootId: rootId,
+            ref: oldTagMap[key].nodeRef,
+            node: oldTagMap[key].node,
+            item: value,
+          });
         } else {
-          await handleAddMatterportTag(layerId, rootId, key, value);
+          await handleAddMatterportTag({ layerId, sceneRootId: rootId, id: key, item: value });
         }
         delete oldTagMap[key];
       }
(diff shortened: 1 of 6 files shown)


packages/scene-composer/src/hooks/useMatterportTags.ts
```diff
@@ -164,31 +165,61 @@ const updateTag = async (
...
+      const createEntity: CreateEntityCommandInput = {
+        workspaceId: undefined,
+        entityId: ref,
+        entityName: item.label + '_' + ref,
+        parentEntityId: sceneRootId,
+        components: {
+          Node: nodeComp,
+        },
+      };
+      if (tagComp) {
```
divya-sea, line 212: Maybe you can make them as switch statement for future use?

packages/scene-composer/src/hooks/useMatterportTags.ts
```diff
@@ -164,31 +165,61 @@ const updateTag = async (
...
+        updateEntity.componentUpdates![KnownComponentType.DataOverlay] = overlayComp;
+      }
+
+      try {
+        await sceneMetadataModule.updateSceneEntity(updateEntity);
+      } catch (e) {
+        console.error('Update scene node entity failed', e);
+      }
+    } else {
+      const createEntity: CreateEntityCommandInput = {
```
divya-sea, line 203: Looks like there is no difference in parameters except ParentEntityId. You can pass them as parameter and have them in a single request? or add `parentEntityId` based on the request?

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
```diff
@@ -102,7 +102,7 @@ export const MatterportTagSync: React.FC = () => {
         }
 
         if (oldTagMap[key]) {
-          await handleUpdateMatterportTag(layerId, oldTagMap[key].nodeRef, oldTagMap[key].node, value);
+          await handleUpdateMatterportTag(layerId, rootId, oldTagMap[key].nodeRef, oldTagMap[key].node, value);
```
divya-sea, line 105: Both for line#85 and #105 looks identitical. Why do we have identitical conditions here?

sheilaXu: that's existing code that I didn't change logic other than adding one more parameter

packages/scene-composer/src/hooks/useMatterportTags.ts
sheilaXu: there isn't a thing to switch on

packages/scene-composer/src/hooks/useMatterportTags.ts
sheilaXu: I tried but make code less readable. the API called is different, input type is different, one use `components` and other use `componentUpdates`, and one with `parentEntityId`

packages/scene-composer/src/hooks/useMatterportTags.spec.ts
```diff
@@ -160,13 +160,14 @@ ${mattertagItem.description}`,
...
       ],
+      properties: { [SceneNodeRuntimeProperty.LayerIds]: undefined },
     });
   });
 
   it('should update matterport tag', () => {
     const { handleUpdateMatterportTag } = renderHook(() => useMatterportTags()).result.current;
 
-    handleUpdateMatterportTag('', '', testInternalNode, tagItem);
+    handleUpdateMatterportTag('', '', '', testInternalNode, tagItem);
```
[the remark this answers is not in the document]
sheilaXu: that's true

packages/scene-composer/src/hooks/useMatterportTags.ts
sheilaXu: create entity has `components` instead of `componentUpdates`, and we cannot assign `undefined` value to them.
create has `parentEntityId` but update doesn't.
having the API call here is to minimize the impact of the dynamic scene change. for R2 when we support other kind of nodes, I will refactor this part to be consistent with other nodes

Issue #1890, opened 2023-09-08
Update table widget - Add preferences for pagination


mnischay:
### Acceptance Criteria

- [ ] Add the default settings cog for the table
- [ ] only pagination preference of number of items per chart should be shown
- [ ] only available in edit mode

PR #1944, opened 2023-09-20
feat: preferences for pagination in table widget

Chandru-HM:
## Overview
adding the preferences for pagination to the table widget ( #1890 ) and it includes

1. preferences is available only in edit mode and not visible in preview mode
2. preferences will persist in preview mode which are applied in edit mode
3. preferences have 3 options (10, 20, 30)

## Verifying Changes
refer to the screen capture
image

## Legal

### DIFF
--- a/packages/dashboard/src/customization/widgets/table/component.tsx
+++ b/packages/dashboard/src/customization/widgets/table/component.tsx
@@ -1,7 +1,9 @@
-import React, { useState } from 'react';
-import { useSelector } from 'react-redux';
+import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { CollectionPreferences, CollectionPreferencesProps } from '@cloudscape-design/components';
 
 import { Table, TableColumnDefinition } from '@iot-app-kit/react-components';
+import { SiteWiseAssetQuery } from '@iot-app-kit/source-iotsitewise';
 
 import EmptyTableComponent from './emptyTableComponent';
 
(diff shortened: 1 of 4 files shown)


packages/dashboard/src/types.ts
```diff
@@ -30,6 +30,7 @@ export type DashboardWidget<T extends Record<string, unknown> = Record<string, u
   z: number;
   height: number;
   width: number;
+  pageSize?: number;
```
jmbuss, line 33: We cannot add this here, it affects the public api for all widgets. Instead we should add it to Table Properties - `packages/dashboard/src/customization/widgets/types.ts`

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -145,6 +146,7 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
       x: Math.floor(x),
       y: Math.floor(y),
       z: 0,
+      ...(widgetPresets.type === 'table' && { pageSize: DEFAULT_PREFERENCES.pageSize }),
```
jmbuss, line 149: see comment on the types.ts file, moving the property into the `properties` key of the widget will remove the need for this line entirely.

packages/dashboard/src/components/grid/gestures/usePointerTracker.ts
```diff
@@ -39,9 +39,11 @@ export const usePointerTracker = ({ readOnly, enabled, union, click }: PointerTr
   };
 
   const onPointerUp: PointerEventHandler = (e) => {
+    const target = e.target as HTMLDivElement;
     if (cancelClick || !enabled || readOnly) return;
 
-    if (e.button === MouseClick.Left) {
+    /* handle classname 'awsui_modal' to keep focus on selected widget when clicking on modal */
+    if (e.button === MouseClick.Left && !target.ownerDocument.body.className.includes('awsui_modal')) {
```
jmbuss, line 46: is there any other way to do this. I don't think we should depend on cloudscape classnames since they are not considered stable from their team. Can we use refs instead maybe?

packages/dashboard/src/types.ts
Chandru-HM: Resolved

packages/dashboard/src/components/internalDashboard/index.tsx
Chandru-HM: Resolved

tracy-french (approved):

PR #1951, opened 2023-09-20
fix(composer): create default entity roots

sheilaXu:
## Overview
- create default $SCENES and $LAYERS entity roots for private beta. they will be created from backend for GA
- do not set selected node ref when adding matterport tag from sync button, so overlays will not all be opened during sync

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
+++ b/packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
@@ -12,7 +12,12 @@ import useMatterportObserver from '../../../hooks/useMatterportObserver';
 import { LayerType, MATTERPORT_TAG_LAYER_PREFIX } from '../../../common/entityModelConstants';
 import { createLayer } from '../../../utils/entityModelUtils/sceneLayerUtils';
 import useDynamicScene from '../../../hooks/useDynamicScene';
-import { createSceneRootEntity } from '../../../utils/entityModelUtils/sceneUtils';
+import {
+  checkIfEntityAvailable,
+  createSceneRootEntity,
+  prepareWorkspace,
+} from '../../../utils/entityModelUtils/sceneUtils';
+import { getGlobalSettings } from '../../../common/GlobalSettings';
 
 export const MatterportTagSync: React.FC = () => {
   const logger = useLifecycleLogging('MatterportTagSync');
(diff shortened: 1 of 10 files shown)


packages/scene-composer/src/utils/entityModelUtils/sceneUtils.ts
```diff
@@ -24,3 +30,37 @@ export const createSceneRootEntity = (): Promise<CreateEntityCommandOutput> | un
 
   return getGlobalSettings().twinMakerSceneMetadataModule!.createSceneEntity(input);
 };
+
+export const checkIfEntityAvailable = async (
+  entityId: string,
+  sceneMetadataModule: TwinMakerSceneMetadataModule,
+): Promise<boolean> => {
+  try {
+    await sceneMetadataModule.getSceneEntity({ entityId });
```
divya-sea, line 39: Maybe you can return the response from the api instead of true?

packages/scene-composer/src/utils/entityModelUtils/sceneUtils.ts
```diff
@@ -24,3 +30,37 @@ export const createSceneRootEntity = (): Promise<CreateEntityCommandOutput> | un
...
+    createRootRequests.push(
+      sceneMetadataModule.createSceneEntity({ entityId: SCENE_ROOT_ENTITY_ID, entityName: SCENE_ROOT_ENTITY_NAME }),
+    );
+  }
+  if (!checkRoots[1]) {
+    createRootRequests.push(
+      sceneMetadataModule.createSceneEntity({ entityId: LAYER_ROOT_ENTITY_ID, entityName: LAYER_ROOT_ENTITY_NAME }),
+    );
+  }
+  await Promise.all(createRootRequests);
```
divya-sea, line 65: We a re assuming here that all promises will be resolved successfully. However, if any of the promises in the createRootRequests array reject, it will throw an unhandled promise rejection error. Either you can add try catch here or in the caller method.

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
```diff
@@ -40,10 +45,15 @@ export const MatterportTagSync: React.FC = () => {
     const layerName = MATTERPORT_TAG_LAYER_PREFIX + matterportModelId;
     let layerId = sceneLayerIds?.at(0);
     let rootId = sceneRootId;
+    const sceneMetadataModule = getGlobalSettings().twinMakerSceneMetadataModule;
```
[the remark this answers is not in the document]
sheilaXu: for most cases when `twinMakerSceneMetadataModule` is used to call APIs, it is triggered by a user action (like click the sync button). so we really just need to get the current version of `twinMakerSceneMetadataModule` and call API, no need to keep track of its change.

for `useDynamicScene` hook, it currently only check for the feature flag, but for later releases when we figured out how we want to determine if the dynamic scene feature is enabled for the scene/node, I want to add those logic to it as well. so I created the separate hook instead of `useFeature` directly.

packages/scene-composer/src/utils/entityModelUtils/sceneUtils.ts
sheilaXu: wouldn't API throw error when not succeeded? at least entity not found will throw

packages/scene-composer/src/utils/entityModelUtils/sceneUtils.ts
sheilaXu: the caller should catch. the current component doesn't support showing error state. I'll do it in a separate PR

PR #1954, opened 2023-09-21
fix(composer): show sync matterport tag status

sheilaXu:
## Overview
show sync matterport tag status for loading, error and success statuses

https://github.com/awslabs/iot-app-kit/assets/9315598/3db69d08-be71-48f0-9789-c6baf0d02a50

https://github.com/awslabs/iot-app-kit/assets/9315598/a72dd039-aff1-40c8-860b-938702a8c369

https://github.com/awslabs/iot-app-kit/assets/9315598/3be6b12b-22be-457f-964a-ee28a9c91ca1

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
+++ b/packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useState } from 'react';
 import { useIntl } from 'react-intl';
 import { Button, Popover, StatusIndicator } from '@awsui/components-react';
 import { isEmpty } from 'lodash';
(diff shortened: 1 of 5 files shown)


packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
```diff
@@ -40,40 +40,37 @@ export const MatterportTagSync: React.FC = () => {
 
   const { handleAddMatterportTag, handleUpdateMatterportTag, handleRemoveMatterportTag } = useMatterportTags();
   const { mattertagObserver, tagObserver } = useMatterportObserver();
+  const [syncTagStatus, setSyncTagStatus] = useState<'loading' | 'success' | 'error'>('loading');
```
divya-sea, line 43: You can make them as type and pass the type here.

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
```diff
@@ -135,6 +132,8 @@ export const MatterportTagSync: React.FC = () => {
     for (const key in oldTagMap) {
       handleRemoveMatterportTag(oldTagMap[key].nodeRef);
     }
+
+    setSyncTagStatus('success');
```
divya-sea, line 136: I think you have to check here response coming from prepareWorkspace then set the state to success. It could be still loading too even though the response is successful.

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
sheilaXu: I think it's simple enough to just be defined inline here, and it will do its job to fail the build when a wrong string is set as the status.

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
```diff
@@ -156,8 +155,12 @@ export const MatterportTagSync: React.FC = () => {
         size='small'
         triggerType='custom'
         content={
-          <StatusIndicator type='success'>
-            {intl.formatMessage({ defaultMessage: 'Tags sync successful', description: 'Status indicator label' })}
+          <StatusIndicator type={syncTagStatus} data-testid='sync-button-status'>
```
divya-sea, line 158: aria-label for statusIndicator?

sheilaXu: it shows icon + text, no need to add the label

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
sheilaXu: this is the end of the entire sync function. all tags have been created already at this point. if any API failed before, it shouldn't reach here. `UPDATING` vs `ACTIVE` is not something is function needs to worry about.

I'll update the catch to cover everything above this line.

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
divya-sea: Do we know why assumed only 3 status type here? I still think having its own type better.

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
divya-sea: You can wrap this inside a div and have an aria-label if we want to support accessibility if polaris component does not support directly.

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
divya-sea: What is the assumed state for success? Even though tags are in updating state?

packages/scene-composer/src/components/panels/scene-settings/MatterportTagSync.tsx
divya-sea: And also initial state is set to loading. When this state get's cleared if user hasn't clicked the SyncMatterport button? Why I am asking is you are setting it to loading state again inside doSync method, and outside as well. But Not sure when this get's cleared?

divya-sea (approved):

Issue #1921, opened 2023-09-14
Chart fit and finish - Delete behavior for charts


diehbria:
Behavior:
Single delete will take the chart title not possible now as there is a dependency on config panel title name. For multi-select: rewrite confirmation to “Delete selected widgets.”

When deleting a chart, there are the following changes required:

```[tasklist]
### UX
- [x] Guidance for the UX for the delete friction
```

```[tasklist]
### Tasks
- [ ] Delete key should open of confirmation (see https://cloudscape.design/patterns/resource-management/delete/delete-with-simple-confirmation/)
- [ ] Implement test which utilizes the modal to delete a widget
- [ ] Prevent the users from dragging the chart by dragging the delete "X" icon
- [ ] Fix issue where delete key sometimes does not result in the deletion
```

nadeswal:
<img width="860" alt="Screenshot 2023-09-20 at 11 28 26 AM" src="https://github.com/awslabs/iot-app-kit/assets/141063047/26ebd3c0-6776-461c-89f1-1b9a12f906b0">

Please use the simple confirmation delete modal for this experience.

gmuralidharreddy:
There is a multi-select option and delete all selected widgets at a time, so can we go like the below screenshots?

for single select:
![image](https://github.com/awslabs/iot-app-kit/assets/139960352/8f5226f6-37c7-45b9-80cf-0e01a84a3906)


for multi-select:
![image](https://github.com/awslabs/iot-app-kit/assets/139960352/08287206-678b-40e0-ab39-ff82c4cc77f6)


Note: CSS changes not yet provided.

PR #1960, opened 2023-09-22
feat: added delete widget confirm modal #1921

gmuralidharreddy:
## Overview

Added confirm modal for delete widget when we click on (X) button on widget tile and click on the 'delete' button in the context menu and when we press 'backspace' or 'delete' buttons in the keyboard. Also, fixed widget dragging when we drag in X icon.

### DIFF
--- a/packages/dashboard/src/components/grid/gestures/usePointerTracker.ts
+++ b/packages/dashboard/src/components/grid/gestures/usePointerTracker.ts
@@ -40,7 +40,6 @@ export const usePointerTracker = ({ readOnly, enabled, union, click }: PointerTr
 
   const onPointerUp: PointerEventHandler = (e) => {
     if (cancelClick || !enabled || readOnly) return;
-
     if (e.button === MouseClick.Left) {
       click({
         position: getDashboardPosition(e),
--- a/packages/dashboard/src/components/internalDashboard/index.tsx
+++ b/packages/dashboard/src/components/internalDashboard/index.tsx
@@ -46,6 +46,7 @@ import type { WidgetsProps } from '../widgets/list';
 import type { UserSelectionProps } from '../userSelection';
 import type { DashboardState } from '~/store/state';
 import { useSelectedWidgets } from '~/hooks/useSelectedWidgets';
+import ConfirmDeleteModal from '../confirmDeleteModal';
 
 import '@iot-app-kit/components/styles.css';
 import './index.css';
@@ -86,6 +87,7 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
   const significantDigits = useSelector((state: DashboardState) => state.significantDigits);
 
   const [viewFrame, setViewFrameElement] = useState<HTMLDivElement | undefined>(undefined);
+  const [visible, setVisible] = useState<boolean>(false);
 
   const dispatch = useDispatch();
   const createWidgets = (widgets: DashboardWidget[]) =>
@@ -120,17 +122,22 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
   };
 
   const deleteWidgets = () => {
+    setVisible(true);
+  };
+
+  const onDelete = () => {
     dispatch(
       onDeleteWidgetsAction({
         widgets: selectedWidgets,
       })
     );
+    setVisible(false);
   };
 
   /**
    * setup keyboard shortcuts for actions
    */
-  useKeyboardShortcuts();
+  useKeyboardShortcuts({ deleteWidgets });
 
   /**
    * setup gesture handling for grid
@@ -277,7 +284,31 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
 
   return (
     <TrendCursorSync>
-      <TimeSync group='dashboard-timesync'>{readOnly ? ReadOnlyComponent : EditComponent}</TimeSync>
+      <TimeSync initialViewport={{ duration: '5m' }} group='dashboard-timesync'>
+        {readOnly ? ReadOnlyComponent : EditComponent}
+        <ConfirmDeleteModal
+          visible={visible}
+          headerTitle={`Delete selected widget${selectedWidgets.length > 1 ? 's' : ''}?`}
+          cancelTitle='Cancel'
+          submitTitle='Delete'
+          description={
+            <Box>
+              <Box variant='p'>
+                {`Are you sure you want to delete the selected widget${
+                  selectedWidgets.length > 1 ? 's' : ''
+                }? You'll lose all the progress you made to the
+                          widget${selectedWidgets.length > 1 ? 's' : ''}`}
+              </Box>
+              <Box variant='p' padding={{ top: 'm' }}>
+                You cannot undo this action.
+              </Box>
+            </Box>
+          }
+          handleDismiss={() => setVisible(false)}
+          handleCancel={() => setVisible(false)}
+          handleSubmit={onDelete}
+        />
+      </TimeSync>
     </TrendCursorSync>
   );
 };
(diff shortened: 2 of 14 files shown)


corteggiano:
could you add some screenshots/gifs of the new modal confirmation

packages/dashboard/src/components/confirmModal/confirmModal.css
```diff
@@ -0,0 +1,9 @@
+.btn-custom-primary {
```
corteggiano, line 1: we should not be styling cloudscape components

packages/dashboard/src/components/confirmModal/confirmModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+      data-testid='confirm-modal'
+      footer={
+        <Box float='right'>
+          <SpaceBetween direction='horizontal' size='xs'>
+            {cancelTitle && (
+              <Button variant='link' onClick={handleCancel}>
+                {cancelTitle}
+              </Button>
+            )}
+            <Button variant='primary' className='btn-custom-primary' onClick={handleSubmit} data-testid='submit'>
```
corteggiano, line 42: if using Cloudscape components className is a deprecated prop, please use buttonVariants to distinguish between them

packages/dashboard/src/components/grid/gestures/usePointerTracker.ts
```diff
@@ -39,9 +39,12 @@ export const usePointerTracker = ({ readOnly, enabled, union, click }: PointerTr
   };
 
   const onPointerUp: PointerEventHandler = (e) => {
+    const target = e.target as HTMLDivElement;
+
     if (cancelClick || !enabled || readOnly) return;
 
-    if (e.button === MouseClick.Left) {
+    /* handle classname 'awsui_modal' to keep focus on selected widget when clicking on modal */
+    if (e.button === MouseClick.Left && !target.ownerDocument.body.className.includes('awsui_modal')) {
```
corteggiano, line 47: should use a ref instead, we cannot depend on cloudscape classNames to select elements

jmbuss: you can instead disable the grid when the modal is show so that none of the gestures are dispatched if the grid is not the focus.

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -111,18 +114,25 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
...
     );
+    setVisible(false);
   };
 
   /**
    * setup keyboard shortcuts for actions
    */
   useKeyboardShortcuts();
 
+  useKeyPress('backspace, del', { filter: keyPressFilter, callback: deleteWidgets });
```
jmbuss, line 134: I think this is already a part of keyboard shortcuts.

packages/dashboard/src/components/internalDashboard/keyboardShortcuts.ts
```diff
@@ -45,28 +44,21 @@ export const useKeyboardShortcuts = () => {
     dispatch(onSendWidgetsToBackAction());
   };
 
-  const deleteWidgets = useCallback(() => {
```
jmbuss, line 48: I think instead of moving this out of the keyboard shortcuts, we can either return the modal component to be added to whatever component this hook is being used in, or we can pass in a ref to the hook which we can attach the modal to.

packages/dashboard/src/components/confirmModal/confirmModal.css
gmuralidharreddy: Yeah, but the requirement asking us to give the orange(#ff9900) background color for the button.

packages/dashboard/src/components/confirmModal/confirmModal.tsx
gmuralidharreddy: but the cloudscape button variants do not have the required color(#ff9900 orange). so to match the required background color and border color I used className.

packages/dashboard/src/components/internalDashboard/index.tsx
gmuralidharreddy: resolved I moved back this to keyboard shortcuts

gmuralidharreddy:
> could you add some screenshots/gifs of the new modal confirmation

I tried to attach screenshots while PR creating but github throwing an error like "Something really went wrong, and we can't process that file"  :)

packages/dashboard/src/components/confirmModal/confirmModal.css
corteggiano: talked to Navita and this is the AWS primary button color which is not found in our version of cloudscape
not a blocker for now

packages/dashboard/src/components/grid/gestures/usePointerTracker.ts
jmbuss: There is an action for this - `packages/dashboard/src/store/actions/changeDashboardGrid/changeEnabled.ts`

packages/dashboard/src/components/confirmModal/confirmModal.css
```diff
@@ -0,0 +1,9 @@
...
+  border-color: var(--colors-background-button) !important;
+  background-color: var(--colors-background-button) !important;
+}
+
+.btn-custom-primary:hover {
+  border-color: var(--colors-button-hover) !important;
```
jmbuss, line 7: Can we refactor the orange button into a component so we don't need to keep adding custom css for it? I think we use this in many places right? This way we can limit the areas in the code where we break with Cloudscape convention.

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -279,6 +286,28 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
     <TrendCursorSync>
       <TimeSync initialViewport={{ duration: '5m' }} group='dashboard-timesync'>
         {readOnly ? ReadOnlyComponent : EditComponent}
+        <ConfirmModal
```
jmbuss, line 289: Can we refactor this to its own component? Something like `<ConfirmDeleteModal />`. Trying to keep the complexity of Internal dashboard to a minimum

packages/dashboard/src/components/widgets/tile/tile.tsx
```diff
@@ -75,7 +86,27 @@ const WidgetTile: React.FC<WidgetTileProps> = ({ children, widget, title, remove
             {title}
           </Box>
           <SpaceBetween size='s' direction='horizontal'>
-            {isRemoveable && <DeletableTileAction widget={widget} />}
+            {isRemoveable && <DeletableTileAction handleDelete={handleDelete} />}
+            <ConfirmModal
```
jmbuss, line 90: we can use the same component in InternalDashboard

jmbuss (changes requested):

packages/dashboard/src/components/confirmModal/confirmModal.css
gmuralidharreddy: Created CustomOrangeButton component under components folder

packages/dashboard/src/components/internalDashboard/index.tsx
gmuralidharreddy: renamed to ConfirmDeleteModal

packages/dashboard/src/components/widgets/tile/tile.tsx
gmuralidharreddy: Used ConfirmDeleteModal component

packages/dashboard/src/components/grid/gestures/usePointerTracker.ts
gmuralidharreddy: Disabled grid when the modal is opened

jmbuss (approved):

tracy-french:
@gmuralidharreddy Could you please tackle these merge conflicts and we will merge this in ASAP?

tracy-french:
@gmuralidharreddy I just rebased this for you and will merge when checks pass.

PR #1961, opened 2023-09-22
fix: use one timesync for all of dashboard

corteggiano:
## Overview
we were using two different timeSync components for edit/view mode, so the viewport would reset if you changed dashboard mode

## Legal

### DIFF
--- a/packages/dashboard/src/components/internalDashboard/index.tsx
+++ b/packages/dashboard/src/components/internalDashboard/index.tsx
@@ -1,6 +1,6 @@
 import React, { CSSProperties, ReactNode, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
-import { TimeSync, WebglContext, TrendCursorSync } from '@iot-app-kit/react-components';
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -189,86 +189,80 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
...
+            <WebglContext viewFrame={viewFrame} />
+          </div>
         </div>
       </TrendCursorSync>
-    </TimeSync>
+    );
+  }
+
+  return (
+    <TrendCursorSync>
```
dpportet, line 227: Is there any timesync tests you think we could add to make sure changing to one timesync works correctly?

dpportet (changes requested):
Overall looks good but I noticed [e2e-tests-playwright-full-suite](https://github.com/awslabs/iot-app-kit/actions/runs/6276629276/job/17047527895?pr=1961#logs) is failing with timeout trying to find a `Time Machine` string.

dpportet (approved):

PR #1965, opened 2023-09-25
feat: updated chart initial widget size

Chandru-HM:
## Overview
This PR is for the chart initial widget size updates. new changes include
Line, scatter, bar, timeline charts will have default size: height- 400, width- 650 and for KPI default size is set to: height-200 and width- 350

## Legal

### DIFF
--- a/packages/dashboard/src/customization/widgets/constants.ts
+++ b/packages/dashboard/src/customization/widgets/constants.ts
@@ -1,3 +1,15 @@
 export const SVG_STROKE_SOLID = '1,0';
 export const SVG_STROKE_DASHED = '10,5';
 export const SVG_STROKE_DOTTED = '3,3';
+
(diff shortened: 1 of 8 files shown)


packages/dashboard/src/customization/widgets/constants.ts
```diff
@@ -1,3 +1,15 @@
 export const SVG_STROKE_SOLID = '1,0';
 export const SVG_STROKE_DASHED = '10,5';
 export const SVG_STROKE_DOTTED = '3,3';
+
```
corteggiano, line 4: these should be defined in the 
`iot-app-kit/packages/dashboard/src/customization/widgets` folder
each subfolder has a `plugin.tsx` file which defines an `initialSize`

jmbuss (approved):

PR #2013, opened 2023-09-29
fix: toggle working linechar

ssjagad:
## Overview
Added legend prop to LineScatterChart that reads visible property controlled by toggle. It passes a empty object if true otherwise undefined. This gets read by the BaseChart component who has a conditional render of the legend depending on if the legend prop is defined. 

https://github.com/awslabs/iot-app-kit/assets/145582655/ef070b95-2ddc-4c8a-8139-fccbb77262a2

## Legal

### DIFF
--- a/packages/dashboard/src/customization/widgets/lineScatterChart/component.tsx
+++ b/packages/dashboard/src/customization/widgets/lineScatterChart/component.tsx
@@ -115,6 +115,7 @@ const LineScatterChartWidgetComponent: React.FC<LineScatterChartWidget> = (widge
     thresholds,
     line,
     symbol,
+    legend,
     significantDigits: widgetSignificantDigits,
   } = widget.properties;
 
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/customization/widgets/lineScatterChart/component.tsx
```diff
@@ -172,6 +173,7 @@ const LineScatterChartWidgetComponent: React.FC<LineScatterChartWidget> = (widge
         thresholds={thresholds}
         significantDigits={significantDigits}
         size={size}
+        legend={legend?.visible == true ? {} : undefined}
```
corteggiano, line 176: `legend?.visible ? {} : undefined`

tracy-french, line 176: No need to include == true on this since visible is a boolean.

tracy-french (approved):

PR #2005, opened 2023-09-29
feat: ability to toggle property visibility in config panel

Chandru-HM:
## Overview
added ability to toggle property visibility in config panel and it includes

1. property toggle now toggles the properties on the chart
2. mouse over shows 'hide' or 'unhide' tooltip wrt property visibility status and also shows opposite visibility icons 

## Legal
This project is available under the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0.html).

Issue #1986, opened 2023-09-27
Dashboard: Add ability to toggle property visibility in config panel


diehbria:
## User journey:

Add the "visibility toggle" within the configuration panel of the dashboard.
![Screenshot 2023-09-26 at 3 48 46 PM (1)](https://github.com/awslabs/iot-app-kit/assets/77755322/9b34d381-a15d-4c1f-ba3e-312fb81e2b3c)

When the property is visible, the "eye icon" should be present while no cursor hover is active.
Upon hovering over the eye button, the icon should display the "crossed out eye" icon to indicate that click will hide the property, and a tooltip should appear that says "hide"

When the property is hidden, the "crossed out eye icon" should be present if the user is not hovering their cursor over the icon. When the cursor is hovering over it, the button should display the "eye icon" to indicate that clicking will un-hide the property, and a tooltip which says "unhide"

## Technical implementation:

react-components <Chart /> does not have a visible flag on the queries property, but the dashboard definition does have one on the query: StyledAssetQuery. This means in the LineScatterChart component in dashboard, in order to toggle visibility of an asset property, we need to filter out any asset properties where that visible flag is false before passing it to the timeSeriesData function.

diehbria:
working on addressing a missing behavior of hiding the data visualization on widgets

PR #2016, opened 2023-09-29
fix(ResourceExplorer): implement toggling on/off of properties visibi…

diehbria:
This PR provides the basic functionality to toggle visibility of data streams on widgets, resolving issue https://github.com/awslabs/iot-app-kit/issues/1986.

Removed alterations to the public API made in https://github.com/awslabs/iot-app-kit/pull/2005/files, as `visible` 

Attached is a screen shot demonstrating that a hidden property is reflected in the widget by no longer being visualized:
Screenshot 2023-09-29 at 3 00 40 PM

Note: this feature will only work with line and scatter chart.

---

Additionally this PR fixes the build process on dashboard that exists currently within `main` branch, where it fails to build the typescript types.

## Legal

### DIFF
--- a/packages/dashboard/src/customization/propertiesSections/propertiesAndAlarmsSettings/sectionTypes.ts
+++ b/packages/dashboard/src/customization/propertiesSections/propertiesAndAlarmsSettings/sectionTypes.ts
@@ -7,7 +7,7 @@ export type OnDeleteAssetQuery = (params: {
   assetId: string;
   propertyId: string;
   siteWiseAssetQuery: IoTSiteWiseDataStreamQuery | StyledAssetQuery;
-  updateSiteWiseAssetQuery: (newQuery: IoTSiteWiseDataStreamQuery) => void;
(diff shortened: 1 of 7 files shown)


tracy-french (approved):

packages/dashboard/src/components/widgets/list.tsx
```diff
@@ -20,6 +21,38 @@ export type WidgetsProps = {
   messageOverrides: DashboardMessages;
 };
 
+export const isQueryWidget = (widget: DashboardWidget): widget is LineScatterChartWidget =>
+  widget.properties.queryConfig != null;
+
+const getWidgetsWithHiddenPropertiesRemoved = (dashboardConfiguration: {
```
jmbuss, line 27: can we move this into `customization/widgets/lineScatterChart/component` since it will only work with that widget?

tracy-french: +1

diehbria: done

PR #2018, opened 2023-09-29
feat(composer): utils to handle camera component in entity

sheilaXu:
## Overview
Add util functions to handle create/update camera component request, and parse camera from entity model back to scene component model.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/utils/entityModelUtils/cameraComponent.ts
+++ b/packages/scene-composer/src/utils/entityModelUtils/cameraComponent.ts
@@ -1,0 +1,95 @@
+import { ComponentRequest, ComponentUpdateRequest } from '@aws-sdk/client-iottwinmaker';
+import { DocumentType } from '@aws-sdk/types';
+
+import { ICameraComponent, KnownComponentType } from '../../interfaces';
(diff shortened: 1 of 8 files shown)


packages/scene-composer/src/utils/entityModelUtils/cameraComponent.ts
```diff
@@ -0,0 +1,95 @@
...
+    properties: {
+      [CameraComponentProperty.CameraType]: {
+        value: {
+          stringValue: camera.cameraType,
+        },
+      },
+    },
+  };
+
+  if (camera.fov) {
```
mumanity, line 31: Nit-pick: I think this would be a good scenario for using a `switch`, but that's just a personal preference.

mumanity (approved):

packages/scene-composer/src/utils/entityModelUtils/cameraComponent.ts
sheilaXu: switch will not work. we want to assign all values that is available instead of handling different cases of a single value

PR #2026, opened 2023-10-02
feat(dynamic scene): crud functionalities for model ref

divya-sea:
## Overview
* crud functionalities for model ref component. 
Followed the design doc to implement the model ref functionalities. 
* Create a model ref.
* Update a model ref.
* Parse the component. 
## Verifying Changes
npm run test:unit

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.spec.ts
+++ b/packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.spec.ts
@@ -1,0 +1,148 @@
+import { componentTypeToId } from '../../common/entityModelConstants';
+import { KnownComponentType } from '../../interfaces';
+
+import { createModelRefComponent, parseModelRefComp, updateModelRefComponent } from './modelRefComponent';
+
+describe('createModelRefComponent', () => {
+  it('should return expected modelRef component with uri and modelType', () => {
+    const result = createModelRefComponent({ type: KnownComponentType.ModelRef, uri: 'abc', modelType: 'modelType' });
+
+    expect(result.properties).toEqual({
+      uri: {
+        value: { stringValue: 'abc' },
+      },
+      modelType: {
+        value: { stringValue: 'modelType' },
+      },
+    });
+  });
+
+  it('should return expected modelRef component by setting castShadow and receiveShadow to false', () => {
+    const result = createModelRefComponent({
+      type: KnownComponentType.ModelRef,
+      uri: 'abc',
+      modelType: 'modelType',
+      unitOfMeasure: 'decimeters',
+      localScale: [1, 1, 1],
+      castShadow: false,
+      receiveShadow: false,
+    });
+
+    expect(result.properties).toEqual({
+      uri: {
+        value: { stringValue: 'abc' },
+      },
+      modelType: {
+        value: { stringValue: 'modelType' },
+      },
+      unitOfMeasure: {
+        value: {
+          stringValue: 'decimeters',
+        },
+      },
+      localScale: {
+        value: {
+          listValue: [{ doubleValue: 1 }, { doubleValue: 1 }, { doubleValue: 1 }],
+        },
+      },
+      castShadow: {
+        value: {
+          booleanValue: false,
+        },
+      },
+      receiveShadow: {
+        value: {
+          booleanValue: false,
+        },
+      },
+    });
+  });
+
+  it('should return expected modelRef component by setting castShadow and receiveShadow to true ', () => {
+    const result = createModelRefComponent({
+      type: KnownComponentType.ModelRef,
+      uri: 'abc',
+      modelType: 'modelType',
+      unitOfMeasure: 'decimeters',
+      localScale: [1, 1, 1],
+      castShadow: true,
+      receiveShadow: true,
+    });
+
+    expect(result.properties).toEqual({
+      uri: {
+        value: { stringValue: 'abc' },
+      },
+      modelType: {
+        value: { stringValue: 'modelType' },
+      },
+      unitOfMeasure: {
+        value: {
+          stringValue: 'decimeters',
+        },
+      },
+      localScale: {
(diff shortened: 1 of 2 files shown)


packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.ts
```diff
@@ -0,0 +1,96 @@
...
+      },
+    };
+  }
+  if (model.modelType) {
+    comp.properties![ModelRefComponentProperty.ModelType] = {
+      value: {
+        stringValue: model.modelType,
+      },
+    };
+  }
```
sheilaXu, line 36: `uri` and `modelType` are required field so no need to check if they exist. just set their value at line 21

packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.ts
```diff
@@ -0,0 +1,96 @@
...
+  if (model.receiveShadow) {
+    comp.properties![ModelRefComponentProperty.ReceiveShadow] = {
+      value: {
+        booleanValue: model.receiveShadow,
+      },
+    };
+  }
```
sheilaXu, line 64: you should check explicitly for if they are` undefined` or not. in your current code, you cannot save the boolean value to be `false`

packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.ts
```diff
@@ -0,0 +1,96 @@
...
+export const updateModelRefComponent = (model: IModelRefComponent): ComponentUpdateRequest => {
+  const request = createModelRefComponent(model);
+  return {
+    componentTypeId: request.componentTypeId,
+    propertyUpdates: request.properties,
+  };
+};
+
+export const parseModelRefComp = (model: DocumentType): IModelRefComponentInternal | undefined => {
+  if (!model?.['properties']) {
```
sheilaXu, line 77: you should also check if `uri` and `modelType` exist since we cannot render without them. if for some reason they are not returned from the query, we should return undefined here as we cannot render it

packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      type: KnownComponentType.ModelRef,
+      uri: 'abc',
+      modelType: 'modelType',
+      unitOfMeasure: 'decimeters',
+      localScale: [1, 1, 1],
+      castShadow: false,
+      receiveShadow: false,
+    });
+
+    expect(result.properties).toEqual({
```
sheilaXu, line 31: result should include `castShadow` and `receiveShadow` property with value false

sheilaXu (changes requested):

packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.ts
divya-sea: Yes If we want to save `false`. Otherwise not needed.

packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.spec.ts
divya-sea: If save yes we should include add false result here.

packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.ts
sheilaXu: false is one of the valid value for boolean, we need to support saving it

divya-sea: If the value of undefined then it is undefined not false. are you saying we need to explicitly set it to false if it is undefined?

divya-sea: Maybe we can check if it is not equal to undefined then set the propertyVlaue.

sheilaXu:
```
if (model.receiveShadow !== undefined) {
  ...
}
```

divya-sea: Yeah that's what I did.

packages/scene-composer/src/utils/entityModelUtils/modelRefComponent.ts
```diff
@@ -0,0 +1,97 @@
...
+  const request = createModelRefComponent(model);
+  return {
+    componentTypeId: request.componentTypeId,
+    propertyUpdates: request.properties,
+  };
+};
+
+export const parseModelRefComp = (model: DocumentType): IModelRefComponentInternal | undefined => {
+  const uriExists = !model?.['properties'].some((prop) => prop['propertyName'] === 'uri');
+  const modelTypeExists = !model?.['properties'].some((prop) => prop['propertyName'] === 'modelType');
```
sheilaXu, line 76: the variable names are misleading. they really mean `xxxMissing` instead of `xxxExists`. they cause problem for people to understand why `uriExists || modelTypeExists` needs to return undefined

sheilaXu: also you can use `find` instead of `some` here. so when their value exist, you can just use those values to construct the component instead of `find` again below

divya-sea: Actually if we want to reuse the value then we don't have to use missing or exists. We can just findUri/FindModelType. If undefined then return undefined or elase using `findUri.propertyValue`

sheilaXu (approved):

PR #2032, opened 2023-10-03
fix(scene): remove transform controls from raycast

haweston:
## Overview
When a camera is selected in the Scene Composer it has Transform Controls added to for editing it.  The controls wrap the camera and cause raycasts from it to return early by detecting the controls instead of the intended scene object.  So the raycasts used to control the camera placement on activating a Camera View will lock on to a target at distance zero in an unpredictable direction from the camera.  The camera is then point in the wrong direction and camera controls will be unresponsive as the target is at distance zero.

Fix by not having Transform Controls respond to the raycast.

Before
https://github.com/awslabs/iot-app-kit/assets/109186219/00d0a2f3-cfac-4933-a6ba-851b9e2bd8d3

After
https://github.com/awslabs/iot-app-kit/assets/109186219/ec1bff31-5115-4af6-818b-2e4ffd145528

## Verifying Changes
Use scene composer on a scene with a camera
Select the camera in the scene hierarchy
Active the same camera in the Camera View list
verify that the camera results looks like the preview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


divya-sea:
Ui tests are failing.

PR #2035, opened 2023-10-04
fix: add eslint rule for hooks

corteggiano:
## Overview
Add eslint rule for rules of hooks, so our useEffects are more efficient

lint rules were added to `packages/dashboard` and `packages/react-components`

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
+++ b/packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
@@ -204,12 +204,15 @@ export const useSeriesAndYAxis = (
   const defaultYAxis: YAXisComponentOption[] = useMemo(() => [convertChartYAxis(axis)], [axis]);
   const convertedThresholds = convertThresholds(thresholds);
   const getStyles = getChartStyleSettingsFromMap(styleSettings);
+  const datastreamDeps = JSON.stringify(datastreams.map(({ id, refId }) => `${id}-${refId}`));
 
   const { series, yAxis, yMaxs, yMins } = useMemo(() => {
     return datastreams
       .map((datastream) => convertSeriesAndYAxis(getStyles(datastream))(datastream))
       .reduce(reduceSeriesAndYAxis, { series: [], yAxis: defaultYAxis, yMins: [], yMaxs: [] });
-  }, [datastreams, styleSettings, defaultYAxis]);
+    // disabling because datastream are stringified as datastreamDeps
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [datastreamDeps, defaultYAxis, getStyles]);
 
   if (series.length > 0) {
     series[0].markArea = convertedThresholds.markArea;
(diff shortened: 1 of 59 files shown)


packages/react-components/src/components/chart/converters/convertSeriesAndYAxis.ts
```diff
@@ -204,12 +204,15 @@ export const useSeriesAndYAxis = (
...
   const getStyles = getChartStyleSettingsFromMap(styleSettings);
+  const datastreamDeps = JSON.stringify(datastreams.map(({ id, refId }) => `${id}-${refId}`));
 
   const { series, yAxis, yMaxs, yMins } = useMemo(() => {
     return datastreams
       .map((datastream) => convertSeriesAndYAxis(getStyles(datastream))(datastream))
       .reduce(reduceSeriesAndYAxis, { series: [], yAxis: defaultYAxis, yMins: [], yMaxs: [] });
-  }, [datastreams, styleSettings, defaultYAxis]);
+    // disabling because datastream are stringified as datastreamDeps
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
mnischay, line 214: I dont know what this exactly means, but looks like its worth adding an exception / turning off this rule in the plugin instead of having this eslint-disable comment all over the place?

corteggiano: in this case the eslint rule doesnt realize that `datastreamDeps` represents the `dataStreams` object
we dont want to turn this rule off everywhere though because it does apply in other situations

packages/react-components/src/components/chart/hooks/useTrendCursorsEvents.ts
```diff
@@ -55,50 +55,58 @@ const useTrendCursorsEvents = ({
...
-    } else {
-      graphicRef.current[toBeDeletedGraphicIndex].$action = 'remove';
-      graphicRef.current[toBeDeletedGraphicIndex].children = []; // Echarts will throw error if children are not empty
-      chartRef.current?.setOption({ graphic: graphicRef.current });
-      graphicRef.current.splice(toBeDeletedGraphicIndex, 1);
-      setGraphicRef.current([...graphicRef.current]);
-    }
-  };
+    },
+    [groupId, graphicRef, chartRef, deleteTrendCursorsInSyncState]
```
mnischay, line 108: the point of having refs like graphicRef was to not have this re run when the graphic changes, the ref will hold the latest value and can be referred directly

packages/react-components/src/components/chart/hooks/useTrendCursorsEvents.ts
```diff
@@ -55,50 +55,58 @@ const useTrendCursorsEvents = ({
...
+            visualization,
+          });
+
+          if (newTc) {
+            setGraphicRef.current([...graphicRef.current, newTc]);
+          }
+        }
+      }
+    },
+    [groupId, chartRef, sizeRef, seriesRef, visualization, addTrendCursorsToSyncState]
```
mnischay, line 89: the point of having refs like seriesRef was to not have this re run when the series changes, the ref will hold the latest value and can be referred directly

tracy-french (approved):
Beautiful change! Thank you, @corteggiano!

PR #2042, opened 2023-10-05
feat: fake sitewise

tracy-french:
## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/dashboard/src/msw/iot-sitewise/resources/assets/index.ts
```diff
@@ -0,0 +1,123 @@
+import {
```
dpportet, line 1: Is there a specific type of "factory" or asset relationship we're setting up for the test data? Do we have any information on the asset relationships for any real use cases? It would be cool to have multiple use cases based on how different real customers setup SiteWise, and then we could run tests which validate all those uses cases.

tracy-french: Nothing specific at the moment, though I completely agree! We could create multiple hierarchies for different use cases and use different handlers to swap between them.

PR #2044, opened 2023-10-05
fix(dashboard): chart respects absolute min and max between data and thresholds

hwandersman:
## Overview
When a threshold is defined higher than the max or lower than the min of the yAxis on the line and scatter chart it did not show up.

Updated the following:
* When there are no properties or property data customer can still add/edit/delete thresholds. They will just not be shown on the chart.
* When a threshold is set >max or <min and the y-axis is on "Auto" then the axis will be updated to have at least that max/min with some padding
* Thresholds will not work when multiple y-axis are on the chart. Will need to wait to get feedback about that feature in general and this will not be in scope.
* The default threshold style setting is line + filled region

## Verifying Changes
Verified with unit tests and manual tests.

Some edge cases:
1. yAxis is set manually - manual min and/or max is always prioritized and observed
2. yAxis is auto, there is data and no thresholds, yAxis is automatically set by echarts
3. yAxis is auto, there is no data and thresholds, nothing show up on the chart
4. yAxis is auto, there is data and thresholds, absolute min > 0, absolute max > data, yAxis max is updated and padded to the nearest whole num / multiple power of 10
5. yAxis is auto, there is data and thresholds, absolute min < data, absolute max < 0, yAxis min is updated and padded to the nearest whole num / multiple power of 10
6. yAxis is auto, there is data and thresholds, absolute min < 0, absolute max > data, yAxis min and max are updated and padded to the nearest whole num / multiple power of 10
7. Absolute min / max is a decimal, padded correctly
8. Absolute min / max is a negative num, padded correctly
9. Absolute min / max is a very large num, padded correctly

https://github.com/awslabs/iot-app-kit/assets/87385528/a658a005-6150-4cae-a112-555ce9fd3276

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/converters/padYAxis.ts
+++ b/packages/react-components/src/components/chart/converters/padYAxis.ts
@@ -1,0 +1,150 @@
+import { DataPoint, DataStream, Threshold } from '@iot-app-kit/core';
+
+import maxBy from 'lodash.maxby';
+import minBy from 'lodash.minby';
+
+import { ScaleDataValue } from 'echarts/types/src/util/types';
+import { dataValue } from './convertSeriesAndYAxis';
+import { YAxisOptions } from '../types';
+
+const PADDING_FACTOR = 0.1;
+
+type AxisValue =
+  | ScaleDataValue
+  | 'dataMin'
+  | 'dataMax'
+  | ((extent: { min: number; max: number }) => ScaleDataValue)
+  | undefined;
+
+// Verify that the input value is a valid number
+export const validateValue = (value: string | undefined) => {
+  return value !== undefined ? +value : undefined;
+};
+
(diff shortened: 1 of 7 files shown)


packages/react-components/src/components/chart/converters/padYAxis.ts
```diff
@@ -0,0 +1,150 @@
+import { DataPoint, DataStream, Threshold } from '@iot-app-kit/core';
```
corteggiano, line 1: note for other reviewers, this is pulled directly from the synchro-charts codebase :) reduce reuse recycle

corteggiano (approved):

PR #2048, opened 2023-10-05
fix(dashboard): minor threshold fixes

hwandersman:
## Overview
Issues:
* Certain widgets within the dashboard are difficult to configure thresholds for due to the dropdown being squeezed.
* "=" thresholds should display as a line

Fixes:
* The dropdown for the comparison operator now always expands fully.
* For all thresholds with "=" in the expression there will be a dotted line. For all other expressions it uses the style to determine whether there is a region and/or dotted line.
* The timeline widget now supports the "Contains" comparison operator
* Selecting multiple widgets of the same threshold type then you can update thresholds for all of them at once. If selecting multiple widgets of different threshold types then you see text saying that you must select one threshold.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/87385528/bfaab770-b494-4ba2-9b54-03d32c824981

https://github.com/awslabs/iot-app-kit/assets/87385528/03f2a2e2-66da-4e75-8b2d-6b5274ca3d33

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdsSection.tsx
+++ b/packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdsSection.tsx
@@ -89,6 +89,27 @@ const ThresholdsSection: FC<ThresholdsSectionProps> = ({
 
   const [isExpanded, setIsExpanded] = useExpandable();
 
+  /**
+   * Wrapper for updating styled thresholds
+   * If a threshold's comparison operator is "=" then always show a visible line
+   * If the update to thresholds is a new style then pass it into this function, otherwise use the style state
+   */
+  const handleUpdateStyledThresholds = (
+    updatedThresholds: (ThresholdWithId & StyledThreshold)[] | undefined,
+    newThresholdStyle?: ThresholdStyleType
+  ) => {
+    if (updateStyledThresholds) {
+      const thresholds = updatedThresholds?.map((t) => {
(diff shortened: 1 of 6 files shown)


diehbria:
looks great from a UX perspective - just need to enable contains operator for thresholds for the status timeline

packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdsSection.tsx
```diff
@@ -89,6 +89,27 @@ const ThresholdsSection: FC<ThresholdsSectionProps> = ({
 
   const [isExpanded, setIsExpanded] = useExpandable();
 
+  /**
+   * Wrapper for updating styled thresholds
+   * If a threshold's comparison operator is "=" then always show a visible line
+   * If the update to thresholds is a new style then pass it into this function, otherwise use the style state
+   */
+  const handleUpdateStyledThresholds = useCallback(
```
tracy-french, line 97: Why do we use `useCallback` for this function? What would happen if we didn't use `useCallback`?

hwandersman: I thought it would help resolve the saved global thresholdStyle, but it wasn't working. Removing it functions the same, so I removed it.

PR #2052, opened 2023-10-05
fix: default resolution and aggregation

tracy-french:
## Overview
This change provides required defaults for aggregation and resolution for widgets which expose aggregation and resolution controls.

## Legal


packages/dashboard/src/customization/widgets/barChart/plugin.tsx
```diff
@@ -14,6 +14,8 @@ export const barChartPlugin: DashboardPlugin = {
         icon: BarIcon,
       },
       properties: () => ({
+        aggregationType: 'AVERAGE',
```
jmbuss, line 17: does bar chart have these at a root level? I thought it was still using the old query config?

jmbuss (approved):

PR #2061, opened 2023-10-09
fix(dashboard): bar chart default aggregation+resolution

corteggiano:
## Overview
Fix assignment of default aggregation and resolution so bar chart renders properly 

bar

## Legal


packages/dashboard/src/customization/widgets/utils/assignDefaultStyleSettings.ts
```diff
@@ -4,17 +4,23 @@ import type { QueryWidget } from '../types';
...
 ) => ({
   assets: assets.map(({ properties, ...others }) => ({
     ...others,
     properties: properties.map((propertyQuery) => ({
       ...propertyQuery,
+      resolution: resAndAggr.resolution,
+      aggregationType: resAndAggr.aggregation,
       refId: propertyQuery.refId || getId(),
     })),
   })),
```
tracy-french, line 26: I believe you need to do the same for the properties field below this to enable unmodeled data streams.

tracy-french (approved):

Issue #1993, opened 2023-09-27
Resource explorer: have the latest value field obey dashboard level decimal place setting


diehbria:
This task is to make the resource explorer take into account the dashboard wide decimal place settings.

Decimal place settings can be configured by clicking the top right gear button, and updating the "Decimal places" input as shown below:
<img width="587" alt="pic1" src="https://github.com/awslabs/iot-application/assets/77755322/99a5b553-c325-49e5-be7e-ffb7b6bd4c0a">

the resource explorer on various tables displays latest value, which currently displays the full decimal without taking into account dashboard level decimal settings (shown below):
<img width="450" alt="pic2" src="https://github.com/awslabs/iot-application/assets/77755322/43266ed5-de80-4447-8984-f68849cd72d1">

The expect behavior is for the resource explorer "latest value" cells to round to the number of decimal places as indicated in the dashboard level settings.

PR #2091, opened 2023-10-12
feat: fix decimal places based on significant value

gmuralidharreddy:
## Overview
This PR is for #1993 

- Fixed decimal places based on the significant value from settings decimal places

## Verifying Changes
https://imgur.com/a/SAdfW3s

## Legal


jmbuss (commented):
We actually have a util for this in react-components - `packages/react-components/src/utils/number.ts`

Let's move this into core-utils so that we can use the same rounding logic everywhere

gmuralidharreddy:
> We actually have a util for this in react-components - `packages/react-components/src/utils/number.ts`
> 
> Let's move this into core-utils so that we can use the same rounding logic everywhere

I have made the changes as requested, not sure why all checks are failing here but those are working locally.

gmuralidharreddy:
Closing this PR and raised the below new PR

https://github.com/awslabs/iot-app-kit/pull/2108

PR #2092, opened 2023-10-12
fix: line charts colors too quickly choose similar colors

ssjagad:
## Overview
Currently when trends are added to a widget, the colors cycle towards similar colors quicker than desired. Changed the getDefaultColor function to offset color selection by 1 so colors don't repeat. 

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/145582655/7e364c13-8ebc-4466-bbdc-c2829ce77271

## Legal


packages/core-util/src/sdks/colorPalette.ts
```diff
@@ -37,6 +36,7 @@ export const colorPalette = [
   '#5f6b7a',
   '#3759ce',
   '#b1325c',
+  '#5978e3',
```
dpportet, line 39: What's the purpose of moving this color?

dpportet (commented):
In the verification video I noticed the green unmodelled data stream didn't show up on the chart, is that intended?

packages/core-util/src/sdks/colorPalette.ts
ssjagad: That blue is very similar to the blue at index 2. Moved it further down and made sure that 4 colors above and below are not similar.

ssjagad:
> In the verification video I noticed the green unmodelled data stream didn't show up on the chart, is that intended?

Seems to be a bug in general. Just testing in RC and none of the modeled data streams are showing up on the line chart.

dpportet (approved):

PR #2096, opened 2023-10-13
fix(echarts): improved x+y axis min and max

corteggiano:
## Overview
- move any viewport+echart logic to live in the useDataZoom hook, so we dont have conflicting logic in different files
- X Axis - fix the max value so that selecting an absolute time range works
- Y Axis - create a default y axis to be displayed when series have no data present

betterAxis

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/utils/getXAxis.ts
+++ b/packages/react-components/src/components/chart/utils/getXAxis.ts
@@ -1,8 +1,8 @@
-import { ChartAxisOptions, ViewportInMs } from '../types';
+import { ChartAxisOptions } from '../types';
 import { DEFAULT_X_AXIS, DEFAULT_X_AXIS_ID } from '../eChartsConstants';
 import { XAXisOption } from 'echarts/types/dist/shared';
 
-export const getXAxis = (viewportInMs: ViewportInMs, axis?: ChartAxisOptions): XAXisOption => {
+export const getXAxis = (axis?: ChartAxisOptions): XAXisOption => {
   return {
     id: DEFAULT_X_AXIS_ID,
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/chart/utils/getXAxis.ts
```diff
@@ -19,6 +19,6 @@ export const getXAxis = (viewportInMs: ViewportInMs, axis?: ChartAxisOptions): X
     },
     splitNumber: 6,
     min: 0,
-    max: viewportInMs.isDurationViewport ? viewportInMs.end : undefined,
+    max: viewportInMs.isDurationViewport ? viewportInMs.end : Date.now(),
```
mnischay, line 22: Date.now will not change the x-axis right ? 

this is basically 

`max: Date.now()`

packages/react-components/src/components/chart/utils/getXAxis.ts
```diff
@@ -19,6 +19,6 @@ export const getXAxis = (viewportInMs: ViewportInMs, axis?: ChartAxisOptions): X
     },
     splitNumber: 6,
     min: 0,
-    max: viewportInMs.isDurationViewport ? viewportInMs.end : undefined,
```
corteggiano, line 22: undefined was leaving the x axis max to be auto calculated
we were running into the situation where using the time machine to select a new range made it impossible to scroll to the right :( 

setting the max to be date.now allows a user to select a date in the past and still scroll left/right as expected

mnischay (dismissed):
LGTM

mnischay (approved):

PR #2102, opened 2023-10-16
fix(react-components): echarts grab on canvas update cursor and tooltip

hwandersman:
## Overview
* Change cursor to open hand when holding shift while over chart area
* When clicking to pan, change to "grab" hand
* Tooltip is hidden while there is an active gesture

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/87385528/f9158635-23f5-45a4-ac22-63a27e261c72

1. Always show the "grab" cursor when shift is held down over the chart canvas
2. Always show the "grabbing" cursor when shift and the mouse is held down over the chart canvas, turn off the tooltip, and drag the canvas left to right
3. Don't show any change in cursor when the mouse is held down
4. Cannot drag on the edges of the canvas, but there is no way to avoid changing the cursor since the event is within the entire canvas

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/baseChart.tsx
+++ b/packages/react-components/src/components/chart/baseChart.tsx
@@ -27,6 +27,7 @@ import { DEFAULT_CHART_VISUALIZATION } from './eChartsConstants';
 import { useDataZoom } from './hooks/useDataZoom';
 import { useViewport } from '../../hooks/useViewport';
 import { getXAxis } from './utils/getXAxis';
+import { useHandleChartEvents } from './hooks/useHandleChartEvents';
 
 /**
  * Developer Notes:
@@ -117,14 +118,24 @@ const BaseChart = ({ viewport, queries, size = { width: 500, height: 500 }, ...o
   // determine the set option settings
   const settings = useChartSetOptionSettings(dataStreams);
 
-  //handle dataZoom updates, which are dependent on user events and viewportInMS changes
+  // handle dataZoom updates, which are dependent on user events and viewportInMS changes
   useDataZoom(chartRef, viewportInMs);
+
+  // handle chart event updates
+  const eventOptions = useHandleChartEvents(chartRef);
+
+  const toolTipOptions = {
+    ...convertedOptions.tooltip,
+    ...eventOptions.tooltip,
+  };
 
   // set all the options on the echarts instance
   useEChartOptions(
     chartRef,
     {
       ...convertedOptions,
+      ...eventOptions,
+      tooltip: toolTipOptions,
       series,
       yAxis,
       xAxis,
--- a/packages/react-components/src/components/chart/hooks/useHandleChartEvents.ts
+++ b/packages/react-components/src/components/chart/hooks/useHandleChartEvents.ts
@@ -1,0 +1,113 @@
+import { MutableRefObject, useEffect, useState } from 'react';
+import { EChartsOption, EChartsType } from 'echarts';
+
+export const useHandleChartEvents = (chartRef: MutableRefObject<EChartsType | null>) => {
+  const [optionState, setOptionState] = useState<EChartsOption>({});
+
+  // When dragging on the chart, turn off tooltip and set cursor to grabbing
+  const setOnDragStart = (chart: EChartsType | null) => {
+    setOptionState({
+      tooltip: {
+        show: false,
+      },
+    });
+    chart?.getZr().setCursorStyle('grabbing');
+  };
+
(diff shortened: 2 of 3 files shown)


packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -62,4 +62,111 @@ export const useDataZoom = (chartRef: MutableRefObject<EChartsType | null>, view
     // ignoring because refs dont need to be in dependency array
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [setViewport]);
+
+  // When dragging on the chart, turn off tooltip and set cursor to grabbing
+  const setOnDragStart = (chart: EChartsType | null) => {
```
mnischay, line 67: All these events have nothing to do with DataZoom, handle all the user events in a different file. 
We also have handlers listening to these events in the trendcursors file,  if they are dependent, I would move these there, if not, i would suggest a different file something like handleChartEvents

mnischay: Also, all the setoptions happen in one place, so I would return the value using state, i.e the end of the handler would return `{ tooltip: {
        show: false,
      },} ` which then will be used in the last hook which sets the options.

mnischay (changes requested):

packages/react-components/src/components/chart/hooks/useDataZoom.ts
hwandersman: Updated! Didn't know `setOptions` is only set in one place, good callout. Moved everything to its own file.

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -117,14 +118,18 @@ const BaseChart = ({ viewport, queries, size = { width: 500, height: 500 }, ...o
...
 
+  // handle chart event updates
+  const eventOptions = useHandleChartEvents(chartRef);
+
   // set all the options on the echarts instance
   useEChartOptions(
     chartRef,
     {
       ...convertedOptions,
+      ...eventOptions,
```
mnischay, line 132: convertedOptions also has options for tooltip, you probably want to seperate out tooltip options or merge them

packages/react-components/src/components/chart/hooks/useHandleChartEvents.ts
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  // Set up event listeners for pressing shift and dragging the mouse to update
+  // the cursor and tooltip
+  useEffect(() => {
+    const chart = chartRef?.current;
+
+    let shiftDown = false;
+    let mouseDown = false;
+    let prevIsGrabbing = false;
```
mnischay, line 35: for prev value, you probably want to use a useref ?

packages/react-components/src/components/chart/hooks/useHandleChartEvents.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+    chart?.getZr().setCursorStyle('grab');
+  };
+
+  // Set up event listeners for pressing shift and dragging the mouse to update
+  // the cursor and tooltip
+  useEffect(() => {
+    const chart = chartRef?.current;
+
+    let shiftDown = false;
```
mnischay, line 33: curious: is this better than using local state ?

mnischay (changes requested):
The merged of options is the only major call out, rest are minor. Also, please test it along with trendcursors i.e add a couple and see how these event react along with it.

packages/react-components/src/components/chart/baseChart.tsx
hwandersman: The convertedOptions tooltip does not specify the `show` parameter. Even if it did we'd want the eventOptions to override it, so this should be safe to do.

packages/react-components/src/components/chart/hooks/useHandleChartEvents.ts
hwandersman: Not using these variables for any re-renders, so only need to manage the logic within the `useEffect`

packages/react-components/src/components/chart/baseChart.tsx
hwandersman: I see `eventOptions` is totally overwriting the `convertedOptions.tooltip`. Will merge the two tooltips.

mnischay (approved):

Issue #1930, opened 2023-09-14
Chart fit and finish - Improve chart legend UX


diehbria:
Current charts legend for line chart and scatter chart has large spacing between rows, and margins on the left and right. Want to Update to a more compact UX

UX Notes:

Fixed width for legends, legends text to be wrapped and the user can see the name of the property on-hover, remove the bottom scroll until it is necessary. Add consistent padding around the widget, try with 16px, 20px. Change border of legend table - gray 200, 2px

```[tasklist]
### UX
- [x] Provide specific guidance on desired spacing
```

```[tasklist]
### Engineering
- [ ] Update spacing/layout for legend (pre. req of UX guidance)
```

nadeswal:
UX specs -

No specific px value, however, I'd like the padding to be reduced to either 2px or 4px so we don't take a lot of space between the rows.

Also, make sure we make the text within the field wrapped.

PR #2118, opened 2023-10-18
feat: updated the chart legend ux(spacing, legend border color, width, tooltip if name truncated)

Chandru-HM:
## Overview
for the #1930 updated the chart legend ux and it includes
1. reduced the legend properties spacing
2. default width is increased for legend
3. tooltip added for the truncated names and properties names will truncate based on visibility width

image ref:
image
image

## Legal


packages/react-components/src/components/chart/hooks/useChartsLegend.tsx
```diff
@@ -47,10 +54,26 @@ const LegendCell = (e: { datastream: DataStream; lineColor: string; name: string
     </div>
   ));
 
+  useEffect(() => {
```
tracy-french, line 57: Is it necessary to use `useEffect` here? Can the name truncation be calculated each render, instead of using state?

Chandru-HM: updated. removed the useEffect and managed using variable.

mnischay (changes requested):
there a bunch of issues in your CR, can you please check

PR #2133, opened 2023-10-19
fix: synchronize requests after switching tabs

tracy-french:
## Overview
When the tab becomes de-focused, request intervals become de-synchronized, which leads to inefficient request batching and potentially throttling when visualizing a large number of data streams. This change ensures synchronization of data streams by resetting intervals when the tab goes back into focus.

## Legal


chejimmy:
LGTM! TY for fixing this tricky bug 👍

chejimmy (approved):

PR #2151, opened 2023-10-23
feat(scene): enable scene backgrounds

haweston:
## Overview
Update layout of fog settings to not require expander click and toggle color picker/distance inputs based on enable checkbox

Add scene background settings with color picker.  Does not support background textures.

https://github.com/awslabs/iot-app-kit/assets/109186219/533ea9f7-b88f-4986-9e0b-cdda28c39be4

## Verifying Changes
Use storybook to open a scene composer
visit settings
go to scene settings
use the enable fog and enable background to test setting and see the colors change in the 3d scene

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/scene-settings/FogSettingsEditor.tsx
+++ b/packages/scene-composer/src/components/panels/scene-settings/FogSettingsEditor.tsx
@@ -7,7 +7,6 @@ import { IFogSettings, KnownSceneProperty } from '../../../interfaces';
 import useLifecycleLogging from '../../../logger/react-logger/hooks/useLifecycleLogging';
 import { useStore } from '../../../store';
 import { ColorSelectorCombo } from '../scene-components/tag-style/ColorSelectorCombo/ColorSelectorCombo';
-import { ExpandableInfoSection } from '../CommonPanelComponents';
 
 export const FogSettingsEditor: React.FC = () => {
   useLifecycleLogging('FogSettingsEditor');
@@ -20,20 +19,20 @@ export const FogSettingsEditor: React.FC = () => {
   );
 
   const [fogEnabled, setFogEnabled] = useState(!!fogSettings);
-  const [internalColor, setInternalColor] = useState(fogSettings?.colorHex || 0xcccccc);
+  const [internalColor, setInternalColor] = useState(fogSettings?.color || '#cccccc');
   const [internalNearDistance, setInternalNearDistance] = useState(fogSettings?.near || 1);
   const [internalFarDistance, setInternalFarDistance] = useState(fogSettings?.far || 1000);
 
-  const tagStyleColors = useStore(sceneComposerId)((state) =>
-    state.getSceneProperty<string[]>(KnownSceneProperty.TagCustomColors, []),
+  const fogColors = useStore(sceneComposerId)((state) =>
+    state.getSceneProperty<string[]>(KnownSceneProperty.FogCustomColors, []),
   );
-  const setTagColorsSceneProperty = useStore(sceneComposerId)((state) => state.setSceneProperty<string[]>);
+  const setFogColorsSceneProperty = useStore(sceneComposerId)((state) => state.setSceneProperty<string[]>);
 
   const onToggleFog = useCallback(
     (checked: boolean) => {
       if (checked) {
         setSceneProperty(KnownSceneProperty.FogSettings, {
-          colorHex: internalColor,
+          color: internalColor,
           near: internalNearDistance,
           far: internalFarDistance,
         });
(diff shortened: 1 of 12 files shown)


packages/scene-composer/src/interfaces/interfaces.tsx
```diff
@@ -148,3 +149,7 @@ export interface IFogSettings {
   near: number;
   far: number;
 }
+
+export interface ISceneBackgroundSetting {
+  colorHex?: number;
```
sheilaXu, line 154: do we need to store the hex number or we can just store the hex string? looks like the color picker returns the string, and the threejs color can handle color string. and we will need to do some extra conversion between color number and string. can we just save the color hex string then?

I found many instances in our code that we allow color to be either number or string, then we add many conversion steps, but the actual value we get from the picker and set to the 3D component can just be a color string generically. then we can simplify some conversion code.

Same comment for the `colorHex` in `IFogSettings`

packages/scene-composer/src/components/panels/scene-settings/FogSettingsEditor.tsx
```diff
@@ -94,54 +93,53 @@ export const FogSettingsEditor: React.FC = () => {
...
+        </Checkbox>
+        {!!fogSettings && (
+          <>
+            <ColorSelectorCombo
+              color={`#${internalColor.toString(16).padStart(6, '0')}`}
+              onSelectColor={(pickedColor) => onColorChange(pickedColor)}
+              onUpdateCustomColors={(chosenCustomColors) =>
+                setTagColorsSceneProperty(KnownSceneProperty.TagCustomColors, chosenCustomColors)
+              }
+              customColors={tagStyleColors}
```
sheilaXu, line 112: is this intentional to update `TagCustomColors` when changing Fog color?

packages/scene-composer/src/components/panels/scene-settings/SceneBackgroundSettingsEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+          {intl.formatMessage({ defaultMessage: 'Enable Background', description: 'checkbox label' })}
+        </Checkbox>
+        {!!backgroundSettings && (
+          <ColorSelectorCombo
+            color={`#${internalColor.toString(16).padStart(6, '0')}`}
+            onSelectColor={(pickedColor) => onColorChange(pickedColor)}
+            onUpdateCustomColors={(chosenCustomColors) =>
+              setTagColorsSceneProperty(KnownSceneProperty.TagCustomColors, chosenCustomColors)
+            }
+            customColors={tagStyleColors}
```
sheilaXu, line 77: same here for whether scene background color should share the same custom colors list with tag

packages/scene-composer/src/components/panels/scene-settings/FogSettingsEditor.tsx
haweston: We don't have fog or background color lists from UX yet so this was a stop gap for the PoC to get feedback.  But I think we can make new fog and background custom colors lists now and just have them initialized to the same values as our tag colors but save custom colors off separately from tags.  Then when we get colors from UX we can just change the initialization and not have to change each editor as well.

packages/scene-composer/src/interfaces/interfaces.tsx
haweston: Let me give this ago, as I agree that the string to number conversion is annoying. I think when I looked at the three.js docs only saw hex values as numbers and strings as 'blue' etc. based on the examples but the full spec makes it look like it also takes strings of format '#ff00ff'

packages/scene-composer/src/components/panels/scene-settings/FogSettingsEditor.tsx
sheilaXu: the tag custom colors are a list of custom colors the user previously has selected. it is also initialized with an empty list

haweston: You're right, I was confusing it with the ColorSelectorCombos initial list of colors of non-custom colors which is set in TagColors.ts. But I'm breaking them up to all use their own custom save list even if the ColorSelectorCombo is still initialized from TagColors for non-custom colors.

packages/scene-composer/src/interfaces/interfaces.tsx
haweston: Swapped fog & scene background to use color: string

packages/scene-composer/src/components/panels/scene-settings/FogSettingsEditor.tsx
haweston: added fogCustom and backgroundCustom save system separate from tagCustom

packages/scene-composer/src/components/panels/scene-settings/SceneBackgroundSettingsEditor.tsx
haweston: see above, added backgroundCustom save separate from tagCustom

sheilaXu (approved):

PR #2156, opened 2023-10-24
fix(react-components): Trendcursors ymin max handling and styling fixes

mnischay:
## Overview

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/trendCursor/resize/useHandleResize.ts
+++ b/packages/react-components/src/components/chart/trendCursor/resize/useHandleResize.ts
@@ -1,42 +1,47 @@
-import { useRef } from 'react';
 import { calculateXFromTimestamp } from '../calculations/calculations';
 
 import { TrendCursorProps } from '../types';
 import { onResizeUpdateTrendCursorYValues } from './updateYValues';
 import { calculateSeriesMakers } from '../calculations/calculateSeriesMakers';
+import { useEffect, useRef } from 'react';
+import { delayedRender } from '../../utils/useDelayedRender';
 
 const useHandleResize = ({ size, series, graphic, setGraphic, chartRef, visualization }: TrendCursorProps) => {
-  const prevSize = useRef(size);
-  // to avoid unnecessary re-rendering
-  if (size.width !== prevSize.current.width || size.height !== prevSize.current.height) {
-    const newG = graphic.map((g) => {
-      // splitting into two separate if block to avoid calculations when not necessary
+  const graphicRef = useRef(graphic);
+  const seriesRef = useRef(series);
+  const sizeRef = useRef(size);
+  const visualizationRef = useRef(visualization);
+  const sizeString = JSON.stringify(size);
 
-      // if height has changed, update Y values
-      if (size.height !== prevSize.current.height) {
+  useEffect(() => {
+    graphicRef.current = graphic;
+    seriesRef.current = series;
+    sizeRef.current = size;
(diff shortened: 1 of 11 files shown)


packages/react-components/src/components/chart/trendCursor/resize/useHandleResize.ts
```diff
@@ -7,36 +7,33 @@ import { calculateSeriesMakers } from '../calculations/calculateSeriesMakers';
...
 const useHandleResize = ({ size, series, graphic, setGraphic, chartRef, visualization }: TrendCursorProps) => {
   const prevSize = useRef(size);
-  // to avoid unnecessary re-rendering
-  if (size.width !== prevSize.current.width || size.height !== prevSize.current.height) {
-    const newG = graphic.map((g) => {
-      // splitting into two separate if block to avoid calculations when not necessary
 
-      // if height has changed, update Y values
-      if (size.height !== prevSize.current.height) {
+  // delaying the updating by 10ms, so that the graph is drawn before reading the values back
```
tracy-french, line 11: What guarantee do you have that the graph will be drawn first? Are we using the setTimeout to control the event loop?

packages/react-components/src/components/chart/trendCursor/seriesYMinMax/useHandleSeriesYMinMax.ts
```diff
@@ -0,0 +1,54 @@
...
+        g.timestampInMs,
+        chartRef,
+        visualizationRef.current
+      );
+      g.yAxisMarkerValue = trendCursorsSeriesMakersValue;
+      g.children = updateTrendCursorLineMarkers(g.children, trendCursorsSeriesMakersInPixels);
+      return g;
+    });
+    setGraphic([...newG]);
+  }, [yAxisOptionsString, setGraphic, chartRef, seriesString]);
```
tracy-french, line 53: Why are we using useEffect to update the state? Is there no event occurring we can call attach a callback to?

mnischay: nope, the function inside the useEffect needs to run when the user/api updates the ymin or ymax or series... the change in value itself is an event within the component,,, we can do use the prev reference to achieve this probably, but useEffect made the code look clean.... open for suggestions

packages/react-components/src/components/chart/trendCursor/resize/useHandleResize.ts
mnischay: no guarantee, i just need one render cycle... 10ms is an arbitrary time to let one render cycle pass...

tracy-french: Seems like an area which may result in bugs in the future. You might be able to get around specifying a time to wait by using https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame.

tracy-french (approved):

packages/react-components/src/components/chart/trendCursor/resize/useHandleResize.ts
mnischay: but requestAnimationFrame lets me know how many times something got render or will render... I cannot skip a frame ? hmmm, let me try.. If there is a alternate solution to using settimeout il take it

corteggiano (approved):
discussed on call and reviewed ✅

PR #2157, opened 2023-10-24
chore: hoist and upgrade cloudscape

tracy-french:
## Overview
This change updates our version of Cloudscape to get access to all the great updates, including `<SpaceBetween />` supporting item alignment, allowing us to remove a lot of custom CSS and divs.

As part of this change, the Cloudscape packages were hoisted to the root and individual package declarations were removed.

## Legal


package.json
```diff
@@ -36,6 +36,12 @@
     "versionup:major": "turbo run version --no-git-tag-version major",
     "prepare": "husky install"
   },
+  "dependencies": {
+    "@cloudscape-design/collection-hooks": "1.0.22",
```
mnischay, line 40: why are we nailing it down to a specific version ? i.e no ~ ?

tracy-french: Cloudscape doesn't follow SemVer and they release breaking changes in patch version bumps. I figured this was safer.

chejimmy (approved):
Nice work! 🙌

PR #2159, opened 2023-10-24
fix: do not override selected data stream preferences

tracy-french:
## Overview
This change fixes a bug where data stream preferences are lost when added new data streams to a widget. 

## Legal


packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/queryExtender/queryExtender.ts
```diff
@@ -0,0 +1,99 @@
+import type { ModeledDataStream } from '../modeledDataStreamQueryEditor/modeledDataStreamExplorer/types';
+import type { UnmodeledDataStream } from '../unmodeledDataStreamExplorer/types';
+import type { IoTSiteWiseDataStreamQuery } from '~/types';
+
+type Query = IoTSiteWiseDataStreamQuery;
+
+export class QueryExtender {
```
mnischay, line 7: Niiiceeeeeeee

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/queryExtender/queryExtender.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  #createAssetQueries(modeledDataStreams: ModeledDataStream[]): NonNullable<Query['assets']> {
+    const assetQueriesWithProperties = modeledDataStreams
+      .filter(this.#isModeledDataStream)
+      .map<NonNullable<Query['assets']>[number]>(({ assetId, propertyId: propertyId }) => ({
+        assetId,
+        properties: [{ propertyId }],
+      }));
+
+    return assetQueriesWithProperties;
```
mnischay, line 34: nit: you do not need this variable

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/queryExtender/queryExtender.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  #isModeledDataStream(dataStream: ModeledDataStream | UnmodeledDataStream): dataStream is ModeledDataStream {
```
mnischay, line 37: arrow functions are better ? so that you do not need to bind to this explicitly

mnischay (commented):
Code changes LGTM, I am curious about not using arrow function in the class

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/queryExtender/queryExtender.ts
tracy-french: I default to using method declarations instead of arrow functions in classes as method declarations are shared across class instances and arrow functions are not. This can cause performance problems at scale, as every class instance creates its own instance of the methods.

This rationale doesn't apply to this class, as there are not many instances being created, but I would rather not think about it, so I just default to method declarations.

I still use arrow functions in classes to handle to avoid the binding problem. It's rare I have a binding issue that needs to be handled. If I was using the method as a callback, I might reach for an arrow function, especially if I am handing control to the browser with an event listener, setTimeout, or other browser API.

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/queryExtender/queryExtender.ts
tracy-french: I typically do this for two reasons: 1. To make my data more descriptive. Sometimes it helps. Sometimes it doesn't. 2. To make it easier to debug with console.log or a debugger. I can drop a console.log without changing the code. When I use the debugger, I can reference the variable name.

mnischay: true, thats what I do as well when debugging, I guess its nice to follow one pattern but I can see this is more helpful to read the code

mnischay (approved):

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/queryExtender/queryExtender.ts
tracy-french: It's also expensive to instantiate a class with multiple arrow functions. When you call `new`, you create all the functions on the spot. What's the point of a class if you don't utilize its ability to act as a template? 😛

tracy-french:
Thank you, @mnischay!

PR #2186, opened 2023-10-26
fix(composer): new tag style field is empty instead of info

MO-Elmu:
## Overview
In this PR Tag style field for newly created tag adjusted  to show Info, since newly created tag uses DefaultAnchorStatus.Info icon anyway so this correctly reflect the default icon that the code default to. No point of hiding that in the UI if we wanted a different Default or new tag creation workflow that should be addressed comprehensively in code.

Before

https://github.com/awslabs/iot-app-kit/assets/20994167/af1cb83c-0e5a-422a-babf-58802052d5a3

After

https://github.com/awslabs/iot-app-kit/assets/20994167/fe47650c-d811-41c9-95de-440f1f89d1fd

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


MO-Elmu:
I will make an adjustment and push again. Snapshot test is failing and looking into another approach.

MO-Elmu:
Done

mukeshsahay (approved):

PR #2187, opened 2023-10-26
fix(scene): stop scene composer from forcing dark mode on page

haweston:
## Overview
Currently the scene viewer only support dark mode, and it forces it's mode on the whole page including other cloudscape packages.
lightModeNoFix

## Verifying Changes
Build example/react-examples but change the src/index.tsx line 14 to applyMode(Mode.Light) then start the server.
The page when visited should have the dark mode scene composer, but the rest of the page in light mode.
lightModeWithFix

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/SceneComposerInternal.tsx
+++ b/packages/scene-composer/src/components/SceneComposerInternal.tsx
@@ -1,6 +1,6 @@
-import React, { useCallback, useEffect, useMemo, useReducer } from 'react';
+import React, { useCallback, useMemo, useReducer, useRef } from 'react';
 import { ThemeProvider } from 'styled-components';
-import { applyMode, Mode } from '@awsui/global-styles';
+import { Mode } from '@awsui/global-styles';
 import { cloneDeep } from 'lodash';
 import { useViewport } from '@iot-app-kit/react-components';
 
@@ -8,6 +8,7 @@ import { SCENE_BODY_CLASS } from '../common/constants';
 import { sceneComposerIdContext } from '../common/sceneComposerIdContext';
 import LogProvider from '../logger/react-logger/log-provider';
 import { GlobalStyles } from '../GlobalStyles';
+import useAwsLightDarkModes from '../hooks/useAwsLightDarkModes';
 import { KnownComponentType, SceneComposerInternalProps, StyleTarget } from '../interfaces';
 import { materialReducer, initialMaterialMaps, addMaterial, removeMaterial, backUpOriginalMaterial } from '../reducers';
 import { useStore } from '../store';
@@ -28,6 +29,7 @@ export const SceneComposerInternal: React.FC<SceneComposerInternalProps> = ({
   config,
   ...props
 }: SceneComposerInternalProps) => {
+  const sceneComposerInternalRef = useRef<HTMLDivElement>(null);
   const currentSceneComposerId = useMemo(() => sceneComposerId ?? generateUUID(), [sceneComposerId]);
   const { viewport } = useViewport();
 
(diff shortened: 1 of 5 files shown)


packages/scene-composer/src/hooks/useAwsLightDarkModes.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Mode } from '@awsui/global-styles';
+
+import useAwsLightDarkModes from './useAwsLightDarkModes';
+
+describe('useAwsLightDarkModes', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('it should apply light mode', () => {
```
mukeshsahay, line 18: NIT: should be changed to it('should apply light mode', ....
`it` in the quotes is redundant when you read the code

packages/scene-composer/src/hooks/useAwsLightDarkModes.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const divRef = createRef<HTMLDivElement>();
+    render(<div ref={divRef}></div>);
+    renderHook(() => useAwsLightDarkModes(divRef, Mode.Light));
+    expect(applyModeMock).toBeCalledWith(Mode.Light, divRef.current);
+  });
+  it('it should apply dark mode', () => {
```
mukeshsahay, line 24: NIT: should be changed to it('should apply light mode', ....
`it` in the quotes is redundant when you read the code

packages/scene-composer/src/components/SceneComposerInternal.tsx
```diff
@@ -39,25 +41,26 @@ export const SceneComposerInternal: React.FC<SceneComposerInternalProps> = ({
...
   document.body.className = document.body.className + ' ' + SCENE_BODY_CLASS;
 
-  useEffect(() => {
-    if (config.colorTheme === 'light') {
-      applyMode(Mode.Light);
-    } else {
-      applyMode(Mode.Dark);
-    }
-  }, [config.colorTheme]);
+  useAwsLightDarkModes(sceneComposerInternalRef, config.colorTheme === 'light' ? Mode.Light : Mode.Dark);
```
mukeshsahay, line 44: NIT: 'light' can be replaced with Mode.Light
There is usage of hard-coded 'light' at line#39 as well.

mukeshsahay (approved):
Approved with a few nice to have comments.

packages/scene-composer/src/hooks/useAwsLightDarkModes.spec.tsx
haweston: fixed

packages/scene-composer/src/hooks/useAwsLightDarkModes.spec.tsx
haweston: fixed

packages/scene-composer/src/components/SceneComposerInternal.tsx
haweston: I would except that's an external API change where config.colorTheme is using colorTheme?: 'dark' | 'light'; not colorTheme?: Mode.  I'd like us to look at changing that when we upgrade the composer to full light mode support?

mukeshsahay: I was suggesting that in this validation `config.colorTheme === 'light' ` instead of using a hard-coded text 'light', we could use a constant... maybe Mode.Light as its value is 'light'.
Hard-coded text 'light' is used at line #39 as well.

mukeshsahay: Got your point.. It is ok for now 👍

PR #2196, opened 2023-10-30
feat: format datetime in resource explorer table

Chandru-HM:
## Overview
This PR is to format datetime values in the resource explorer table to MM/DD/YY HH:MM:SS format.

corrent format:
image

new update:
image

## Legal


mnischay (commented):
Probably need the timezone, we cannot show time without saying somewhere which timezone it is

packages/dashboard/src/components/util/dateTimeUtil.tsx
```diff
@@ -0,0 +1,7 @@
+// Format Date object to get date and time to display MM/DD/YY HH:MM:SS format
+export const getFormattedDateTime = (rawDate: Date) => {
```
mnischay, line 2: Mind checking if we use a date formatter already ?

Chandru-HM: yes, using in react-components but not with same format. here added this to get in MM/DD/YY HH:MM:SS

mnischay (approved):

PR #2200, opened 2023-10-30
fix(react-components): echarts grab on canvas update cursor and tooltip

hwandersman:
## Overview
My previous [commit](https://github.com/awslabs/iot-app-kit/commit/a29da3a08a769137610bc37efde5605bf6b62dc2) was reverted because you could not add/edit trend cursors after changing any chart property.

This bug was fixed by removing the check for a chart ref change in the `useTrendCursorsEvents` hook. The `useEffect` setting up the event handlers were being turned off on a re-render, and not turned back on.

Added back my code with the general chart drag events with shift+click.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/87385528/1c255e75-c21c-47b2-aeee-bd1ac3d8d099

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/events/useHandleChartEvents.ts
+++ b/packages/react-components/src/components/chart/events/useHandleChartEvents.ts
@@ -1,0 +1,135 @@
+import { MutableRefObject, useCallback, useEffect, useState } from 'react';
+import { EChartsOption, EChartsType } from 'echarts';
+import { KeyMap } from 'react-hotkeys';
+
+export const useHandleChartEvents = (chartRef: MutableRefObject<EChartsType | null>) => {
+  const [chartEventsOptions, setChartEventsOptions] = useState<EChartsOption>({});
+
+  const [shiftDown, setShiftDown] = useState(false);
+  const [mouseDown, setMouseDown] = useState(false);
+  const [prevIsGrabbing, setPrevIsGrabbing] = useState(false);
+
+  // When dragging on the chart, turn off tooltip and set cursor to grabbing
+  const setOnDragStart = (chart: EChartsType | null) => {
+    setChartEventsOptions({
+      tooltip: {
+        show: false,
+      },
+    });
+    chart?.getZr().setCursorStyle('grabbing');
+  };
+
+  // When dragging is done turn back on the tooltip and set cursor to grab
+  // Use when the shift key is still held down but mouse is up
+  const setOnDragEnd = (chart: EChartsType | null) => {
+    setChartEventsOptions({
+      tooltip: {
+        show: true,
+      },
+    });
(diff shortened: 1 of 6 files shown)


packages/react-components/src/components/chart/hooks/useHandleChartEvents.ts
```diff
@@ -0,0 +1,113 @@
+import { MutableRefObject, useEffect, useState } from 'react';
```
mnischay, line 1: instead of hooks, probably worth creating a Events folder under chart

mnischay (commented):
Code changes LGTM, except the file placement

packages/react-components/src/components/chart/events/useHandleChartEvents.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    // When shift is released go back to the default cursor
+    const shiftUpListener = (event: KeyboardEvent) => {
+      if (event.key.toLowerCase() === 'shift') {
+        shiftDown = false;
+        chart?.getZr().setCursorStyle('default');
+      }
+    };
+
+    window.addEventListener('keydown', shiftDownListener);
```
mnischay, line 56: curious : the event handler should be on the chart right ? why on the entire window ?

hwandersman: echarts does not have an event interface for `keydown` or `keyup` so I needed to define my own.

mnischay: curious: could we leverage HotKeys which is already integrated ? my concern is that this handler will get triggered anytime the user presses shift ... i.e it is not specific to charts ,,

hwandersman: Can you give me more info about hot keys? This handler will get triggered on shift, but it only impacts the xyplot chart. If you have multiple of those charts on your dashboard then hovering around with shift will set the `grab` cursor style on those charts. This doesn't impact anything else.

mnischay: https://github.com/awslabs/iot-app-kit/blob/main/packages/react-components/src/components/chart/baseChart.tsx#L153 --> take a look at this, its probably better to use this instead of a global event handler ? 

this is the library https://www.npmjs.com/package/react-hotkeys

hwandersman: Updated to use hotkeys after discussing offline

PR #2207, opened 2023-10-31
feat(scene-composer): support tag occlusion

mukeshsahay:
## Overview
Issue: Tags clutter the UI when they are always visible through other Meshes

This is especially an issue when you are in immersive mode, like Matterport. Because tags are always rendered closer to the camera than any meshes, large sites with lots of tags become very difficult to navigate in immersive mode.

Fix: Provided an option to show/hide tags through objects in the Scene Tag Settings.

Default behabior:
ShowTagThroughObjects-Default

Tag Occlusion Enabled:
ShowTagThroughObjects-Disabled

Screen recording:
https://github.com/awslabs/iot-app-kit/assets/43108986/9c2c4cf3-9ce5-49d2-b808-263aa24797b0

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/common/constants.ts
+++ b/packages/scene-composer/src/common/constants.ts
@@ -131,6 +131,7 @@ export const DRACO_PATH = 'https://www.gstatic.com/draco/versioned/decoders/1.4.
 export const DEFAULT_TAG_GLOBAL_SETTINGS: ITagSettings = {
   autoRescale: false,
   scale: 1,
+  enableOcclusion: false,
 };
 
 export const DEFAULT_OVERLAY_GLOBAL_SETTINGS: IOverlaySettings = {
--- a/packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
+++ b/packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useContext, useEffect, useState } from 'react';
 import { useIntl } from 'react-intl';
-import { Checkbox, FormField, Grid, Input } from '@awsui/components-react';
+import { Checkbox, FormField, Grid, Input, SpaceBetween, Toggle } from '@awsui/components-react';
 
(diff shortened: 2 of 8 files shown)


packages/scene-composer/src/common/constants.ts
```diff
@@ -131,6 +131,7 @@ export const DRACO_PATH = 'https://www.gstatic.com/draco/versioned/decoders/1.4.
 export const DEFAULT_TAG_GLOBAL_SETTINGS: ITagSettings = {
   autoRescale: false,
   scale: 1,
+  disableOcclusion: true,
```
sheilaXu, line 134: when the customer has `autoRescale` or `scale` configured, this default settings will not be used, and the behavior will be disableOcclusion = false and changed existing behavior. you could revert the meaning of this new flag and make sure the behavior is not changed when the value is `false`

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -96,45 +96,55 @@ export const SceneTagSettingsEditor: React.FC = () => {
...
+              onMouseDown={() => {
+                setDraggingSlider(true);
+              }}
+              onChange={onSliderChange}
+            />
+          )}
+        </FormField>
+        <Toggle
+          checked={!!tagSettings.disableOcclusion}
+          onChange={() => {
```
sheilaXu, line 141: it should return a `checked` value? using it to set the new disableOcclusion value will have less race condition in case disableOcclusion is changed somewhere else

sheilaXu (approved):

packages/scene-composer/src/common/constants.ts
mukeshsahay: Done

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
mukeshsahay: Done

PR #2208, opened 2023-10-31
feat(react-components): hide/show properties from legend

ssjagad:
## Overview
Allow users to toggle the visibility of a property from within the legend. 

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/145582655/f9fa0e01-9866-4bfb-b3c6-1fae7f2a8757

## Legal


dpportet (commented):
This unit test is failing: `src/components/chart/trendCursor/tests/useVisualizedDataStreams.spec.ts` which looks related to your change.

packages/react-components/src/components/chart/legend/icons.tsx
mnischay, line 1: this should be a .svg file

ssjagad: Importing a svg file makes it a react functional component, which is not accepted by Button in cloudscape. Did it this way to ensure it has type of React.ReactNode

mnischay: I am not entirely sure what you mean ... hmmmm use iconUrl ? or you can just render a span with icon inside an button

packages/react-components/src/components/chart/legend/useChartsLegend.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useEffect, useRef, useState } from 'react';
 import { TableProps } from '@cloudscape-design/components/table/interfaces';
 import { GraphicComponentTextOption } from 'echarts/types/src/component/graphic/GraphicModel';
 import { SeriesOption } from 'echarts';
-import { LineSeriesOption } from 'echarts/types/src/chart/line/LineSeries';
+// import { LineSeriesOption } from 'echarts/types/src/chart/line/LineSeries';
```
mnischay, line 5: nit: lets remove this ?

packages/react-components/src/components/chart/store/contextDataStreams.ts
```diff
@@ -0,0 +1,39 @@
...
+  highlightDataStream: (datastream?: DataStream) => void;
+  unHighlightDataStream: (datastream?: DataStream) => void;
+  hideDataStream: (datastream?: DataStream) => void;
+  unHideDataStream: (datastream?: DataStream) => void;
+}
+
+export const createDataStreamsSlice: StateCreator<DataStreamsState> = (set) => ({
+  highlightedDataStreams: [],
+  hiddenDataStreams: [],
+  highlightDataStream: (datastream?: DataStream) =>
```
mnischay, line 19: curious: is there a scenario where dataStream is undefined and this action needs to be called ?

packages/react-components/src/components/chart/store/contextDataStreams.ts
```diff
@@ -0,0 +1,39 @@
+import { DataStream } from '@iot-app-kit/core';
+import { StateCreator } from 'zustand';
+
+export interface DataStreamsData {
+  highlightedDataStreams: DataStream[];
+  hiddenDataStreams: DataStream[];
+}
+
+export interface DataStreamsState extends DataStreamsData {
+  highlightDataStream: (datastream?: DataStream) => void;
```
mnischay, line 10: Nice! i like that you are not re-using the same action to perform 2 dofferent things

packages/react-components/src/components/chart/trendCursor/tests/useVisualizedDataStreams.spec.ts
```diff
@@ -9,7 +9,7 @@ it('converts empty query to empty data stream', async () => {
     result: { current },
```
mnischay, line 9: worth adding some tests for the unhide/ hide  and highlighting operations ?

mnischay (changes requested):
I think the icons should be in their own .svg files, please take a look at the other examples. other comments are just minor nits

packages/react-components/src/components/chart/store/contextDataStreams.ts
ssjagad: When we filter the query based on hidden or not, there are times that the query passed into get highlighted datastreams is undefined. Its also a good check for if there are no datastreams.

packages/react-components/src/components/chart/trendCursor/tests/useVisualizedDataStreams.spec.ts
ssjagad: I agree. Going to cut a ticket for this because I really need to start on composition API for re:Invent and revisit it.

packages/react-components/src/global.d.ts
```diff
@@ -2,6 +2,6 @@ declare module '*.module.css';
 declare module '*.module.scss';
 declare module 'cytoscape-cise';
 declare module '*.svg' {
-  const content: React.FunctionComponent<React.SVGAttributes<SVGElement>>;
-  export default content;
+  const src: string;
```
mnischay, line 5: just make sure this is not breaking the existing svg imports

packages/react-components/src/components/chart/chartOptions/seriesAndYAxis/convertSeriesAndYAxis.ts
```diff
@@ -68,9 +68,10 @@ const convertSeries = (
     lineThickness,
     emphasis,
     significantDigits,
+    hidden,
   }: ChartStyleSettingsWithDefaults
 ) => {
-  const opacity = emphasis === 'de-emphasize' ? DEEMPHASIZE_OPACITY : 1;
+  const opacity = hidden ? 0 : emphasis === 'de-emphasize' ? DEEMPHASIZE_OPACITY : 1;
```
corteggiano, line 74: could we pull this out into a lil function without the nested ternary operator for legibility :)

packages/react-components/src/components/chart/trendCursor/getTrendCursor/components/markers/index.ts
```diff
@@ -21,7 +21,7 @@ const addTCMarker = ({
   return {
     id: `circle-${index}-${uId}`,
     type: 'circle',
-    ignore: false,
+    ignore: (series[index] as LineSeriesOption).lineStyle?.opacity === 0 ? true : false,
```
corteggiano, line 24: when an `if` statement returns `true` else `false`
its a sign we dont need the if! 

this can be simplified to `(series[index] as LineSeriesOption).lineStyle?.opacity === 0`

packages/react-components/src/components/chart/trendCursor/getTrendCursor/components/markers/index.ts
```diff
@@ -94,6 +94,7 @@ export const addTCMarkers = (uId: string, yAxisMarkers: number[], series: Series
     type: 'circle',
     z: TREND_CURSOR_Z_INDEX + 1,
     y: marker,
+    ignore: (series[index] as LineSeriesOption).lineStyle?.opacity === 0 ? true : false,
```
corteggiano, line 97: same comment as above ^

packages/react-components/src/components/chart/trendCursor/series/useHandleSeries.ts
```diff
@@ -44,5 +44,5 @@ export const useHandleSeries = ({
     };
 
     delayedRender({ updateFunction: update });
-  }, [chartRef, series.length, setGraphic]);
+  }, [chartRef, series, setGraphic]);
```
corteggiano, line 47: why the change from series.length?

packages/react-components/src/components/chart/utils/getStyles.ts
```diff
@@ -12,7 +12,7 @@ export type ChartStyleSettingsWithDefaults = Omit<
   Required<ChartStyleSettingsOptions>,
   keyof OptionalChartStyleSettingsOptions
 > &
-  OptionalChartStyleSettingsOptions & { emphasis: Emphasis };
+  OptionalChartStyleSettingsOptions & { emphasis: Emphasis } & { hidden: boolean };
```
corteggiano, line 15: is there a reason `hidden: boolean` is not part of the emphasis object?

packages/react-components/src/components/chart/legend/useChartsLegend.tsx
```diff
@@ -119,13 +145,23 @@ const useChartsLegend = ({
...
+  };
+
   useEffect(() => {
     const tcColumnDefinitions = graphic.map((g) => {
       const id = g.id as string;
       return {
         id,
         header: getHeaderNode(g),
-        cell: (e: { [x: string]: number }) => e[id],
+        cell: (e: { [x: string]: number | string | DataStream }) => <TcCell {...e} tcId={id} />,
```
corteggiano, line 164: could we make the type of `e` less generic?

packages/react-components/src/components/chart/trendCursor/series/useHandleSeries.ts
ssjagad: We change the actual series when we add the hidden property. To trigger the re-render of all the graphics we need to check that the series changed.

corteggiano: im worried that this will cause rerenders on the trendcursors every time anything in the series changes including styles (color, line wdith, etc.) which shouldnt be triggering a rerender

packages/dashboard/src/customization/widgets/types.ts
```diff
@@ -72,7 +72,6 @@ export type LineAndScatterStyles = {
   line?: LineStyles;
   aggregationType?: AggregateType;
   resolution?: string;
-  visible?: boolean; // defaults to true
```
corteggiano, line 75: this is a public API change

ssjagad: We discussed this change when this ticket was created. Since it is optional, we decided to remove it completely.

ssjagad: https://quip-amazon.com/KgFkAMKfbT8n/PFR-27-HideShow-property-Feature-requirement

packages/react-components/src/components/chart/trendCursor/series/useHandleSeries.ts
```diff
@@ -44,5 +45,5 @@ export const useHandleSeries = ({
     };
 
     delayedRender({ updateFunction: update });
-  }, [chartRef, series.length, setGraphic]);
+  }, [chartRef, seriesString, setGraphic]);
```
corteggiano, line 48: just tested locally and this change causes TCs to rerender a ton of times if any part of the series changes (color, line style, symbol, y axis)

corteggiano: could we keep series.length and also do number of series with an opacity of 0?

corteggiano (approved):

PR #2219, opened 2023-11-02
fix(react-components): pagination can move forward on first click from relative range

ssjagad:
## Overview
Can move forward from relative range. 

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/145582655/a105d49f-48c8-4557-8140-c1e928eec570

## Legal

### DIFF
--- a/packages/react-components/src/components/time-sync/viewportAdapter.tsx
+++ b/packages/react-components/src/components/time-sync/viewportAdapter.tsx
@@ -120,30 +120,33 @@ export const viewportToDateRange = (viewport?: Viewport): DateRangePickerProps.V
   return null;
 };
 
-export const getViewportStartOnBackwardRelative = (value: DateRangePickerProps.RelativeValue): Date => {
-  const newStart = new Date();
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/time-sync/viewportAdapter.tsx
```diff
@@ -147,3 +147,31 @@ export const getViewportStartOnBackwardRelative = (value: DateRangePickerProps.R
   }
   return newStart;
 };
+
+export const getViewportStartOnFowardRelative = (value: DateRangePickerProps.RelativeValue): Date => {
```
mnischay, line 151: this and the above getViewportStartOnBackwardRelative loooks the same ?

mnischay: probably worth just having the something like , `value.amount  * forward ? 1 : -1 `

ssjagad: got it

mnischay (approved):

PR #2223, opened 2023-11-02
fix(scene): only show selected tags overlay

haweston:
## Overview
When using the option to not display all overlays, overlays staying open when not selected can make the screen cluttered.  We want to make overlays only visible if their tag is selected OR we're in display all mode.

## Verifying Changes

Use local storybook composer 
have a scene with 2 tags
ensure that when overlay visible setting is toggled on their both visible.
ensure that when overlay visible setting is toggled off that the overlay is only visible if the tag is selected.

https://github.com/awslabs/iot-app-kit/assets/109186219/b04eee73-d9e5-4436-941d-e6bbc5310e68

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/__tests__/__snapshots__/DataOverlayContainerSnap.spec.tsx.snap
+++ b/packages/scene-composer/src/components/three-fiber/DataOverlayComponent/__tests__/__snapshots__/DataOverlayContainerSnap.spec.tsx.snap
@@ -48,15 +48,6 @@ exports[`DataOverlayContainer should render with panel visible correctly when th
     style="overflow: auto; border-radius: 2px; box-shadow: 0px 1px 4px -2px rgba(0, 28, 36, 0.5); padding-top: 12px; padding-bottom: 12px; width: 210px; min-height: 2em;"
   >
     <div
-      style="float: right; margin-top: 8px; margin-right: 8px;"
-    >
-      <button
-        style="background-color: transparent; border-color: transparent; color: white;"
-      >
-        X
-      </button>
-    </div>
-    <div
       data-testid="rows"
     >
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/components/three-fiber/DataOverlayComponent/__tests__/__snapshots__/DataOverlayContainerSnap.spec.tsx.snap
```diff
@@ -47,15 +47,6 @@ exports[`DataOverlayContainer should render with panel visible correctly when th
   <div
     style="overflow: auto; border-radius: 2px; box-shadow: 0px 1px 4px -2px rgba(0, 28, 36, 0.5); padding-top: 12px; padding-bottom: 12px; width: 210px; min-height: 2em;"
   >
-    <div
```
mukeshsahay, line 50: If we remove this, then user won't be able to close the overlay if they want to. 
Is this approved by UX?

haweston: The selection logic closes it instead.  That button is only present when 'Show All Overlays' isn't active, which in the new logic those overlays automatically close when their tag is no longer selected. So you now close by selecting something else instead of using the button. I did show the video without buttons to UX and call out that specific change.

sheilaXu: did we decide to change this behavior without a config that customer can turn on/off? also can clean up the styles for the close button

haweston: Discussed not needing 3rd option. Removing dead css.

sheilaXu (approved):

PR #2228, opened 2023-11-06
perf: use requestAnimationFrame when dragging widgets on the dashboard

tracy-french:
## Overview
This change improves the performance when dragging widgets around the dashboard by ensuring the dragging movement is aligned with browser repainting using requestAnimationFrame.

### Before change
https://github.com/awslabs/iot-app-kit/assets/67283114/5d913a56-0bee-4d11-a5da-7f9f20e6bea1

### After change
https://github.com/awslabs/iot-app-kit/assets/67283114/8530fed4-45c9-46fb-875c-64705bd8b12b

## Legal

### DIFF
--- a/packages/dashboard/src/components/grid/gestures/useDragMonitor.ts
+++ b/packages/dashboard/src/components/grid/gestures/useDragMonitor.ts
@@ -75,6 +75,9 @@ export const useDragMonitor = ({
   const { isDragging, clientOffset } = collected;
 
   useEffect(() => {
+    // We capture the current animation frame to enable cleanup on unmount
+    let animationFrameId: number;
+
     if (isDragging && clientOffset) {
       if (!dashboardGrid) return;
       const constrainedOffset = constrainPosition({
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/components/grid/gestures/useDragMonitor.ts
```diff
@@ -33,6 +33,8 @@ export const useDragMonitor = ({
   enabled,
   dashboardGrid,
 }: DragMonitorProps) => {
+  const rafRef = useRef<number>();
```
mnischay, line 36: do you need a ref here ? i.e 
``` const anyVar = requestAnimationFrame() ```

tracy-french: Yeah, we need a ref here so that we maintain the reference to the current animation frame after the re-render. Without this, we wouldn't be able to cancel the right animation frame at unmount.

tracy-french: Oh, no, you're right. We can avoid it since everything can be contained in the effect. I'll change that!

mnischay (approved):

PR #2229, opened 2023-11-06
feat(dashboard): describe unmodeled data stream

corteggiano:
## Overview
good old async/await request for unmodeled data stream descriptions

## Legal

### DIFF
--- a/packages/dashboard/src/components/csvDownloadButton/getDescribedTimeSeries.tsx
+++ b/packages/dashboard/src/components/csvDownloadButton/getDescribedTimeSeries.tsx
@@ -1,0 +1,35 @@
+import { DescribeTimeSeriesCommand, IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
+import invariant from 'tiny-invariant';
+
+const isEnabled = (input?: string): input is string => Boolean(input);
+
+export const getDescribedTimeSeries = async ({
+  client,
+  assetId,
+  propertyId,
+  alias,
+}: {
+  client: IoTSiteWiseClient;
+  assetId?: string;
+  propertyId?: string;
+  alias?: string;
+}) => {
+  invariant(
+    isEnabled(alias) || (isEnabled(assetId) && isEnabled(propertyId)),
+    'Expected alias or assetID+propertyID to be defined as required by the enabled flag.'
+  );
+
+  const params = { assetId, propertyId, alias };
+  const command = new DescribeTimeSeriesCommand(params);
+
+  try {
+    const data = await client.send(command);
+    return { data, isSuccess: true };
+  } catch (error) {
+    console.error(`Failed to get described time series. Error: ${error}`);
+    console.info('Request input:');
+    console.table(params);
+
+    throw error;
+  }
+};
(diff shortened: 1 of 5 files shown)


packages/dashboard/src/components/csvDownloadButton/getDescribedTimeSeries.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: {
+  client: IoTSiteWiseClient;
+  assetId?: string;
+  propertyId?: string;
+  alias?: string;
+}) => {
+  const isModeledData = assetId && propertyId;
+
+  if (!isModeledData && !alias) {
+    return {};
```
corteggiano, line 17: this situation shouldnt happen unless there is an issue with the data so i will throw an error here instead

jmbuss: I think you can use this library Tracy introduced with his resource explorer change to guard against this - `import invariant from 'tiny-invariant';`

jmbuss (commented):
One small thing!

packages/dashboard/src/components/csvDownloadButton/getDescribedTimeSeries.tsx
corteggiano: oh yes! ty! had used this in the v1 but completely forgot

packages/dashboard/src/components/csvDownloadButton/getDescribedTimeSeries.tsx
```diff
@@ -0,0 +1,35 @@
...
+  );
+
+  const params = { assetId, propertyId, alias };
+  const command = new DescribeTimeSeriesCommand(params);
+
+  try {
+    const data = await client.send(command);
+    return { data, isSuccess: true };
+  } catch (error) {
+    console.error(`Failed to get latest values. Error: ${error}`);
```
mnischay, line 29: you probably want to log the error in all 3 cases ?

mnischay (approved):

packages/dashboard/src/components/csvDownloadButton/getDescribedTimeSeries.tsx
corteggiano: im following the same error logging as the ones found the resource explorer
which have the following definitions
```
#handleError(error: unknown): never {
    console.error(`Failed to get asset models. Error: ${error}`);
    console.info('Request input:');
    console.table(this.#command.input);

    throw error;
}
```

packages/dashboard/src/components/csvDownloadButton/useViewportData.ts
```diff
@@ -46,34 +49,41 @@ export const useViewportData = ({
...
       // do not include data point if it falls outside viewport range
       if (pointWithinViewport) {
-        const isUnmodeledData = queryConfig.query?.properties?.some((pr) => pr.propertyAlias === dataStream.id);
         const assetPropId = !isUnmodeledData ? dataStream.id.split('---') : []; // modeled datastream IDs follow the pattern {assetID}---{propertyID}
         const describedModelProperty = describedAssetsMap[assetPropId[0]]?.properties.find(
-          (p) => p.propertyId === assetPropId[1]
+          ({ propertyId }) => propertyId === assetPropId[1]
         );
 
         const flatDataPoint: CSVDownloadObject = {
```
mnischay, line 73: super nit: instead of having inline ternary ops. better to have something lie ` if(unmodeled) { flatDataPoint = {..common, ...} else { flatDataPoint = {...common, ... }}}`

mnischay (approved):

packages/dashboard/src/components/csvDownloadButton/useViewportData.ts
corteggiano: great idea, will update code to follow this pattern
will def be much cleaner to read

jmbuss (approved):

Issue #2141, opened 2023-10-20
[Dashboard] pasting at edges of dashboard causes invalid selection states


diehbria:
## Description:
- add widget
- right click widget, select copy
- at the right edge of the dashboard, right click and hit paste

**Observe** below, the selection state does not align with the newly pasted widget

**Expectation** is that the selection states of widgets always align with the respective selected widget.

below is an example of this strange state the widget gets into. note that the selection rectangle doesn't correspond with the widget that is selected
<img width="1023" alt="invalid-paste-birder" src="https://github.com/awslabs/iot-app-kit/assets/77755322/b27ca871-eb5e-4e14-a802-f7d4f4eb1112">

```[tasklist]
### Acceptance criteria
- [ ] selection states of widgets always align with the respective selected widget, including when pasting widgets at the boundaries of the dashboard
- [ ] test to prevent regression
```

PR #2241, opened 2023-11-08
fix: pasting at the edge of dashboard should paste widget within the …

Priyanshu44:
Issue: https://github.com/awslabs/iot-app-kit/issues/2141

Fix:
-> User pasting widgets at the edge of the dashboard should keep widget within the grid, it won't go off the grid.
-> Added test cases to check the above scenario.

Verification:
[Edge-Pasting.webm](https://github.com/awslabs/iot-app-kit/assets/81667589/689a91b9-847d-498c-9285-1166c787447e)

### DIFF
--- a/packages/dashboard/src/store/actions/pasteWidgets/index.ts
+++ b/packages/dashboard/src/store/actions/pasteWidgets/index.ts
@@ -1,8 +1,7 @@
-import first from 'lodash/first';
-import sortBy from 'lodash/sortBy';
+import { minBy, max } from 'lodash';
 import { v4 } from 'uuid';
 import type { Action } from 'redux';
-import type { Position } from '~/types';
+import type { Position, DashboardWidget } from '~/types';
 import type { DashboardState } from '../../state';
 
 type PasteWidgetsActionPayload = {
@@ -23,9 +22,15 @@ export const pasteWidgets = (state: DashboardState, action: PasteWidgetsAction):
 
   const cellSize = state.grid.cellSize;
   const copyGroup = state.copiedWidgets;
+  const gridWidth = state.grid.width;
+  const gridHeight = state.grid.height;
   let pasteCounter = state.pasteCounter + 1;
 
   let offset: Position = {
+    x: 0,
+    y: 0,
+  };
+  const correctionOffset: Position = {
     x: 0,
     y: 0,
   };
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/store/actions/pasteWidgets/index.ts
```diff
@@ -37,18 +42,43 @@ export const pasteWidgets = (state: DashboardState, action: PasteWidgetsAction):
...
     };
 
-    const widgetToCompare = first(sortBy(copyGroup, ['x', 'y']));
+    const leftmostWidget: DashboardWidget = minBy(copyGroup, 'x') || copyGroup[0];
+    const groupLeftX = leftmostWidget.x;
+
+    const topmostWidget: DashboardWidget = minBy(copyGroup, 'y') || copyGroup[0];
+    const groupTopY = topmostWidget.y;
+
+    const widgetsRightPos = map(copyGroup, (widget) => {
```
mnischay, line 51: you can just do a copyGroup.map ?

packages/dashboard/src/store/actions/pasteWidgets/index.ts
```diff
@@ -23,12 +22,18 @@ export const pasteWidgets = (state: DashboardState, action: PasteWidgetsAction):
 
   const cellSize = state.grid.cellSize;
   const copyGroup = state.copiedWidgets;
+  const xgrid = state.grid.width;
```
mnischay, line 25: i think renaming this to width or gridWidth would make more sense ?

packages/dashboard/src/store/actions/pasteWidgets/index.ts
```diff
@@ -37,18 +42,43 @@ export const pasteWidgets = (state: DashboardState, action: PasteWidgetsAction):
...
+    const groupHeight = groupBottomY - groupTopY;
+
     offset = {
-      x: cellPosition.x - (widgetToCompare?.x ?? 0),
-      y: cellPosition.y - (widgetToCompare?.y ?? 0),
+      x: cellPosition.x - groupLeftX,
+      y: cellPosition.y - groupTopY,
     };
+
+    if (cellPosition.x + groupWidth > xgrid) {
```
mnischay, line 69: I think the logic makes sense, can you add some comments so that when someone is looking at this,  it makes it a bit easier to understand?

packages/dashboard/src/store/actions/pasteWidgets/index.ts
Priyanshu44: Resolved

packages/dashboard/src/store/actions/pasteWidgets/index.ts
Priyanshu44: Resolved

mnischay (approved):

PR #2244, opened 2023-11-08
fix: dashboard mock service worker lifecycle

jmbuss:
## Overview
Fix for mock service worker lifecycle in storybook. Adds a decorator to the Mocked data story which will start and stop the service worker on story mount / unmount. This way, the service worker will not be running for any other story.

_Note: if switching between mocked / connected stories, you will need to refresh the page as react query caches all calls in memory, and there is no access to the client at the Storybook story level._

## Legal

### DIFF
--- a/packages/dashboard/src/msw/useWorker.ts
+++ b/packages/dashboard/src/msw/useWorker.ts
@@ -1,0 +1,12 @@
+import { useEffect } from 'react';
+import { worker } from './browser';
+
+export const useWorker = () => {
+  useEffect(() => {
+    worker.start();
+
+    return () => {
+      worker.stop();
+    };
+  }, []);
+};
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/msw/useWorker.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from 'react';
+import { worker } from './browser';
+
+export const useWorker = () => {
+  useEffect(() => {
+    worker.start();
+
+    return () => {
+      worker.stop();
+    };
```
mnischay, line 10: do you think its worth adding something like location.reload() here to force a memory flush ?

jmbuss: The problem is that the tanstack client is cached across story renders, I don't like the idea of putting a patch for that with our mock service worker code because it ties them together in a weird way.

tracy-french (approved):

PR #2245, opened 2023-11-08
feat: dashboard resource explorer e2e tests

jmbuss:
## Overview
Add integration tests for resource explorer. In order to get mock service worker to work within playwright, a wrapper around the test method is added which utilizes the `playwright-msw` package. This package uses the same handlers as the `msw` implementation but has extra functionality to plug into playwright testing.

Tests Added:
1. basic visibility test
2. asset model happy path test for selecting an asset model and property and adding it to a xy-plot widget.

## Legal

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46916,6 +46916,23 @@
         "node": ">=16"
       }
     },
+    "node_modules/playwright-msw": {
+      "version": "2.2.1",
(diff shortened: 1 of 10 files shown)


tracy-french (approved):
@jmbuss Thank you for adding tests!

Issue #2164, opened 2023-10-25
[Dashboard] Long chart name makes delete button invisible


diehbria:
## Overview
When a chart has a long name, it pushes the "X" delete button to the right making it invisible.

<img width="550" alt="Screenshot 2023-10-25 at 12 12 39 PM" src="https://github.com/awslabs/iot-app-kit/assets/77755322/1507442c-2e03-47be-bfe1-4b1566c2998c">

**Expected behavior**: When a chart name is long enough that it runs into the delete button, it should be truncated with elipses.

```[tasklist]
### Acceptance criteria
- [ ] delete button is always visible, even with long chart names
- [ ] truncated using elipses i.e. "my chart name is long..." 
```

PR #2247, opened 2023-11-09
fix: delete button invisible fix - #2164

YadhunandanGowda:
The issue was the delete icon will be invisible if the title is too long, so I have give a fix for it, now the title is getting wrapped and show in a ellipsis (...) form, please find the attached video below.

[screen-capture.webm](https://github.com/awslabs/iot-app-kit/assets/65999096/b2f26a44-81a9-4af5-ab10-3d3146e68801)


packages/dashboard/src/components/widgets/tile/tile.tsx
```diff
@@ -58,6 +57,10 @@ const WidgetTile: React.FC<WidgetTileProps> = ({ children, widget, title, remove
   const [visible, setVisible] = useState(false);
   const { iotSiteWiseClient } = useClients();
   const { onDelete } = useDeleteWidgets();
+  const titleRef = useRef<HTMLDivElement | null>(null);
+  const isTitleTruncated =
```
jmbuss, line 61: is it not possible to solve this with css alone? Otherwise, we can use - `https://github.com/streamich/react-use/blob/master/docs/useMeasure.md` for these types of things.

jmbuss (changes requested):

packages/dashboard/src/components/widgets/tile/tile.tsx
YadhunandanGowda: Hi,
I tried to achieve it using 'useMeasure' but it didn't happen because we won't get scrollWidth value from it, to show the tooltip conditionally we need the scrollWidth of the title.
So to avoid this extra code, can I just add a title without a condition, which will display regardless of the long or short title?

YadhunandanGowda: Hi @jmbuss,

Since 'useMeasure' didn't help to replace the logic I have removed the condition to avoid extra code and just kept the title, I request you to review it once and merge the PR.

YadhunandanGowda:
due to conflict issue.

Issue #2136, opened 2023-10-20
[Dashboard] need to paste a widget twice to paste widget (hotkey bug)


diehbria:
## Description
Copy and paste via hot keys ctrl-c , ctrl-v does not work the first time, and requires a user to press "ctrl-v" twice to paste.

```[tasklist]
### Acceptance criteria
- [ ] ctrl-v correctly pastes on the first time using the hotkey
```

PR #2248, opened 2023-11-09
fix: copy paste issue on macos

gmuralidharreddy:
## Overview
This PR is for ticket #2136 

I added an event listener for copy-and-paste.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/139960352/58234a9e-1542-4404-9f11-d5e4fb8a41b7

## Legal

### DIFF
--- a/packages/dashboard/src/hooks/useKeyPress.ts
+++ b/packages/dashboard/src/hooks/useKeyPress.ts
@@ -1,10 +1,10 @@
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { isHotkey } from 'is-hotkey';
 
-type KeyPressCallback = (e: KeyboardEvent) => void;
+type KeyPressCallback = (e: KeyboardEvent | ClipboardEvent) => void;
 type KeyPressOptions = {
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/hooks/useKeyPress.ts
```diff
@@ -56,9 +56,13 @@ export const useKeyPress = (key: string, options?: KeyPressOptions | KeyPressCal
   );
 
   useEffect(() => {
+    window.addEventListener('copy', onKeyPress as EventListener);
+    window.addEventListener('paste', onKeyPress as EventListener);
     window.addEventListener('keydown', onKeyPress);
```
mnischay, line 61: do we still need keyup and keydown ?

gmuralidharreddy: yes, we need keydown and keyup listeners since isHotkey supports only **KeyboardEvent** but copy and paste are **ClipboardEvent**, due to this if we remove keydown and keyup then isHotkey returns false.

corteggiano (approved):

Issue #2242, opened 2023-11-08
[ResourceExplorer] UX improvements to the resource explorer


diehbria:
## Description

Update various minor UX improvements to the resource explorer to improve 
- loading states
- error states
- increase page size

### Search query error UX
<img width="592" alt="Screenshot 2023-11-08 at 8 40 58 AM" src="https://github.com/awslabs/iot-app-kit/assets/77755322/cbd9bdcc-f795-42cf-a046-b188047d8ab4">

### Results Error UX
<img width="339" alt="Screenshot 2023-11-08 at 8 17 14 AM" src="https://github.com/awslabs/iot-app-kit/assets/77755322/deabde48-e59a-41d9-9cda-ef096fc0f60c">

### Loading UX
<img width="332" alt="loading-state" src="https://github.com/awslabs/iot-app-kit/assets/77755322/4f376a54-d025-4766-afe8-476b3e78f413">

```[tasklist]
### Acceptance criteria
- [ ] Update validation of search query to be 3 to 50 characters (see "Search query error UX")
- [ ] Update query description text to say 'Enter a query using character string or wildcard character % to search the data.' (see "search query error UX")
- [ ] browse experience datastream table has a error state that matches the provided UX above labeled "Results error UX"
- [ ] search experience datastream table has a error state that matches the provided UX above labeled "Results error UX" 
- [ ] browse experience asset navigation has a error state that matches the provided UX above labeled "Results error UX"
- [ ] search experience has a loading state that matches the provided UX above labeled "loading UX"
- [ ] Update the "default page size" for the data stream results for the browse page to be 250
- [ ] Update the "default page size" for the data stream results for the search page to be 250
```

Chandru-HM:
for the 1st AC: Update validation of search query to be 3 to 50 characters (see "Search query error UX")
as per Tracy's comments in the PR, "The limit should be 1 to 48, as wildcards are appended to the query, counting as 2 of the characters." so updated to 1 to 48.

PR #2252, opened 2023-11-10
fix: pagination default pagesize is set to 250 and error message upda…

Chandru-HM:
## Overview
This PR is for ticket (only a few sub tasks are covered here) #2242  and to update  pagination default page size in Resource explorer tables and ux update
1.  Default page size is now set to 250 for all the tables in resource explorer (assets and data stream tables in modeled and data stream table in unmodeled)
2. Search query validation error message and description is updated.

## Verifying Changes
[re-ux-update.webm](https://github.com/awslabs/iot-app-kit/assets/142866907/28dd813f-0226-4218-bc5b-ed9f2537713f)

## Legal


tracy-french (changes requested):
@Chandru-HM The limit should be 1 to 48, as wildcards are appended to the query, counting as 2 of the characters.

tracy-french (approved):

PR #2265, opened 2023-11-13
fix(video-player): toggle playback mode and update time range

mukeshsahay:
## Overview
- Fixed the toggle playback mode functionality
- Updated the video player to respond to the time range changes

https://github.com/awslabs/iot-app-kit/assets/43108986/cd5590b2-9144-487e-9f9f-0c61f08b1c16

### Video Player
- Run the react-app example from iot-app-kit
- Use the video player in this app to test the functionality

## Legal

### DIFF
--- a/examples/react-app/src/components/VideoPlayer.tsx
+++ b/examples/react-app/src/components/VideoPlayer.tsx
@@ -1,7 +1,9 @@
-import { VideoPlayer } from '@iot-app-kit/react-components';
+import { VideoPlayer, useViewport } from '@iot-app-kit/react-components';
 import { dataSource } from '../dataSource';
-import { kvsStreamName, videoEntityId, videoComponentName, videoViewport } from '../configs';
+import { kvsStreamName, videoEntityId, videoComponentName } from '../configs';
 import './VideoPlayer.scss';
+import { Viewport, isHistoricalViewport, parseDuration } from '@iot-app-kit/core';
+import { useMemo } from 'react';
 
 const videoData = dataSource.videoData({
   kvsStreamName: kvsStreamName,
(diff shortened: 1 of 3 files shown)


examples/react-app/src/components/VideoPlayer.tsx
```diff
@@ -10,9 +11,22 @@ const videoData = dataSource.videoData({
...
+  const { viewport } = useViewport();
+  if (viewport) {
+    if (!isHistoricalViewport(viewport)) {
+      const start = new Date(Date.now() - parseDuration(viewport.duration));
+      const end = new Date();
+      videoViewPort = { start, end };
+    } else {
+      videoViewPort = viewport;
+    }
+  }
```
sheilaXu, line 25: `videoViewPort` should be `useMemo` that depend on `viewport`. the current logic will create a new instance of `{ duration: '0' }` during every rerender which will cause unnecessary rerender to VideoPlayer component

packages/react-components/src/components/video-player/videoPlayer.tsx
```diff
@@ -61,6 +61,8 @@ export const VideoPlayer = (props: IVideoPlayerProps) => {
   const [playbackMode, setPlaybackMode] = useState<typeof PLAYBACKMODE_ON_DEMAND | typeof PLAYBACKMODE_LIVE>(
     initialPlaybackMode
   );
+  const playbackModeRef = useRef({});
+  playbackModeRef.current = playbackMode;
```
sheilaXu, line 65: why do we need this? it seems to be overly complex to have playback in props, states and also ref

sheilaXu: it doesn't seem like there are code needs to be executed when the playbackMode state is changed except `updateVideoSource`, you can probably just delete the state and call updateVideoSource when toggle the mode

examples/react-app/src/components/VideoPlayer.tsx
mukeshsahay: Done - updated the code

packages/react-components/src/components/video-player/videoPlayer.tsx
mukeshsahay: videoPlayerReady() is a callback which is called when video player is initialized. Live button is added to the control bar here with onClick(). Problem was that the event was running in its own closure and had the value at the time when the event was bounded. To fix this, I had to use useRef().

mukeshsahay: I made the changes to remove the state var for playbackMode and calling the updateVideoSource() directly when toggling the playbackMode and it is not working. The instance of VideoPlayer is null in that case. I think it is because of the context in which the call is made. The instance is video player in case of this call is not present.

Currently, it is working because on toggle I change the state and the state change for a component variable updates the component i.e. the current video player.

sheilaXu (approved):

PR #2266, opened 2023-11-14
feat: experimental plugins registry; logger and metrics recorder plugins

chejimmy:
### High level architecture

image (7)

#### Components

High level description of each components:

* AppKit - see [IoT AppKit](https://github.com/awslabs/iot-app-kit)
* [NEW] PluginsRegistry - a basic / stand-in registry to register and provide plugins - an initial plugin system until we have a better design
* [NEW] StorageLogger - a plugin to handle received log and metric emissions and send to Storage; not yet implemented;
* [NEW] Storage - a log and metric storage; not yet implemented;
* Client - extra; please ignore;

## Legal

### DIFF
--- a/packages/core/src/plugins/pluginsRegistry.ts
+++ b/packages/core/src/plugins/pluginsRegistry.ts
@@ -1,0 +1,44 @@
+import type { LoggerSettings } from './loggerSettings';
+import type { MetricsRecorderSettings } from './metricsRecorderSettings';
+
+/**
+ * Internal object to hold all of the plugins.
+ *
+ * @emarks Not exported to encapsulate the registry
+ */
+const pluginsRegistry: {
+  loggerSettings?: LoggerSettings;
+  metricsRecorderSettings?: MetricsRecorderSettings;
+} = {};
+
+/**
+ * Register a Logger for logging AppKit logs.
+ * @param param0 settings for Logger
+ * @alpha
+ */
+export const registerLogger = (settings: LoggerSettings) => {
+  pluginsRegistry.loggerSettings = settings;
+};
+
+/**
+ * Return the registered Logger.
+ * @returns the registered Logger.
+ * @alpha
+ */
+export const getLogger = () => pluginsRegistry.loggerSettings?.provider();
+
+/**
+ * Register a MetricsRecorder for recording AppKit metrics.
+ * @param param0 settings for MetricsRecorder
+ * @alpha
+ */
+export const registerMetricsRecorder = (settings: MetricsRecorderSettings) => {
+  pluginsRegistry.metricsRecorderSettings = settings;
+};
+
+/**
+ * Return the registered MetricsRecorder.
+ * @returns the registered MetricsRecorder.
+ * @alpha
+ */
+export const getMetricsRecorder = () => pluginsRegistry.metricsRecorderSettings?.provider();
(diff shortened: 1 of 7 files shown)


packages/core/src/plugins/pluginsRegistry.ts
```diff
@@ -0,0 +1,44 @@
...
+  loggerSettings?: LoggerSettings;
+  metricsRecorderSettings?: MetricsRecorderSettings;
+} = {};
+
+/**
+ * Register a Logger for logging AppKit logs.
+ * @param param0 settings for Logger
+ * @alpha
+ */
+export const registerLogger = (settings: LoggerSettings) => {
```
jmbuss, line 19: Is there some way we can make a generic registerPlugin method? I am thinking we'll end up having a lot of methods in the long run

jmbuss: 2 ideas, Not sure this will work nicely with typings, 

```
const registerPlugin = (nameSpace: string; plugin: IPlugin) => pluginRegistry[nameSpace] = plugin;
```

```
type Plugin = {
  install: (pluginRegistry: PluginRegistry) => void;
}

const loggerPlugin: Plugin = {
  install: (pluginRegistry) => {
    pluginRegistry.logger = myCustomLogger;
  }
}
```

jmbuss:
Is there a difference between logger and metrics? Should they both rollup to just one logger?

chejimmy:
> Is there a difference between logger and metrics? Should they both rollup to just one logger?

the interfaces are also different, most likely have very different implementations between them; split to comply with single-responsibility principle (SRP);

packages/core/src/plugins/pluginsRegistry.ts
chejimmy: lemme try this 🙏

chejimmy: i'm able to get nice typed function for registration; however, I'm not sure how the retrieval would work 🤔

```
// Registration

export type PluginSettings<NameSpace extends string, Plugin> = { namespace: NameSpace; provider: () => Plugin };
export type PluginType<NameSpace extends string, Plugin> = [NameSpace, Plugin];

export type LoggerSettings = PluginSettings<'logger', Logger>;
export type MetricsRecorderSettings = PluginSettings<'metricsRecorder', MetricsRecorder>;

const pluginsRegistry: {
  loggerSettings?: LoggerSettings;
  metricsRecorderSettings?: MetricsRecorderSettings;
} = {};

export const registerPlugin = (settings: LoggerSettings | MetricsRecorderSettings) => {
  if (settings.namespace === 'logger') {
    pluginsRegistry.loggerSettings = settings;
  }
  if (settings.namespace === 'metricsRecorder') {
    pluginsRegistry.metricsRecorderSettings = settings;
  }
};

```

```
// Retrieval

export const getPlugin = <P>(namespace: 'logger' | 'metricsRecorder'): P | undefined => {
  if (namespace === 'logger') {
    return pluginsRegistry.loggerSettings?.provider();
  }
  if (namespace === 'metricsRecorder') {
    return pluginsRegistry.metricsRecorderSettings?.provider();
  }
};

```

How do we provide something to the function and make sure it is type safe? the `getPlugin` above has errors of
```
Type 'Logger | undefined' is not assignable to type 'P | undefined'.
  Type 'Logger' is not assignable to type 'P'.
    'P' could be instantiated with an arbitrary type which could be unrelated to 'Logger'.
```
and 
```
Type 'MetricsRecorder | undefined' is not assignable to type 'P | undefined'.
  Type 'MetricsRecorder' is not assignable to type 'P'.
    'P' could be instantiated with an arbitrary type which could be unrelated to 'MetricsRecorder'.
```

jmbuss (approved):

Issue #2256, opened 2023-11-10
[Dashboard] fix glitchy configuration panel resizing


diehbria:
## Overview
There are some weird behaviors of the configuration panel, which can occur from stretching / shrinking the window or switching back and forth from preview mode.

The configuration panel is now just a fixed static width. All code around "listening for resizing" should be deleted entirely.

https://github.com/awslabs/iot-app-kit/assets/77755322/e3c3297a-836f-4bf4-b7f2-5ba25424892b

### Steps to reproduce:
- open dashboard in edit mode
- open configuration panel
- shrink window down to 100px
- stretch window back out

```[tasklist]
### Tasks
- [ ] Glitchy behavior is fixed
- [ ] All old code related to watching for "resizes" of the configuration panel are deleted
```

PR #2269, opened 2023-11-15
fix: panel resizing issue fix

Chandru-HM:
## Overview
This PR is for ticket #2256 and to fix resizing issue for configuration and resource explorer panels when the user toggle between edit and preview mode.
1.  first time resource explorer will be open and configuration panel will closed on dashboard open
2. when the user goes to preview mode and comes back to edit mode
- the selected widgets will reset (i.e selected widgets should be an empty array)
- config panel will be closed
- resource explorer will be open  
3. removed the right draghandle  code related to resize listening

## Verifying Changes
[edit-preview.webm](https://github.com/awslabs/iot-app-kit/assets/142866907/bb2dd6e4-3d9b-48d2-a464-62ba4a1960d7)

## Legal

### DIFF
--- a/packages/dashboard/src/components/resizablePanes/index.tsx
+++ b/packages/dashboard/src/components/resizablePanes/index.tsx
@@ -1,19 +1,15 @@
 import React, { useEffect, useState, useRef, useCallback } from 'react';
 import type { MouseEvent, FC, ReactNode } from 'react';
-import { useSelector } from 'react-redux';
 
 import { spaceStaticXs } from '@cloudscape-design/design-tokens';
-import { DashboardState } from '~/store/state';
 import {
   LEFT_WIDTH_PERCENT,
-  RIGHT_WIDTH_PERCENT,
   MINIMUM_CENTER_PANE_WIDTH,
   MINIMUM_SIDE_PANE_WIDTH,
   DEFAULT_SIDE_PANE_WIDTH,
   DEFAULT_COLLAPSED_SIDE_PANE_WIDTH,
   MAXIMUM_PANES_PROPORTION,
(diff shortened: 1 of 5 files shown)


packages/dashboard/src/components/resizablePanes/index.tsx
```diff
@@ -56,9 +54,8 @@ export const ResizablePanes: FC<ResizablePanesProps> = ({ leftPane, centerPane,
   /** Current widths of the three panes, in px */
   const [leftPaneWidth, setLeftPaneWidth] = useState(DEFAULT_SIDE_PANE_WIDTH);
   const [rightPaneWidth, setRightPaneWidth] = useState(DEFAULT_SIDE_PANE_WIDTH);
-  const selectedWidget = useSelector((state: DashboardState) => state.selectedWidgets);
-  const [isRightPaneCollapsed, setRightPaneCollapsed] = useState(selectedWidget?.length === 0);
-  const [isLeftPaneCollapsed, setLeftPaneCollapsed] = useState(selectedWidget?.length === 0);
+  const [isRightPaneCollapsed, setRightPaneCollapsed] = useState(true);
```
jmbuss, line 57: what about using `import useLocalStorage from 'react-use/lib/useLocalStorage';` instead? We can preserve the state this way?

Chandru-HM: as per the new requirement we are not persisting the panel open/closed status. question is raised in the team, based on the product call, i'll update this.

Chandru-HM: updated with default open for resource explorer and close for configuration panel. currently not preserving the panel status, so useLocalStorage not used.

PR #2272, opened 2023-11-15
fix(dashboard): csv will download if viewport has no data

corteggiano:
## Overview
- Download a CSV file when viewport is empty and there are no errors present
- File will still have column names so it not just an empty CSV
- Fix tests to match this new behavior
image

example empty CSV file:
[xy-plot 2023-11-15T14 48 06.034Z.csv](https://github.com/awslabs/iot-app-kit/files/13366219/xy-plot.2023-11-15T14.48.06.034Z.csv)

## Legal

### DIFF
--- a/packages/dashboard/src/components/csvDownloadButton/index.tsx
+++ b/packages/dashboard/src/components/csvDownloadButton/index.tsx
@@ -7,6 +7,24 @@ import { StyledSiteWiseQueryConfig } from '~/customization/widgets/types';
 import { useViewportData } from '~/components/csvDownloadButton/useViewportData';
 import { IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
 
+const EMPTY_DATA = [
+  {
+    timestamp: '',
+    dataQuality: '',
+    value: '',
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/components/csvDownloadButton/index.tsx
```diff
@@ -7,6 +7,24 @@ import { StyledSiteWiseQueryConfig } from '~/customization/widgets/types';
 import { useViewportData } from '~/components/csvDownloadButton/useViewportData';
 import { IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
 
+const EMPTY_DATA = [
```
jmbuss, line 10: What do you think about making a CSV constants file in this folder for these types of things?

jmbuss (approved):

packages/dashboard/src/components/csvDownloadButton/index.tsx
mnischay: +1

mnischay (approved):

PR #2278, opened 2023-11-16
feat(react-components): brush zoom line chart

corteggiano:
## Overview
- enable the brush zoom + zoom reset toolbox
- works in dashboard + in the solo react-component
- does not break existing pan gestures

panning+brush zoom:

view

trend Cursors:
TC

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/events/useHandleChartEvents.ts
+++ b/packages/react-components/src/components/chart/events/useHandleChartEvents.ts
@@ -1,16 +1,18 @@
 import { MutableRefObject, useCallback, useEffect, useState } from 'react';
-import { EChartsOption, EChartsType } from 'echarts';
+import { EChartsOption, EChartsType, ElementEvent } from 'echarts';
 import { KeyMap } from 'react-hotkeys';
 
 export const useHandleChartEvents = (chartRef: MutableRefObject<EChartsType | null>) => {
   const [chartEventsOptions, setChartEventsOptions] = useState<EChartsOption>({});
 
   const [shiftDown, setShiftDown] = useState(false);
-  const [mouseDown, setMouseDown] = useState(false);
-  const [prevIsGrabbing, setPrevIsGrabbing] = useState(false);
+  const [prevIsPanning, setPrevIsPanning] = useState(false);
+  const [isPanning, setIsPanning] = useState(false);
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/chart/events/useHandleChartEvents.ts
```diff
@@ -75,61 +78,68 @@ export const useHandleChartEvents = (chartRef: MutableRefObject<EChartsType | nu
...
         if (stateChanged) {
-          setOnDragEnd(chart);
+          setOnPanEnd(chart);
         }
       }
     };
 
     chart?.getZr().on('mousemove', mouseMoveEventHandler);
 
+    const zoomClickHandler = (e: any) => {
```
mnischay, line 131: you cant use any :)

corteggiano: yeah i know 😢 looking into what event this would be bc echarts defines it as:
`WithThisType<EvtDef[EvtNm], CbThis<Ctx, this>>`

jmbuss (commented):
Maybe we don't need to do this now, but writing down some ideas!

Is there any way we can separate out gestures taken on echarts as individual hooks? I think in general we have click and drag/drop triggers on the echarts chart. Maybe we can do something like dashboard where we have generic hooks for those and then we can have something abstract on top of it. For instance pan / brush are just a drag/drop gestures. highlight / toggle context menu is a click gesture.

packages/react-components/src/components/chart/events/useHandleChartEvents.ts
```diff
@@ -75,61 +78,70 @@ export const useHandleChartEvents = (chartRef: MutableRefObject<EChartsType | nu
...
         if (stateChanged) {
-          setOnDragEnd(chart);
+          setOnPanEnd(chart);
         }
       }
     };
 
     chart?.getZr().on('mousemove', mouseMoveEventHandler);
 
+    const zoomClickHandler = (e: unknown) => {
```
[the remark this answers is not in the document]
corteggiano: nope! i posted it expected type in nischay's comment above this ^ 
the type is `WithThisType<EvtDef[EvtNm], CbThis<Ctx, this>>`

mnischay (approved):

PR #2285, opened 2023-11-20
fix: bugfix for overlapping colors in color palette

jmbuss:
## Overview
Bugfix for selecting colors to apply to configured queries in the dashboard definition and to time series data. This change adds a Colorizer object which handles providing colors based on a rotation of colors in a palette. This is used in the dashboard to provide default colors to each query added by the resource explorer. It also replaces the selection mechanism in useTimeSeries data so that we choose the same colors for datastreams between renders.

Also adds in a test function for our core-utils package so that they are run on CI

## Legal


packages/core-util/src/sdks/colorizer.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { colorPalette } from './colorPalette';
```
corteggiano, line 1: awesome tests 🤩

corteggiano (approved):

corteggiano (dismissed):

dpportet (approved):

corteggiano (approved):

PR #2286, opened 2023-11-21
fix: chart gesture performance

tracy-french:
## Overview
The purpose of this change is to restore chart performance by preventing manual data zoom dispatching when the data zoom event is triggered through a gesture. The code was already attempting to provide this guard, yet the guard was ineffective, as it relied on toggling state within the same handler. React optimizes state changes through batching, so it was not possible to control the toggle effectively within a single handler.

Other optimizations include:
- Using a ref instead of state to prevent re-renders during events.
- Using requestAnimationFrame to throttle the data zoom event and synchronize the animation with the browser refresh rate.

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/hooks/useDataZoom.ts
+++ b/packages/react-components/src/components/chart/hooks/useDataZoom.ts
@@ -1,11 +1,10 @@
-import { MutableRefObject, useEffect, useRef, useState } from 'react';
+import { MutableRefObject, useEffect, useRef } from 'react';
 import { useViewport } from '../../../hooks/useViewport';
 import { DataZoomComponentOption, EChartsType } from 'echarts';
 import { ECHARTS_GESTURE } from '../../../common/constants';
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -1,11 +1,10 @@
-import { MutableRefObject, useEffect, useRef, useState } from 'react';
+import { MutableRefObject, useEffect, useRef } from 'react';
 import { useViewport } from '../../../hooks/useViewport';
 import { DataZoomComponentOption, EChartsType } from 'echarts';
 import { ECHARTS_GESTURE } from '../../../common/constants';
-import { DEFAULT_DATA_ZOOM, ECHARTS_ZOOM_DEBOUNCE_MS } from '../eChartsConstants';
```
corteggiano, line 5: we can delete `ECHARTS_ZOOM_DEBOUNCE_MS` if its not used anywhere else

tracy-french: Good call! I didn't check lol

PR #2296, opened 2023-11-23
feat: added the error state for resource explorer tables

Chandru-HM:
## Overview
This PR is for ticket (a few sub tasks are covered here(3, 4 & 5)) #2242 and to update error state in Resource explorer tables
1. added error state for browse assets table
2. added error state for browse data streams table
3. added error state for search data streams table

## Verifying Changes
Assets table error state:
assets-error
Browse data stream table:
datastream-error
Search data stream table:
search-error-state

## Legal

### DIFF
--- a/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/useSearch.ts
+++ b/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/useSearch.ts
@@ -21,7 +21,7 @@ interface UseSearchProps {
 
 /** Use to send a search request. */
 export function useSearch({ workspaceId, searchQuery, client }: UseSearchProps) {
-  const { data, hasNextPage, isFetching, fetchNextPage } = useInfiniteQuery({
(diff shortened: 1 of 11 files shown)


Chandru-HM:
> The header above the Alert is missing in this CR ? --> "Search Result for "Query"" or is it not part of AC in the sub tasks ?

updated.

packages/dashboard/src/components/queryEditor/helpers/resourceExplorerErrorState.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
mnischay, line 1: the folder name seems to be not super appropriate... these is a component... helpers seems like utility functions.... probably just components might make sense ?

mnischay (changes requested):
code change LGTM, just a small naming change request

tracy-french (approved):
Looks like the change has been made. Based on Nischay's guidance that the code looks good and that the requested change was made, I will approve and merge to move this allow.

Issue #2292, opened 2023-11-22
[ResourceExplorer] Resource explorer becomes tiny when switching from edit -> view -> edit


corteggiano:
# Description:

This task is to make it so that the resource explorer does not become very thin when switching between view and edit modes.

example of bug: 
![bad](https://github.com/awslabs/iot-app-kit/assets/28601414/400fcc19-2b96-43d3-a788-731738f4a29e)

```[tasklist]
### Acceptance criteria
- [ ] width should be the same when first opening a dashboard
- [ ] width should be the same when closing/expanding the resource explorer
- [ ] width should be the same when switching from edit -> view -> edit
- [ ] video of bug not occurring 
```

PR #2297, opened 2023-11-23
fix: updated the resource explorer default width

Chandru-HM:
## Overview
This PR is for ticket #2292 and to fix resource explorer width in different scenarios
1. default width (500px) on first opening the dashboard
2. default width on expanding resource explorer
3.  default width on changing edit and view mode if width is equal/less than default width
4. persists width between edit and preview view if width is greater than default 

## Verifying Changes
[resource-explorer-width.webm](https://github.com/awslabs/iot-app-kit/assets/142866907/78b80464-28fe-46d4-afd1-d01cfa120292)

## Legal

### DIFF
--- a/packages/dashboard/src/components/resizablePanes/index.tsx
+++ b/packages/dashboard/src/components/resizablePanes/index.tsx
@@ -5,7 +5,6 @@ import { spaceStaticXs } from '@cloudscape-design/design-tokens';
 import {
   LEFT_WIDTH_PERCENT,
   MINIMUM_CENTER_PANE_WIDTH,
-  MINIMUM_SIDE_PANE_WIDTH,
   DEFAULT_SIDE_PANE_WIDTH,
   DEFAULT_COLLAPSED_SIDE_PANE_WIDTH,
   MAXIMUM_PANES_PROPORTION,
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/components/resizablePanes/index.tsx
```diff
@@ -67,11 +66,12 @@ export const ResizablePanes: FC<ResizablePanesProps> = ({ leftPane, centerPane,
 
     if (storedLeftWidthPercent) {
       const storedLeftWidth = elementWidth * storedLeftWidthPercent;
-      setLeftPaneWidth(storedLeftWidth);
+      setLeftPaneWidth(storedLeftWidth > DEFAULT_SIDE_PANE_WIDTH ? storedLeftWidth : DEFAULT_SIDE_PANE_WIDTH);
```
mnischay, line 69: DEFAULT_SIDE_PANE_WIDTH this should be DEFAULT_RESOURCE_EXPLORER_WIDTH ?

Chandru-HM: but we are using same width as default for right (configuration panel) as well.

mnischay (approved):

PR #2302, opened 2023-11-27
fix: accessibility fixes

Priyanshu44:
Accessibility fixes :
Ensures all elements with a role attribute use a valid value : Abstract role cannot be directly used: widget:

Ensures buttons have discernible text: aria-label attribute does not exist or is empty

Ensures every form element has a label: aria-label attribute does not exist or is empty (color-picker)


tracy-french (approved):
Thank you, @Priyanshu44!

PR #2303, opened 2023-11-27
fix(react-components): trendcursor hotkey indicates addition state

ssjagad:
## Overview
The trend cursor hotkey (T key) did not specify that the user was in a state to add a TC. To fix this, we change the cursor to a cross hair cursor on keydown to a crosshair instead of on mouse move.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/145582655/68daa931-0b51-420b-897f-e56eca9fd813

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/trendCursor/mouseEvents/useTrendCursorsEvents.ts
+++ b/packages/react-components/src/components/chart/trendCursor/mouseEvents/useTrendCursorsEvents.ts
@@ -126,8 +126,13 @@ const useTrendCursorsEvents = ({
   // standalone mode --> updates the local state
   useEffect(() => {
     const currentChart = chartRef.current;
+    if (isInCursorAddModeRef.current && graphicRef.current.length < MAX_TREND_CURSORS) {
+      chartRef.current?.getZr().setCursorStyle('crosshair');
+    } else {
+      chartRef.current?.getZr().setCursorStyle('default'); // in case we get out of add mode and mouse does not move, switch back to default cursor
+    }
 
-    const mouseMoveEventHandler = () => mouseoverHandler(isInCursorAddModeRef.current, chartRef);
+    const mouseMoveEventHandler = () => mouseoverHandler(isInCursorAddModeRef.current, graphicRef.current, chartRef);
 
     // this handles all the clicks on the chart
     // this click would be either
@@ -246,7 +251,15 @@ const useTrendCursorsEvents = ({
       currentChart?.getZr()?.off('dragstart', dragStartEventHandler);
       currentChart?.off('datazoom', zoomHandler);
     };
-  }, [addNewTrendCursor, deleteTrendCursor, groupId, updateTrendCursorsInSyncState, chartRef, setGraphic]);
+  }, [
+    addNewTrendCursor,
+    deleteTrendCursor,
+    groupId,
+    updateTrendCursorsInSyncState,
+    chartRef,
+    setGraphic,
+    isInCursorAddMode,
+  ]);
 
   const copyTrendCursorData = (posX: number) => {
     const timestampOfClick = calculateTimeStamp(posX, chartRef);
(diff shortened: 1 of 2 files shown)


mnischay (changes requested):
I dont think the below changes will fix the issue, right now , mouseMoveEventHandler is when the cursor style is set, instead it should on the `isInCursorAddMode === true`

ssjagad:
> I dont think the below changes will fix the issue, right now , mouseMoveEventHandler is when the cursor style is set, instead it should on the `isInCursorAddMode === true`

Good point. I had it change to cross hair on mouse move. Added that check and it should be good :)

packages/react-components/src/components/chart/trendCursor/mouseEvents/useTrendCursorsEvents.ts
```diff
@@ -66,6 +78,7 @@ const useTrendCursorsEvents = ({
             tcId: `trendCursor-${uuid()}`,
             timestamp: timestampInMs,
           });
+          setIsInCursorAddModeRef.current(false); // change cursor back to regular from cross hair (hotkeys)
```
mnischay, line 81: this should be added after the else loop...

packages/react-components/src/components/chart/trendCursor/useTrendCursors.ts
```diff
@@ -83,12 +84,13 @@ const useTrendCursors = ({
 
   const trendCursorKeyMap: KeyMap = {
     commandDown: { sequence: 't', action: 'keydown' },
-    commandUp: { sequence: 't', action: 'keyup' },
   };
 
   const trendCursorHandlers = {
-    commandDown: () => setIsInCursorAddMode(true),
-    commandUp: () => setIsInCursorAddMode(false),
+    commandDown: () => {
```
mnischay, line 90: I am not sure why commandUp was removed.... if the user continues to hold T, we should keep them in the add mode as long as that ?

packages/react-components/src/components/chart/trendCursor/mouseEvents/useTrendCursorsEvents.ts
mnischay: Also, why forcefully stop the add mode ? like user is still holding the t ... and we are kicking them out of the add mode ?

mnischay (changes requested):
I think you are missing the RC of the issue...
We are showing the cross hair on the mousemove event... but when the user
1. presses the t button AND
2. does not move the mouse
even though we are in "add" mode, the mouse does not reflect it.. 

So the solution would to make the "change in cursor style"  not dependent on the mouse but instead make it dependent on 
(ignoreHotKey || isInCursorAddModeRef.current) && graphicRef.current.length < MAX_TREND_CURSORS

i.e when we allow the user to add a TC

packages/react-components/src/components/chart/trendCursor/mouseEvents/useTrendCursorsEvents.ts
```diff
@@ -113,8 +113,13 @@ const useTrendCursorsEvents = ({
...
   useEffect(() => {
     const currentChart = chartRef.current;
+    if (isInCursorAddModeRef.current && graphicRef.current.length < MAX_TREND_CURSORS) {
+      chartRef.current?.getZr().setCursorStyle('crosshair');
+    } else {
+      chartRef.current?.getZr().setCursorStyle('default'); // in case we get out of add mode and mouse does not move, switch back to default cursor
+    }
 
-    const mouseMoveEventHandler = () => mouseoverHandler(isInCursorAddModeRef.current, chartRef);
+    const mouseMoveEventHandler = () => mouseoverHandler(isInCursorAddModeRef.current, graphicRef.current, chartRef);
```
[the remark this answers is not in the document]
ssjagad: updated video to show this

mnischay (approved):

PR #2322, opened 2023-12-04
fix(dashboard): bold, italics, underline interactable by keyboard

ssjagad:
## Overview
Changed div element into button element to make it focusable and interactive by keyboard users. 

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/145582655/3b71a8f8-00c0-4b83-968c-a00fce1a470e

## Legal

### DIFF
--- a/packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
+++ b/packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
@@ -22,7 +22,7 @@ const ButtonWithState: FC<ButtonWithStateProps> = ({
   ...others // passing other attributes like data-test-id
 }) => {
   return (
-    <div
+    <button
       role='checkbox'
       aria-checked={checked}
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
```diff
@@ -22,15 +22,15 @@ const ButtonWithState: FC<ButtonWithStateProps> = ({
   ...others // passing other attributes like data-test-id
 }) => {
   return (
-    <div
+    <button
```
corteggiano, line 25: aria-label so a screen reader knows what the button is for?

ssjagad: I think those are added when the component is used. ie line 94 in text.tsx

corteggiano (approved):

PR #2372, opened 2023-12-11
feat(dashboard): add metrics for query editor usage

jmbuss:
## Overview
Implement metrics recording for query editor flows. Tracking when users add a data stream via modeled / unmodeled / asset modeled, and tracking which browsing feature they use for modeled data.

## Legal


chejimmy (approved):
LGTM 🙏

Issue #1947, opened 2023-07-31
Dashboard: header design update


nadeswal:
This task is to update the dashboard header to contain the dashboard name and other UX improvements details below.

## UX details
<img width="797" alt="header-ux" src="https://github.com/awslabs/iot-app-kit/assets/77755322/a4e30237-9fbb-4599-b005-2a3268061153">

-  Create another dashboard component
-  On the left, add the name of the dashboard, use title header- page component.
- Move time machine and action stripe to the right as shown in the picture above.
- Use grey 900 as bg-color
-  Use h= 80px for the component box
- do not add the "info" link that is shown on the mock.

```[tasklist]
### Acceptance criteria
- [ ] Adheres to provided UX details
- [ ] Is accessible
- [ ] has a snapshot test
```

diehbria:
will reopen once UX is provided

nadeswal:
<img width="1296" alt="Screenshot 2023-10-20 at 2 43 50 PM" src="https://github.com/awslabs/iot-app-kit/assets/141063047/eeb57685-b870-44b4-8189-37c7cc09b3a4">

Updated the header section!

gmuralidharreddy:
> <img alt="Screenshot 2023-10-20 at 2 43 50 PM" width="1296" src="https://user-images.githubusercontent.com/141063047/277025969-eeb57685-b870-44b4-8189-37c7cc09b3a4.png">
> Updated the header section!

@nadeswal @diehbria Can I remove titles for **Time machine** and **Actions**? or keep it as it is?

![image](https://github.com/awslabs/iot-app-kit/assets/139960352/808f3db7-af09-4aa1-ada9-68b132e70b64)

nadeswal:
Yes, please remove it.

nadeswal:
<img width="815" alt="Screenshot 2023-11-29 at 4 32 29 PM" src="https://github.com/awslabs/iot-app-kit/assets/141063047/56492c3e-d40a-4b34-b6c8-c8a52be57b08">

Updated mock

PR #2381, opened 2023-12-13
feat: dashboard new header design #1947

gmuralidharreddy:
## Overview
This change is for the ticket #1947 

- Added new header design at the top.
- Displayed dashboard title on the left side and time series and actions buttons on the right side.
- Also displayed the asset model selection dropdown on the white background toolbar in preview mode.
- Fixed scroll issue and bar chart alignment issue due to new header height.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/139960352/fc438b1d-fc4f-4f73-a49a-cca82414df65

## Legal

### DIFF
--- a/packages/dashboard/src/components/assetModelSelection/assetModelSelect.tsx
+++ b/packages/dashboard/src/components/assetModelSelection/assetModelSelect.tsx
@@ -11,8 +11,16 @@ type AssetModelSelectOptions = {
   client: IoTSiteWiseClient;
   assetModelId: string;
   selectedAssetId?: string;
+  hideTitle?: boolean;
+  controlId?: string;
 };
-export const AssetModelSelect = ({ client, assetModelId, selectedAssetId }: AssetModelSelectOptions) => {
+export const AssetModelSelect = ({
+  client,
+  assetModelId,
+  selectedAssetId,
+  hideTitle = false,
+  controlId,
+}: AssetModelSelectOptions) => {
   const { updateSelectedAsset } = useModelBasedQuery();
 
   const { assetSummaries } = useAssetsForAssetModel({ assetModelId, client, fetchAll: true });
@@ -20,12 +28,13 @@ export const AssetModelSelect = ({ client, assetModelId, selectedAssetId }: Asse
   const selectedAsset = assetSummaries.find(({ id }) => id === selectedAssetId);
 
   return (
-    <FormField label='Asset'>
+    <FormField label={`${!hideTitle ? 'Asset' : ''}`}>
       <AssetForAssetModelSelect
         assetModelId={assetModelId}
         selectedAsset={selectedAsset}
         onSelectAsset={updateSelectedAsset}
         client={client}
+        controlId={controlId}
       />
     </FormField>
   );
--- a/packages/dashboard/src/components/assetModelSelection/assetModelSelection.tsx
+++ b/packages/dashboard/src/components/assetModelSelection/assetModelSelection.tsx
@@ -1,8 +1,17 @@
 import React from 'react';
 
 import { IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
+import {
+  colorBorderDividerDefault,
+  colorTextBodyDefault,
+  spaceScaledS,
+  spaceScaledXxxs,
+  spaceStaticXxl,
+} from '@cloudscape-design/design-tokens';
+
 import { AssetModelSelect } from './assetModelSelect';
 import { useModelBasedQuery } from '../queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/modelBasedQuery/useModelBasedQuery';
+import './assetModelSelection.css';
 
 type AssetModelSelectionOptions = {
   client: IoTSiteWiseClient;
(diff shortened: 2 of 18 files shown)


corteggiano:
please fix the merge conflicts so we can get this in!

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -230,86 +248,110 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({ onSave, edit
...
-              dashboardConfiguration={dashboardConfiguration}
-              grid={grid}
-              significantDigits={significantDigits}
-              editable={editable}
-            />
-          </SpaceBetween>
+    <ContentLayout
+      disableOverlap
+      header={
+        <Box padding={{ horizontal: 's', top: 'm' }}>
```
jmbuss, line 254: let's make a component for dashboard header.

jmbuss:
It looks like in your recording that the asset selection was showing for bar chart even though you only added modeled data. I don't think we want to show the asset selection unless the dashboard contains an assetModel query. There is a hook for that check, you can reference the current implementation of that component to see how it worked

gmuralidharreddy:
> It looks like in your recording that the asset selection was showing for bar chart even though you only added modeled data. I don't think we want to show the asset selection unless the dashboard contains an assetModel query. There is a hook for that check, you can reference the current implementation of that component to see how it worked

Yeah, in the recorder video, I intentionally showed the asset selection to visualize how it looks like in the white toolbar by removing the condition to display asset selection. Now I re-recorded the video and showed the asset selection based on assetModel query and hiding the asset selection when you add only modeled data.

packages/dashboard/src/components/internalDashboard/index.tsx
gmuralidharreddy: Done

packages/dashboard/src/components/assetModelSelection/assetModelSelect.tsx
```diff
@@ -11,16 +11,22 @@ type AssetModelSelectOptions = {
...
+}: AssetModelSelectOptions) => {
   const { updateSelectedAsset } = useModelBasedQuery();
 
   const { assetSummaries } = useAssetsForAssetModel({ assetModelId, client, fetchAll: true });
 
   const selectedAsset = assetSummaries.find(({ id }) => id === selectedAssetId);
 
   return (
-    <FormField label='Asset'>
+    <FormField label={`${!hideTitle ? 'Asset' : ''}`}>
```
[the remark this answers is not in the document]
gmuralidharreddy: Didn’t get you exactly, but I added hideTitle because in the new design Asset label(title) is not required at the top of the select dropdown. So, I am not disturbing this component by completely removing the title just added a check to this label.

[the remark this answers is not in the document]
gmuralidharreddy: Got it, Now added the label tag with htmlFor and passed controlId to the Select component for accessibility.

diehbria (changes requested):
This will require a change to move the `name` property from being appended to the Dashboard Configuration, to instead moving it as a top level property on the `<Dashboard />` component.

Rather than passing an empty string to denote no name, make this property optional. When it is optional, no dashboard name should be passed in.

In general, want to avoid passing around empty strings to denote undefined

This change will require a corresponding change on iot-app code base to ensure dashboard name is passed in, but that can be done as a phase two of this issue since we will make the `name` property optional on `dashboard`

gmuralidharreddy:
> This will require a change to move the `name` property from being appended to the Dashboard Configuration, to instead moving it as a top level property on the `<Dashboard />` component.
> 
> Rather than passing an empty string to denote no name, make this property optional. When it is optional, no dashboard name should be passed in.
> 
> In general, want to avoid passing around empty strings to denote undefined
> 
> This change will require a corresponding change on iot-app code base to ensure dashboard name is passed in, but that can be done as a phase two of this issue since we will make the `name` property optional on `dashboard`

@diehbria  Moved dashboardName out of dashboard configuration and made it optional.

diehbria (changes requested):
change `dashboardName` to `name` (part of the public API, so important that it matches exactly what we want it to be)

gmuralidharreddy:
> change `dashboardName` to `name` (part of the public API, so important that it matches exactly what we want it to be)

renamed from dashboardName to name

packages/dashboard/src/components/assetModelSelection/assetModelSelection.tsx
```diff
@@ -11,8 +20,33 @@ type AssetModelSelectionOptions = {
...
   if (!hasModelBasedQuery || !assetModelId) return null;
 
-  return <AssetModelSelect assetModelId={assetModelId} selectedAssetId={selectedAssetId} client={client} />;
+  return (
+    <div className='asset-model-selection'>
+      <h4 style={{ color: colorTextBodyDefault }}>
+        <label htmlFor={assetModelSelectionControlId}>Assets</label>
+      </h4>
+      <Divider />
+      <div style={{ width: '300px' }}>
```
jmbuss, line 41: is it always the case that this should only be 300px wide? Is there any other way we can make this more responsive?

jmbuss (commented):
Leaving one non blocking comment, but I think we should try to fix

jmbuss (approved):

diehbria (approved):

PR #2391, opened 2023-12-14
perf: prevent duplication of simultaneous data stream metadata requests

tracy-french:
## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/source-iotsitewise/src/asset-modules/describeModeledDataStreamRequest/describeAssetRequest.ts
```diff
@@ -1,30 +1,53 @@
...
+
+      const command = this.#createCommand(assetId);
+      const request = this.#client.send(command);
+
+      this.#cachePendingRequest(assetId, request);
+
+      const asset = await request;
 
       this.#cacheAsset(assetId, asset);
+      this.#deletePendingRequest(assetId);
```
chejimmy, line 50: should the deletion happens always?

tracy-french: I believe it is safe to do so. The asset has been cached at this point. Any simultaneous requests have a reference directly to the pending promise. If the pending promise is deleted from the cache, it won't affect the any simultaneous request.

Issue #2361, opened 2023-12-08
[Accessibility][ConfigPanelText] Controls do not have an accessible name


mnischay:
#### Notes and Advice 
Interactive control "Color" do not have an accessible name. The purpose of these controls will not be clear to users of assistive technologies.
 
 #### Recommendation 
The accessible name is the text that identifies and describes the control to assistive technology users. It is programmatically exposed by the browser to assistive technology products (such as screen readers or speech recognition tools).
 
Provide an [accessible name](https://w3c.github.io/html-aam/#accessible-name-and-description-computation) for the control using native HTML elements/attributes or ARIA attributes.
 
For native labelable controls use a programmatically associated <label> element that specifies the accessible name:
 
<label _for="test"_>Color</label>
<input type="color" value="var(--color-text-body-default-mmvcvg, #000716)" __id="test"_>
 Resources 
[What is an accessible name?](https://www.tpgi.com/what-is-an-accessible-name/)
[The label element](https://thepaciellogroup.github.io/AT-browser-tests/test-files/label.html)
[The difference between ](https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/)[aria-label](https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/)[ and ](https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/)[aria-labelledby](https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/)

### AC
The above violations is fixed

PR #2410, opened 2023-12-20
fix: programatically linked the color picker control with color label in config panel text

Kausar-HM:
This PR addresses the accessibility issue [#2361](https://github.com/awslabs/iot-app-kit/issues/2361)

### Verifying changes:
There was no difference in screen reader behaviour with this change, the screen reader was already announcing the color label and color control without this change because aria-label is already present.

**DOM before making this change:**
image

**DOM after making this change:**
image

### DIFF
--- a/packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
+++ b/packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
@@ -85,8 +85,8 @@ const TextSettings: FC<TextSettingsProps> = ({
   return (
     <ExpandableSection headerText={defaultMessages.title} defaultExpanded>
       <div className='text-configuration' style={{ gap: awsui.spaceScaledS }}>
-        <label>{defaultMessages.color}</label>
-        <ColorPicker color={fontColor} updateColor={updateFontColor} />
+        <label htmlFor='text-color-picker'>{defaultMessages.color}</label>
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
```diff
@@ -85,8 +85,8 @@ const TextSettings: FC<TextSettingsProps> = ({
   return (
     <ExpandableSection headerText={defaultMessages.title} defaultExpanded>
       <div className='text-configuration' style={{ gap: awsui.spaceScaledS }}>
-        <label>{defaultMessages.color}</label>
-        <ColorPicker color={fontColor} updateColor={updateFontColor} />
+        <label htmlFor='color-picker'>{defaultMessages.color}</label>
+        <ColorPicker id='color-picker' color={fontColor} updateColor={updateFontColor} />
```
jmbuss, line 89: Can we make this more specific for text settings color picker? Also, do our other color pickers in the properties / alarms section need this also?

Kausar-HM: I have changed the id to be specific to 'text' now. Also this change was requested for configPanelText colorPicker label only.

mnischay (approved):

PR #2412, opened 2023-12-20
fix(react-components): remove data point symbols after a threshold

jmbuss:
## Overview
Turn off series symbols once there are over 4k datapoints on the chart. This will increase performance when panning the chart along the x axis.

## Legal


mnischay (commented):
is there anyway we can quantify this ???

corteggiano (approved):

PR #2414, opened 2023-12-20
fix: panning performance

jmbuss:
## Overview
Consolidate viewport tick hook and datazoom hook into 1 spot.

## Legal


packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -22,8 +22,8 @@ import { MultiYAxisLegend } from './multiYAxis/multiYAxis';
 
 import './chart.css';
 import { useContextMenu } from './contextMenu/useContextMenu';
-import { useViewportToMS } from './hooks/useViewportToMS';
-import { DEFAULT_CHART_VISUALIZATION, DEFAULT_TOOLBOX_CONFIG } from './eChartsConstants';
+// import { useViewportToMS } from './hooks/useViewportToMS';
```
corteggiano, line 25: non blocker but should delete

corteggiano (approved):

PR #2423, opened 2023-12-27
fix(dashboard): padding for tabs in config panel + remove scroll in threshold panel

ssjagad:
## Overview
Added 8px padding on the left to every item in each tab in the config panel. Also removed the horizontal scroll on the threshold panel. 

## Verifying Changes
Screenshot 2023-12-26 at 4 13 35 PM

Screenshot 2023-12-26 at 4 13 59 PM

## Legal

### DIFF
--- a/packages/dashboard/src/customization/propertiesSections/lineAndScatterStyleSettings/section.tsx
+++ b/packages/dashboard/src/customization/propertiesSections/lineAndScatterStyleSettings/section.tsx
@@ -1,6 +1,4 @@
 import React from 'react';
-
-import SpaceBetween from '@cloudscape-design/components/space-between';
 import { PropertiesSection } from '~/customization/propertiesSectionComponent';
 import { ChartLegend, LineScatterChartWidget, LineStyles, SymbolStyles } from '~/customization/widgets/types';
 import { DashboardWidget } from '~/types';
@@ -220,7 +218,7 @@ const RenderLineAndScatterStyleSettingsSection = ({
   };
 
   return (
-    <SpaceBetween size='s' direction='vertical'>
+    <>
       <AggregationAndResolutionSection
         aggregation={aggregationType}
         resolution={resolution}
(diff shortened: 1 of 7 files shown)


packages/dashboard/src/customization/propertiesSections/lineAndScatterStyleSettings/section.tsx
```diff
@@ -220,7 +219,7 @@ const RenderLineAndScatterStyleSettingsSection = ({
   };
 
   return (
-    <SpaceBetween size='s' direction='vertical'>
+    <>
```
mnischay, line 222: you can completely remove this ?

packages/dashboard/src/customization/propertiesSections/propertiesSectionsStyle.css
```diff
@@ -0,0 +1,21 @@
+.section-padding {
+  padding-left: 8px;
```
mnischay, line 2: worth using cloudscape spacing tokens

mnischay (commented):
Just small nits

packages/dashboard/src/customization/propertiesSections/lineAndScatterStyleSettings/section.tsx
ssjagad: Need to wrap everything under one div/fragment. Removed the SpaceBetween because navita's design shows that we dont need to have a gap between each drop down.

tracy-french:
@ssjagad You mention removing scroll on the thresholds panel. Have you tested it with enough thresholds to overflow the box? If so, does it work fine?

ssjagad:
> @ssjagad You mention removing the horizontal scroll on the thresholds panel. Have you tested the use of the contains operator to see if it overflows with the larger text and other large element contents? Or are all the elements static in width and it doesn't matter the contents of the elements?

The drop down seems to be static in width. Manually tested with different window sizes and zooms and it works

tracy-french (approved):
Ship it! Thank you, @ssjagad!

Issue #2384, opened 2023-12-13
[ResourceExplorer] Automatically select first workspace


diehbria:
When trying to search using the resource explorer, automatically select the first available workspace to reduce friction in performing a search

```[tasklist]
### Acceptance criteria
- [x] twinmaker workspace automatically selects first available workspace
- [x] if a different workspace is selected, store in local storage this selection and automatically select whatever the last selection was if it's available
```

PR #2424, opened 2023-12-27
feat: auto select previously selected workspace or first workspace from options

Kausar-HM:
This PR adds below functionality to the workspace dropdown in resource explorer #2384 
1. Automatically selects first available workspace from the options list, if nothing was selected previously
2. If some workspace option was selected previously, it is stored in localStorage, and on next load, the same option is set as default(i.e. automatically selected)
3. If Previously selected option which is stored in localStorage is no longer present in the workspace options list, it will not be selected automatically, instead first option from the list will be selected.

### Verifying changes:
First time load:
[Screencast from 05-01-24 05:31:06 PM IST.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/4ec97422-7430-4325-88e1-939130f3af0b)

Next renders:
[Screencast from 05-01-24 05:32:01 PM IST.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/19da88e3-30b2-455c-92e7-ec241d646817)

In video - 
1. By default first option from list is selected automatially
2. User selects second option, which is then stored to localStorage
3. The value in localStorage is not present in current options list, so by defautl first option from list is selected

### DIFF
--- a/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
+++ b/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
@@ -1,9 +1,11 @@
 import { type IoTTwinMakerClient } from '@aws-sdk/client-iottwinmaker';
 import FormField from '@cloudscape-design/components/form-field';
 import Select, { type SelectProps } from '@cloudscape-design/components/select';
+import { OptionDefinition } from '@cloudscape-design/components/internal/components/option/interfaces';
 import { type useQuery } from '@tanstack/react-query';
 import React, { useEffect } from 'react';
 import { Controller, type Control } from 'react-hook-form';
+import { useLocalStorage } from 'react-use';
 
 import { useWorkspaces } from './useWorkspaces';
 import { WorkspaceOptionFactory } from './workspaceOptionFactory';
@@ -26,10 +28,33 @@ export const WorkspaceSelector = ({
   const { workspaces, status } = useWorkspaces({ client });
   const workspaceOptions = createWorkspaceOptions(workspaces);
 
+  const [storedWorkspace, setStoredWorkspace] =
+    useLocalStorage<OptionDefinition | null>('storedWorkspace', null);
+
   useEffect(() => {
     OnGettingError(status === 'error');
   }, [status, OnGettingError]);
 
+  useEffect(() => {
+    if (workspaceOptions.length) {
+      const isStoredValueInOptions =
+        storedWorkspace &&
+        workspaceOptions.some(
+          (option: OptionDefinition) => option.value === storedWorkspace.value
+        );
+      if (!isStoredValueInOptions) {
+        setStoredWorkspace(workspaceOptions[0]);
+      }
+    }
+  }, [storedWorkspace, setStoredWorkspace, workspaceOptions]);
+
+  const onChangeHandler = (
+    detail: SelectProps.ChangeDetail,
+    onChange: (...event: unknown[]) => void
+  ) => {
+    setStoredWorkspace(detail.selectedOption);
+    onChange(detail.selectedOption);
+  };
   return (
     <Controller
       control={control}
@@ -44,8 +69,8 @@ export const WorkspaceSelector = ({
           <Select
             disabled={status == 'error'}
             options={workspaceOptions}
-            selectedOption={field.value}
-            onChange={({ detail }) => field.onChange(detail.selectedOption)}
+            selectedOption={storedWorkspace ?? null}
+            onChange={({ detail }) => onChangeHandler(detail, field.onChange)}
             placeholder='Select a workspace'
             loadingText='Loading workspaces...'
             errorText='Failed to load workspaces.'


packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
```diff
@@ -16,16 +18,30 @@ export interface WorkspaceSelectorProps {
...
   OnGettingError: (isError: boolean) => void;
+  setValue: UseFormSetValue<SearchFields>;
 }
 
-export const WorkspaceSelector = ({ control, client, OnGettingError }: WorkspaceSelectorProps) => {
+export const WorkspaceSelector = ({ control, client, OnGettingError, setValue }: WorkspaceSelectorProps) => {
   const { workspaces, status } = useWorkspaces({ client });
   const workspaceOptions = createWorkspaceOptions(workspaces);
 
+  const [selectedWorkspace, setSelectedWorkspace] = useLocalStorage<OptionDefinition | null>('selectedWorkspace', null);
```
mnischay, line 28: instead of using null, can it be undefined ?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
```diff
@@ -41,7 +57,10 @@ export const WorkspaceSelector = ({ control, client, OnGettingError }: Workspace
             disabled={status == 'error'}
             options={workspaceOptions}
             selectedOption={field.value}
-            onChange={({ detail }) => field.onChange(detail.selectedOption)}
+            onChange={({ detail }) => {
```
mnischay, line 60: probably having a onChangeHandler function instead of inline

mnischay (changes requested):

mnischay (commented):
https://github.com/awslabs/iot-app-kit/assets/135036199/1074638f-ec42-4ab0-81db-2397e3629951

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
Kausar-HM: We are setting null as defaultValue for 'workspace' form field, so using the same type for localStorage as well
image

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
Kausar-HM: Added onChangeHandler

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
Kausar-HM: No, null is used as defaultValue for workspace form field so we have used the same
image

Kausar-HM:
> Screen.Recording.2023-12-28.at.10.37.26.AM.mov 
> looks this there is a bug,,,, the value flips after the first render as soon as you click on search

@mnischay 
@nadeswal

We watched the recording with 0.25 playback speed, and turns our the flipping is between a placeholder text 'Select workspace' and then the defaultValue is being set.

Here is a video with delay of few seconds to see what is happening in slow motion:
[When-placeholder-enabled.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/e6f74bb1-0ace-4bf6-8351-ea8ef08ae579)

If we disable that placeholder, we will not see it on the screen before setting the defaultValue for that fraction of time:
[when-placeholder-disabled.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/a26885a3-1afe-43aa-9b83-4a976c629f67)

The workspaces-list API takes around 0.3 to 0.5 seconds to return response.

Please suggest what should be done here?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
```diff
@@ -16,16 +18,34 @@ export interface WorkspaceSelectorProps {
...
+  setValue: UseFormSetValue<SearchFields>;
 }
 
-export const WorkspaceSelector = ({ control, client, OnGettingError }: WorkspaceSelectorProps) => {
+export const WorkspaceSelector = ({ control, client, OnGettingError, setValue }: WorkspaceSelectorProps) => {
   const { workspaces, status } = useWorkspaces({ client });
   const workspaceOptions = createWorkspaceOptions(workspaces);
 
+  const [selectedWorkspace, setSelectedWorkspace] = useLocalStorage<OptionDefinition | null>('selectedWorkspace', null);
+  if (selectedWorkspace) setValue('workspace', selectedWorkspace);
```
mnischay, line 29: Do you need to set the value of workspace on every render ?

mnischay, line 29: Do you need to set the value of workspace on every render ?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
```diff
@@ -16,16 +18,34 @@ export interface WorkspaceSelectorProps {
...
+
   useEffect(() => {
     OnGettingError(status === 'error');
   }, [status, OnGettingError]);
 
+  useEffect(() => {
+    if (workspaceOptions.length) {
+      if (!selectedWorkspace) setSelectedWorkspace(workspaceOptions.at(0));
```
mnischay, line 37: I am a bit confused on this logic , 

if (No workspace) --> i.e data still fetching , set ""
if(no localstorage workspace or localstorage workspace not in workspaceslist ) --> set the first workspace.. 

1. i think the "selected" keywork is a bit confusing, i would change it to stored or something... 

Also, the line 37, dont you have to set the value as well ?

mnischay (changes requested):

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
Kausar-HM: Yes, without this we cannot avoid the flickering issue between renders.

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
Kausar-HM: We have modified the code to be more readable now, and simplifies the condition. 
We don't need setValue again since we update the storedWorkspace and setValue will be handled by line 29.

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/dataStreamSearch/workspaceSelector/workspaceSelector.tsx
```diff
@@ -16,20 +18,46 @@ export interface WorkspaceSelectorProps {
...
   client,
   OnGettingError,
+  setValue,
 }: WorkspaceSelectorProps) => {
   const { workspaces, status } = useWorkspaces({ client });
   const workspaceOptions = createWorkspaceOptions(workspaces);
 
+  const [storedWorkspace, setStoredWorkspace] =
+    useLocalStorage<OptionDefinition | null>('storedWorkspace', null);
+  if (storedWorkspace) setValue('workspace', storedWorkspace);
```
mnischay, line 35: I still do not know why we are setting the value on every render .... am i missing something ?

Kausar-HM: @mnischay 
We changed our approach, now there is no need for doing this on each render, the code changes have been reduced to one file only. Please review. Functionality wise there is no change, the same videos in description can be referred.

mnischay (approved):

Issue #2427, opened 2023-12-28
line style needs to be disabled for scatter chart 


mnischay:
**Describe the bug**
[Bug] per property line style needs to be disabled for scatter chart (chart line style is none )

### AC

- [ ] when scatter chart is selected the line style dropdown should be none and disabled

PR #2432, opened 2023-12-29
feat: when scatter chart is selected the line style dropdown should b…

Priyanshu44:
Issue: https://github.com/awslabs/iot-app-kit/issues/2427

Verification:
[x-y-chart.webm](https://github.com/awslabs/iot-app-kit/assets/81667589/c9568c6c-e34b-4ea0-b4db-e367abe269c6)

Fix:
-> when scatter chart is selected the line style dropdown should be none and disabled


mnischay (changes requested):
The testing video attached does not show the AC.. the chart type is non-scatter but the line style is disabled .... which is not what we want

Priyanshu44:
> The testing video attached does not show the AC.. the chart type is non-scatter but the line style is disabled .... which is not what we want

Updated the video and changes.

mnischay (changes requested):
looks like you have done per property but not at widget level

mnischay (approved):

PR #2434, opened 2024-01-02
fix(react-components): clear ymin and ymax was getting emitted on eve…

mnischay:
clear ymin and ymax was getting emitted on every loop

## Overview
The per chart state was getting updated on every loop

BEFORE:

https://github.com/awslabs/iot-app-kit/assets/135036199/32a34816-fecb-4ada-8d85-18e8fa637940

AFTER :

https://github.com/awslabs/iot-app-kit/assets/135036199/532162a9-d388-4fe7-b491-9b3eefcb2356

## Legal

### DIFF
--- a/packages/react-components/src/echarts/extensions/yAxisSync/updateYAxis.ts
+++ b/packages/react-components/src/echarts/extensions/yAxisSync/updateYAxis.ts
@@ -1,5 +1,6 @@
 import minBy from 'lodash.minby';
 import maxBy from 'lodash.maxby';
+import isEqual from 'lodash.isequal';
 
 import LineSeriesModel from 'echarts/types/src/chart/line/LineSeries';
 import { hasCustomYAxis } from './yAxisPredicates';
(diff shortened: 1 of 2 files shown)


jmbuss:
This was put in place as a safe guard to clear the old custom y-axis values when a user toggles a custom y-axis on and off. We need to find another way to prevent the loop

jmbuss (changes requested):
y-axis-bug

mnischay:
https://github.com/awslabs/iot-app-kit/assets/135036199/ea70293c-1bee-480b-87fc-290428324679

jmbuss (approved):

packages/react-components/src/echarts/extensions/yAxisSync/updateYAxis.ts
```diff
@@ -1,5 +1,6 @@
 import minBy from 'lodash.minby';
 import maxBy from 'lodash.maxby';
+import { isEqual } from 'lodash';
```
jmbuss, line 3: Can we update this to use the same pattern as the above lodash methods? It enables treeshaking.

jmbuss (dismissed):

mnischay:
https://github.com/awslabs/iot-app-kit/assets/135036199/1aa96676-dc7e-4bb3-9751-a11a6ed2aa32

ssjagad (approved):

PR #2435, opened 2024-01-02
perf: increase batch entry and result size

tracy-french:
## Overview
This change improves performance of the SiteWise data source by:
- Increasing the number of latest values requested in a single request from 16 to 128.
- Increasing the number of raw historical data points returned in a single request from 4k to 20k.

## Legal


packages/source-iotsitewise/src/time-series-data/client/batch.spec.ts
```diff
@@ -1,15 +1,231 @@
...
+              maxResults: 10,
+            },
+          ],
+          10,
+        ],
+      ])
+    );
+  });
+
+  it('chunks batches that exceed max entry size (128)', () => {
```
chejimmy, line 84: nit: `128` in test case name can be parametrized with `MAX_RAW_LATEST_REQUEST_ENTRIES`

chejimmy (approved):
LGTM 🚀

PR #2440, opened 2024-01-03
fix: fix data-zoom behavior for base chart

jmbuss:
## Overview
Bug fix for data zoom in base chart. In order to de-dupe setting the dataZoom option on the echarts instance, we are not calling setOption if the viewport changes from panning / zooming the chart. Echarts does this internally across all charts that share the same group. We can determine the origin of the setViewport call using the lastUpdatedBy string. Currently, this is being stored in a react state variable. This causes problems if a new chart is created when the lastUpdatedBy value was an echarts gesture as the initial setting of the viewport would be treated as if we had just panned the chart and ignored.

To fix this, I am adding the lastUpdatedBy value to the viewport group update method as a topic, so that in the dataZoom hook, we only use it on events that occur after the chart has officially subscribed to the viewport group.

## Legal


packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -42,25 +44,63 @@ const reducer = (state: TickState, action: TickAction): TickState => {
...
+
+      /**
+       * update the current viewport in the reducer
+       * this ensures that the mode is up-to-date
+       * in the event that the user changes from static -> live
+       * this will restart the tick interval
+       *
+       * this also ensures that the duration is up-to-date
+       * in the event that the user changes from for eg. 1min -> 10min
+       */
```
tracy-french, line 97: Nit: The formatting for your comments is a little hard to understand. I would try to match a max column length of 80 or 120 to keep width consistent. Also, `/** */` is primarily for doc comments. `/* */` is probably the right format to use when you aren't intending to use TSDoc.

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -108,24 +149,36 @@ export const useDataZoom = (chartRef: MutableRefObject<EChartsType | null>, view
...
+    return unsubscribe;
+  }, [handleViewportUpdate, group]);
+
+  /**
+   * ensure the chart is zoomed to the current viewport before it
+   * can start reacting to events
+   */
+  useEffectOnce(() => {
+    handleViewportUpdate(viewport ?? DEFAULT_VIEWPORT);
+  });
```
tracy-french, line 181: Nice! I think using this will prevent bugs from an accidental update of a dependency array, especially when ESLint starts yelling.

packages/react-components/src/components/chart/hooks/useDataZoom.ts
```diff
@@ -42,25 +44,63 @@ const reducer = (state: TickState, action: TickAction): TickState => {
...
-  const { setViewport, lastUpdatedBy } = useViewport();
+  const { setViewport, group } = useViewport();
   const [{ mode, convertedViewport }, dispatch] = useReducer(reducer, stateFromViewport(viewport));
 
   /**
-   * mode will be used to start the tick interval again
-   * once the user sets the viewport back to relative
-   * we also need to update the viewport incase the user
-   * sets it from the chart prop or viewport picker
+   * function for setting the dataZoom chart option on the echart instance
```
tracy-french, line 63: Nit: Can we use sentence casing for doc comments?

tracy-french (approved):
LGTM! Thank you, @jmbuss!

PR #2443, opened 2024-01-03
fix(dashboard): msw composite properties can be added through resource explorer

ssjagad:
## Overview
- created a composite model in the asset Model resources index file and added the model to a both the site and production line model. 
- listAssetProperties handler call takes the composite model properties and adds it to the existing properties

## Legal


packages/dashboard/src/msw/iot-sitewise/handlers/listAssetProperties/listAssetProperties.ts
```diff
@@ -12,8 +12,11 @@ export function listAssetPropertiesHandler() {
       return res(ctx.status(404));
     }
 
+    const properties = asset.assetProperties;
+    asset.assetCompositeModels?.forEach((model) => model.properties?.forEach((property) => properties?.push(property)));
+
     const response: ListAssetPropertiesResponse = {
-      assetPropertySummaries: asset.assetProperties,
+      assetPropertySummaries: properties,
```
tracy-french, line 19: Instead of mutating the properties array, could you create a new array (i.e., `assetPropertySummaries`) which you push both `assetProperties` and all of the asset composite model properties? With the current format, it's possible you are mutating the asset directly and will append all the composite model properties to `asset.assetProperties` each time the handler is called for a specific asset.

packages/dashboard/src/msw/iot-sitewise/resources/assetModels/index.ts
```diff
@@ -8,6 +9,27 @@ const reactorAssetModelMinTemperatureId = uuid();
...
+const compositeAttributeId = uuid();
+
+const assetModelCompositeModels: AssetModelCompositeModel[] = [
+  {
+    id: compositeModelId,
+    name: 'site-composite',
+    type: 'composite-model',
+    properties: [
+      {
+        name: 'status (composite)',
```
tracy-french, line 22: Preferably, try and be creative and make these composite model properties sound like real things which would be in a factory.

packages/dashboard/src/msw/iot-sitewise/resources/assets/AssetFactory.ts
```diff
@@ -33,6 +33,18 @@ export class AssetFactory {
           { id: id, name: name },
         ],
       })) ?? [];
+
+    const assetCompositeModels: AssetCompositeModel[] =
+      this.#assetModel.assetModelCompositeModels?.map(({ id, name, type, properties }) => ({
+        id,
+        name,
+        type,
+        properties: (properties as AssetProperty[]).map((property) => ({
```
tracy-french, line 42: Why do we need to cast here?

tracy-french:
Overall structure of the change is exactly how it should be! Just a few minor changes. Nice work!

diehbria:
closing as stale PR

PR #2449, opened 2024-01-04
fix(react-components): refactor chart to use dataset

jmbuss:
## Overview
Refactor base chart hooks that deal with creating and setting echarts options into 3 hooks.
1. useDataZoom - handles moving the echarts viewport
2. useDataSet - handles adding data to the chart from useTimeSeriesData
3. useChartConfiguration - handles configuring the visual style of the chart including things like datastream styling, title, axis styling, thresholds, tooltip formatting, axis styling, etc.

Refactor for useTimeSeriesData and useVisualizedDataStreams to eliminate extra rerenders and calls to setOption in BaseChart

## Legal


packages/react-components/src/components/chart/chartOptions/useChartDataset.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const convertToDataSet = (dataStreams: DataStream[]) =>
+  dataStreams.map(({ data }) => ({
+    source: data,
+  }));
+
+/**
+ * Hook to set the data points values for each datastream on a chart
+ */
+export const useChartDataset = (
```
tracy-french, line 13: Cool hook!

tracy-french (dismissed):
Thank you for your hard work on this refactor and optimization, @jmbuss! Looking forward to higher performance on the dashboard! :)

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -148,6 +139,14 @@ const BaseChart = ({
     },
   });
 
+  // removed from the chart configuration because it will be refactored into echarts in an upcoming PR
```
mnischay, line 142: nit: probably worth moving into the useTrendcursors hook so all the change wrt TC lives there

packages/react-components/src/components/chart/chartOptions/seriesAndYAxis/convertSeriesAndYAxis.ts
```diff
@@ -216,38 +226,92 @@ export const reduceSeriesAndYAxis = (
...
-    series[0].markArea = convertedThresholds.markArea;
-    series[0].markLine = convertedThresholds.markLine;
-  }
+  const { highlightedDataStreams } = useChartStore(
+    (state) => ({ highlightedDataStreams: state.highlightedDataStreams }),
+    isEqual
+  );
+  const isDataStreamHighlighted = useCallback(
+    (datastream: DataStreamInfo) =>
+      isDataStreamInList(highlightedDataStreams)(datastream),
```
mnischay, line 252: its probably worth to update isDataStreamInList to just accept datastream as a prop and not return a function which accepts datastram

packages/react-components/src/components/chart/chartOptions/seriesAndYAxis/convertSeriesAndYAxis.ts
```diff
@@ -216,38 +226,92 @@ export const reduceSeriesAndYAxis = (
...
+      series[0].markArea = convertedThresholds.markArea;
+      series[0].markLine = convertedThresholds.markLine;
+    }
+
+    return {
+      series,
+      yAxis,
+    };
+  }, [
+    datastreams,
```
mnischay, line 306: i believe the dependencies change on every render... so not worth adding memo, unless all there are also memorized or use state

packages/dashboard/src/customization/widgets/lineScatterChart/component.tsx
```diff
@@ -122,13 +122,21 @@ const mapStyledAssetPropertyToChartStyleSettings = (
...
 
-const convertAxis = (axis: ChartAxisOptions | undefined) => ({
-  showY: axis?.yVisible,
-  showX: axis?.xVisible,
-  yMin: axis?.yMin,
-  yMax: axis?.yMax,
-  yLabel: axis?.yLabel,
-});
+// memoizing so any change to the chart does not cause the axis to create a new reference
+const useConvertedAxis = (axis: ChartAxisOptions | undefined = {}) => {
```
mnischay, line 126: nit: axis?: ChartAxisOptions = {}

packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
```diff
@@ -0,0 +1,183 @@
+import {
```
mnischay, line 1: can we add some tests ? given this only deals with data

mnischay (changes requested):

packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
jmbuss: This is an aggregation of hooks / converters we already had created and tested so I didn't think it was necessary. useSeriesAndYAxis is the most complicated part of this hook and we already test that.

packages/react-components/src/components/chart/baseChart.tsx
jmbuss: 👍

packages/react-components/src/components/chart/chartOptions/seriesAndYAxis/convertSeriesAndYAxis.ts
jmbuss: No, we need this memo to prevent the setOption in useChartConfiguration from being excessively called. I had the same thought, but tested with and without and it is required.

packages/react-components/src/components/chart/chartOptions/seriesAndYAxis/convertSeriesAndYAxis.ts
mnischay: any thoughts ?

packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
```diff
@@ -0,0 +1,183 @@
...
+      text: hasSeries ? titleText ?? '' : 'No data present',
+    }),
+    [hasSeries, titleText]
+  );
+};
+
+const toDataStreamIdentifiers = (dataStreams: DataStream[]) =>
+  dataStreams.map(({ id, name, color, refId }) => ({ id, name, color, refId }));
+
+type ChartConfigurationOptions = Pick<
```
mnischay, line 41: nit: probably worth cleaning this up a bit or just creating a new type object

packages/react-components/src/components/chart/chartOptions/seriesAndYAxis/convertSeriesAndYAxis.ts
jmbuss: Can we address this change in a separate PR? That change will also affect the legend table, and this PR is already quite big.

packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
mnischay: Fair

packages/react-components/src/components/chart/trendCursor/useTrendCursors.ts
```diff
@@ -150,6 +150,14 @@ const useTrendCursors = ({
...
   );
 
+  // removed from the chart configuration because it will be refactored into echarts in an upcoming PR
+  // will only live here temporarily
+  useEffect(() => {
+    chartRef.current?.setOption({
+      graphic: orderedTrendCursors,
+    });
+  }, [chartRef, orderedTrendCursors]);
+
```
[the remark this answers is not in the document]
jmbuss: yes, that was a requested feature, to sort them by time

packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
jmbuss: Can we address nits in a separate PR? I would like to get this into rc soon so that it can be validated by everyone working on features.

packages/react-components/src/components/chart/trendCursor/useTrendCursors.ts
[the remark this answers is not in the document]
mnischay: no, i meant this hook returns the array, previously we were setting the options elsewhere, and returning tc array made sense, now no other parts of charts needs this tc array ?

mnischay (dismissed):
Approving for now, with the expectation that the nits will be fixed in a fast follow CR

tracy-french (approved):

PR #2464, opened 2024-01-09
fix: source-iotsitewise batch API options

chejimmy:
## Overview

The following options could be all true at a given time `fetchMostRecentBeforeStart`, `fetchMostRecentBeforeEnd`, and `fetchFromStartToEnd`; however, the current code logic does not handle them correctly when more than 1 option at a time.

This PR fixes the issue aforementioned.

more about what the options should do, see [doc](https://awslabs.github.io/iot-app-kit/?path=/docs/core-time-series-data--docs)

### Scene Composer

Manual testings - panning with aggregation/row data

## Legal

### DIFF
--- a/packages/source-iotsitewise/src/time-series-data/client/batch.spec.ts
+++ b/packages/source-iotsitewise/src/time-series-data/client/batch.spec.ts
@@ -237,6 +237,83 @@ describe('createRawHistoricalEntryBatches', () => {
     );
   });
 
+  it.each(['fetchMostRecentBeforeEnd', 'fetchMostRecentBeforeStart'])(
+    'separate all %s entries',
+    (mostRecentOption) => {
+      const onError = jest.fn();
+      const onSuccess = jest.fn();
+      const startDate = new Date(2000, 0, 0);
+      const endDate = new Date(2001, 0, 0);
+      const resolution = '0';
+
+      const batches = createRawHistoricalEntryBatches([
+        {
+          requestInformation: {
+            id: '1',
+            resolution,
+            [mostRecentOption]: true,
+          },
+          maxResults: 1,
+          onError,
+          onSuccess,
+          requestStart: startDate,
+          requestEnd: endDate,
+        },
+        {
+          requestInformation: {
+            id: '2',
+            resolution,
+            [mostRecentOption]: true,
+          },
+          maxResults: 1,
+          onError,
+          onSuccess,
+          requestStart: startDate,
+          requestEnd: endDate,
+        },
+      ]);
+
+      expect(batches).toEqual(
+        expect.arrayContaining([
+          [
+            [
+              {
+                requestInformation: {
+                  id: '1',
+                  resolution,
+                  [mostRecentOption]: true,
+                },
+                maxResults: 1,
+                onError,
+                onSuccess,
+                requestStart: startDate,
+                requestEnd: endDate,
+              },
+            ],
+            1,
+          ],
+          [
+            [
+              {
+                requestInformation: {
+                  id: '2',
+                  resolution,
+                  [mostRecentOption]: true,
+                },
+                maxResults: 1,
+                onError,
+                onSuccess,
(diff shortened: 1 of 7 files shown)


packages/source-iotsitewise/src/time-series-data/client/batch.spec.ts
```diff
@@ -237,6 +237,83 @@ describe('createRawHistoricalEntryBatches', () => {
     );
   });
 
+  it.only.each(['fetchMostRecentBeforeEnd', 'fetchMostRecentBeforeStart'])(
```
tracy-french, line 240: Looks like you left 2 `it.only` calls on accident?

packages/source-iotsitewise/src/time-series-data/client/batch.ts
```diff
@@ -48,11 +49,19 @@ export function createRawLatestEntryBatches<T extends { maxResults?: number }>(
 }
 
 export function createRawHistoricalEntryBatches<
-  T extends { maxResults?: number }
+  T extends { requestInformation?: RequestInformation; maxResults?: number }
 >(entries: T[]): [T[], number][] {
   const buckets: Record<number, T[]> = {};
+  const singles: [T[], number][] = [];
```
tracy-french, line 55: Nit: It would be beneficial to have a comment here (or somewhere) explaining why this is done (i.e., prevent grouping of singles).

packages/source-iotsitewise/src/time-series-data/client/batch.spec.ts
```diff
@@ -237,6 +237,83 @@ describe('createRawHistoricalEntryBatches', () => {
...
+    'separate all %s entries',
+    (mostRecentOption) => {
+      const onError = jest.fn();
+      const onSuccess = jest.fn();
+      const startDate = new Date(2000, 0, 0);
+      const endDate = new Date(2001, 0, 0);
+      const resolution = '0';
+
+      const batches = createRawHistoricalEntryBatches([
```
tracy-french, line 249: Would it be valuable to add or extend test cases to show the single data point requests are not combined with requests of other sizes? Or is this already covered in the existing tests?

packages/source-iotsitewise/src/time-series-data/client/batchGetAggregatedPropertyDataPoints.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      params: [param],
+      client,
+    });
+
+    await flushPromises();
+
+    expect(batchGetAssetPropertyAggregates).toBeCalledTimes(1);
+    expect(batchGetAssetPropertyAggregates).toHaveBeenCalledWith(
+      batchParamForStartToEnd
+    );
```
tracy-french, line 96: Would it be worth testing is `onSuccess` is called and `onError` is not called?

tracy-french (changes requested):
Overall looks great! Well done! The only thing that looks necessary to change before merging this change is the `it.only` calls in the batch tests.

chejimmy:
> Overall looks great! Well done! The only thing that looks necessary to change before merging this change is the `it.only` calls in the batch tests.

good catch 🙏

packages/source-iotsitewise/src/time-series-data/client/batch.spec.ts
chejimmy: that's the intent of this test; @line#276, it expects 2 batches of requests 🙏

packages/source-iotsitewise/src/time-series-data/client/batchGetAggregatedPropertyDataPoints.spec.ts
chejimmy: I omitted testing those in the batch APIs since there are coverage in the client.spec.ts.

tracy-french (approved):

PR #2465, opened 2024-01-09
feat: async listAssetPropertiesDescription

corteggiano:
## Overview
create async version of `useListAssetPropertiesMapQuery`
and add shared logic into nicely organized files :) 

had to pull out shared code in order to preserve existing caching logic
by using the same cache key for both the hook and the async function, they can share a cache thanks to the magic of tanstack :) 

 
## Legal

### DIFF
--- a/packages/dashboard/src/data/listAssetPropertiesMap/fetchListAssetPropertiesMap.tsx
+++ b/packages/dashboard/src/data/listAssetPropertiesMap/fetchListAssetPropertiesMap.tsx
@@ -1,0 +1,32 @@
+import { IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
+import {
+  SiteWiseAssetQuery,
+  SiteWisePropertyAliasQuery,
+} from '@iot-app-kit/source-iotsitewise';
+import { QueryClient } from '@tanstack/react-query';
+import {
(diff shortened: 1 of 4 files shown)


packages/dashboard/src/data/listAssetPropertiesMap/query.tsx
```diff
@@ -0,0 +1,47 @@
+import { IoTSiteWiseClient } from '@aws-sdk/client-iotsitewise';
```
corteggiano, line 1: nothing new, just moved from the useAssetDescriptions file

packages/dashboard/src/data/listAssetPropertiesMap/fetchListAssetPropertiesMap.tsx
```diff
@@ -0,0 +1,32 @@
...
+    siteWiseQuery
+  );
+
+  const data = await queryClient.fetchQuery({
+    queryKey,
+    queryFn,
+  });
+
+  return selectListAssetPropertiesMap(data);
+};
```
tracy-french, line 32: This is exactly how we discussed. Glad to see everything you needed was available!

tracy-french (approved):
Cool!

PR #2478, opened 2024-01-15
fix: make context menu appear on top of chart tooltip

dpportet:
## Overview
* Fix for an issue where chart tooltips will overlap and display on top of a charts context menu, which makes it harder to use the context menu since it should always be on top
* This change increases the z-index for the context menu such that it appears on top of the tooltip
* I wasn't sure of a better solution than just hardcoding z-index greater than the tooltip, so let me know if you have any other ideas

## Verifying Changes
1. Create any chart
2. Hover mouse over chart to see tooltip
3. Right click to see context menu
4. Validate context menu displays on top of tooltip

## Legal

## Demo

https://github.com/awslabs/iot-app-kit/assets/66272633/b39c68b4-9558-4618-bf5f-24506234a7a3

### DIFF
--- a/packages/react-components/src/components/chart/contextMenu/ChartContextMenu.tsx
+++ b/packages/react-components/src/components/chart/contextMenu/ChartContextMenu.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Menu, MenuOption } from '../../menu';
 
+import { CONTEXT_MENU_Z_INDEX } from '../eChartsConstants';
 import { MAX_TREND_CURSORS } from '../trendCursor/constants';
 import { InternalGraphicComponentGroupOption } from '../trendCursor/types';
 
@@ -29,6 +30,7 @@ const ChartContextMenu = ({
         position: 'absolute',
         top: `${position.y}px`,
         left: `${position.x}px`,
+        zIndex: CONTEXT_MENU_Z_INDEX,
       }}
       onClickOutside={onOutSideClickHandler}
     >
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/chart/contextMenu/ChartContextMenu.tsx
```diff
@@ -23,12 +23,19 @@ const ChartContextMenu = ({
   onOutSideClickHandler,
   trendCursors,
 }: ChartContextMenu) => {
+  /**
+   * Echarts tooltip has z-index of 9999999.
+   * Must set context menu above this so its always displayed on top.
+   */
+  const toolTipZIndex = 9999999;
```
mnischay, line 30: This should be a constant.. i believe we have a dedicate file for these constants. 
Also, not sure if we need 2 variables here, the comments gives exellent context as to why the value is so high, so probably can do with just one vairable `MENU_Z_INDEX = 10000000`

packages/react-components/src/components/chart/contextMenu/ChartContextMenu.tsx
```diff
@@ -23,12 +23,19 @@ const ChartContextMenu = ({
   onOutSideClickHandler,
   trendCursors,
 }: ChartContextMenu) => {
+  /**
+   * Echarts tooltip has z-index of 9999999.
```
mnischay, line 27: NIce comments!!!!

mnischay (dismissed):

mnischay (dismissed):

packages/react-components/src/components/chart/contextMenu/ChartContextMenu.tsx
dpportet: Sounds good, I updated to make it a constant

mnischay (approved):

PR #2490, opened 2024-01-17
chore(scene-composer): remove old feature flags part 4 - EnvironmentModel

hwandersman:
## Overview
Removed the `EnvironmentModel` feature flag and deprecated it's functionality since we do not plan to use it in the future.

## Verifying Changes

Ran unit tests and verified functionality of local scene composer and viewer on storybook.

Tested adding every widget type to the scene composer with and without a parent. All previous functionality is maintainted. 

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/components/three-fiber/EditorTransformControls.spec.tsx
```diff
@@ -40,11 +40,9 @@ jest.doMock('../../../src/three/TransformControls', () => {
 });
 
 const mockFindComponentByType = jest.fn();
-const mockisEnvironmentNode = jest.fn();
 jest.doMock('../../../src/utils/nodeUtils', () => {
```
mitchlee-amzn, line 43: nit: Not part of your CR, but when you see usages of `jest.doMock` we should refactor them to use `jest.mock` properly.

doMock doesn't hoist the mocking logic up, so we still end up with importing all the dependencies we are mocking, and it leads to inconsistencies and obscure test issues.

mitchlee-amzn (approved):
Nice, thanks for the effort!

PR #2498, opened 2024-01-22
fix: hidden and highlighted datastreams persist correctly

ssjagad:
## Overview
This change solves https://app.asana.com/0/1204375185707273/1206385082271207 using persist() in the Zustand library. https://docs.pmnd.rs/zustand/integrations/persisting-store-data#hashydrated

- Linked and persisted each chartStore with the widget id, so each store has its own unique store that remains populated with the correct data through edit/preview modes and refresh. 

- Persisted the global store, so that it does not reset to an empty object on edit/preview modes and refresh

## Verifying Changes

These cases are outlined in the ticket linked above. I tested more within each video to try to catch edge cases. 

Case 1:

https://github.com/awslabs/iot-app-kit/assets/145582655/3e21a6d1-318c-4411-bd47-0ebb7d1759c2

Case 2:

https://github.com/awslabs/iot-app-kit/assets/145582655/2695ea46-ddd9-48c7-a388-a60cbce60440

Case 3:

https://github.com/awslabs/iot-app-kit/assets/145582655/04bc1983-47f2-4a76-a8f7-f8b75e9568bf

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/store/context.tsx
+++ b/packages/react-components/src/components/chart/store/context.tsx
@@ -10,7 +10,7 @@ import useDataStore from '../../../store';
 // Recommended usage from zustand https://docs.pmnd.rs/zustand/previous-versions/zustand-v3-create-context
 export const ChartStoreContext = createContext<
   ReturnType<typeof createChartStore>
->(createChartStore());
+>(createChartStore({ chartId: 'contextStore' }));
 
 export const ChartStoreProvider = ({
   id,
--- a/packages/react-components/src/components/chart/store/store.ts
+++ b/packages/react-components/src/components/chart/store/store.ts
@@ -3,16 +3,21 @@ import { devtools, persist } from 'zustand/middleware';
 import { createMultiYAxisSlice, MultiYAxisState } from './multiYAxis';
 import { createDataStreamsSlice, DataStreamsState } from './contextDataStreams';
 
-export type StateData = DataStreamsState & MultiYAxisState;
-export const createChartStore = () =>
+export type StateData = DataStreamsState &
+  MultiYAxisState & { chartId: string };
+
+export const createChartStore = (
+  initProps: Partial<StateData> & Required<Pick<Partial<StateData>, 'chartId'>>
+) =>
   create<StateData>()(
     devtools(
       persist(
         (...args) => ({
+          ...initProps,
           ...createMultiYAxisSlice(...args),
           ...createDataStreamsSlice(...args),
         }),
-        { name: 'chartStore' }
+        { name: `chart-store-${initProps?.chartId}` }
       )
     )
   );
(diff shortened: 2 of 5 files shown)


packages/react-components/src/components/chart/store/context.tsx
```diff
@@ -10,7 +10,7 @@ import useDataStore from '../../../store';
 // Recommended usage from zustand https://docs.pmnd.rs/zustand/previous-versions/zustand-v3-create-context
 export const ChartStoreContext = createContext<
   ReturnType<typeof createChartStore>
->(createChartStore());
+>(createChartStore({ chartId: 'contextStore' }));
```
jmbuss, line 13: is there a way to indicate to the persist middleware not to persist something? This store should actually never be used, we only create the reference here because typescript would complain otherwise. I think if it's possible, we shouldn't save it since it would just be taking up space best case and a potential for bugs worst case.

jmbuss: Maybe it's as simple as persisting the store only if an id is provided to the `createChartStore`? and just doing a ternary in the function?

packages/react-components/src/components/chart/store/store.ts
```diff
@@ -3,16 +3,22 @@ import { devtools, persist } from 'zustand/middleware';
 import { createMultiYAxisSlice, MultiYAxisState } from './multiYAxis';
 import { createDataStreamsSlice, DataStreamsState } from './contextDataStreams';
 
-export type StateData = DataStreamsState & MultiYAxisState;
-export const createChartStore = () =>
+export type StateData = DataStreamsState &
+  MultiYAxisState & { chartId: string; name: string };
```
jmbuss, line 7: We can probably get rid of name here since actually chartId is the name.

packages/react-components/src/components/chart/store/store.ts
```diff
@@ -3,16 +3,22 @@ import { devtools, persist } from 'zustand/middleware';
...
     devtools(
       persist(
         (...args) => ({
+          ...initProps,
+          name: initProps.chartId,
           ...createMultiYAxisSlice(...args),
           ...createDataStreamsSlice(...args),
         }),
-        { name: 'chartStore' }
+        { name: `${initProps?.chartId}` }
```
jmbuss, line 21: Can we prefix this with `chart-store-<id>` or something similar? It will help in the future looking through local storage in devtools to know what we're looking at since the ids are just random guids.

jmbuss (changes requested):

packages/react-components/src/components/chart/store/context.tsx
jmbuss: Also, maybe instead of actually creating the store reference in the context, we should just use null/undefined as the initial value, and then use invariants to safeguard and not have to null coalesce everything. I would have to play around with that to see if it's actually doable without a big structural change, so don't feel like you need to make that change, just another thought!

ssjagad: Setting it to null makes sense since that is technically the initial value of any store.

ssjagad: Made a ticket for this, so this gets in and removes the weird highlight/hidden persistence issues

jmbuss (approved):

PR #2508, opened 2024-01-25
feat(dashboard): adding new property panel layout

mnischay:
## Overview
This CR is adding a layout for the new property panel
1. This feature is feature flagged and user will NOT see this new workflow
2.  new playground parent component
3. A new proper panel drawer 
4. A new modal for the RE
5. config panel and Resources explorer components are untouched 
6.  The table component in the property panel is WIP (shifting focus to testing as we discussed in sprint planning) , the data hook is just a copy of the existing property panel component. 

The only change which this CR will affect the current user is that of the CSS change of the chart's legend resize handler

#### The table component and its corresponding data hook is WIP and hence not tested. Both the files will change significantly in the follow up CR. 
prop-panel (1)

## Verifying Changes
Refer to the attached screen capture

https://github.com/awslabs/iot-app-kit/assets/135036199/81038ebc-f6da-4a3f-a391-26c15764edb7

## Legal


packages/dashboard/src/components/customOrangeButton/index.tsx
```diff
@@ -8,8 +8,16 @@ import './index.css';
 const CustomOrangeButton = ({
   title,
   handleClick,
+  disabled,
   ...rest
 }: { title: string; handleClick?: () => void } & ButtonProps) => {
+  if (disabled) {
```
corteggiano, line 14: at this point should we just make our own orange button so we're not overriding all the cloudscape settings for every possible state

packages/dashboard/src/components/playground/propertiesTable/propertiesTable.tsx
```diff
@@ -0,0 +1,267 @@
...
+        stickyHeader
+        onSelectionChange={collectionProps?.onSelectionChange}
+        selectedItems={collectionProps.selectedItems}
+        loadingText='Loading resources'
+        selectionType='multi'
+        trackBy='key'
+        contentDensity='compact'
+        resizableColumns
+        stickyColumns={{ first: 1 }}
+        columnDefinitions={[
```
corteggiano, line 88: can all the larger props be moved into their own files for legibility

`columnDefinition`
`emptyState`
`header`
`preferences`

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -285,32 +287,63 @@ const InternalDashboard: React.FC<InternalDashboardProperties> = ({
...
-                {activeGesture === 'select' && (
-                  <UserSelection {...selectionProps} />
-                )}
-              </GestureableGrid>
-              <WebglContext viewFrame={viewFrame} />
-            </div>
-          }
-          rightPane={propertiesPanel}
-        />
+        {showPlayground() ? (
```
corteggiano, line 290: can we split this conditional into its own components instead of doing one large return? that way its easier to debug

packages/dashboard/src/components/customOrangeButton/index.tsx
mnischay: this is the only state we need explicitly, i think it should be ok for now... any more customization, we probably should fall back to the "blue" primary.. instead the internal orange promary

packages/dashboard/src/components/playground/propertiesTable/propertiesTable.tsx
mnischay: Yes, this file is 100% WIP, and hence not changing much here. yes this file is the first one to change in the follow up CR

packages/dashboard/src/components/internalDashboard/index.tsx
mnischay: Sure throwing the creator into a component should help with readability

packages/dashboard/src/components/playground/propertyDrawer/propertyDrawer.tsx
```diff
@@ -0,0 +1,109 @@
...
+  // opening the drawer when user selects a widget
+  useEffect(() => {
+    if (
+      prevSelectedWidgets.current.length === 0 &&
+      selectedWidgets.length === 1
+    ) {
+      setIsDrawerCollapsed(false);
+    }
+    prevSelectedWidgets.current = selectedWidgets;
+  }, [selectedWidgets, selectedWidgets.length]);
```
tracy-french, line 56: This seems more complicated than it needs to be, meaning the usage of a ref and an effect? Is there a simpler way? Can it be encapsulated?

packages/dashboard/src/components/playground/playground.tsx
```diff
@@ -0,0 +1,55 @@
...
+import PropertiesPanelIcon from '~/components/resizablePanes/assets/propertiesPane.svg';
+import { CollapsiblePanel } from '~/components/internalDashboard/collapsiblePanel';
+import {
+  CONFIG_PANEL_COLLAPSE_WIDTH,
+  CONFIG_PANEL_OPEN_WIDTH,
+} from '~/components/playground/constants';
+import './playground.css';
+import { PropertyDrawer } from '~/components/playground/propertyDrawer/propertyDrawer';
+
+export const Playground = ({
```
tracy-french, line 11: I'm confused by the naming of playground. Is there a more descriptive name we can use?

packages/dashboard/src/components/playground/drawerHeader/drawer-header.tsx
```diff
@@ -0,0 +1,37 @@
+import { Button } from '@cloudscape-design/components';
+import React from 'react';
+import {
+  spaceScaledXs,
+  spaceScaledXxxs,
+} from '@cloudscape-design/design-tokens';
+import './drawer.css';
+export const DrawerHeader = ({
```
tracy-french, line 8: There are a few formatting inconsistencies in this change, including this (lack of newline) and double newlines in other places.

tracy-french (changes requested):
I would expect there to be tests in this change. Why are there none?

packages/dashboard/src/components/playground/drawerHeader/drawer-header.tsx
mnischay: i ran npm run lint, guess something in eslint is broken ?

packages/dashboard/src/components/playground/playground.tsx
mnischay: I wanted a less confusing name than everything either has panel/panes or dashbaord in them. but yes, this will be changed

packages/dashboard/src/components/playground/propertyDrawer/propertyDrawer.tsx
mnischay: this behaviour only when the value goes from not-selected to  selected, guess I could make it simpler by looking for a value change rather than transition

mnischay:
@tracy-french : good chunk of this CR will be changed in the follow up CR, especially the table component.. and hence no test for it ... have noted this in the description ....

packages/dashboard/src/components/playground/drawerHeader/drawer-header.tsx
tracy-french: Might not test newline behavior.

tracy-french:
> @tracy-french : good chunk of this CR will be changed in the follow up CR, especially the table component.. and hence no test for it ... have noted this in the description ....

It's much easier to test as you go thann at the end. I would prefer we test incrementally. You're likely to miss behavior and we're more likely to feel compelled to release without completing testing.

Alternatively, if you cannot complete the tests now, could we put this in a separate branch than in `rc`? I know there's a feature flag, yet the code is already there and I think that increases the risk that we don't test effectively in the end.

packages/dashboard/src/components/playground/propertiesTable/propertiesTable.tsx
tracy-french: I don't understand why not do this now? Why does the WIP state mean we don't improve it now?

corteggiano: +1 would push for this to be done sooner rather than later 🙏 
WIP doesn't guarantee you will return to make the changes

mnischay:
closing this out, given we are focusing on testing for now, will take this on in a week

PR #2514, opened 2024-01-25
chore: spliting unit tests execution per package

mnischay:
## Overview
spliting up the unit test execution to individual jobs to provide scaling

## Verifying Changes
check the steps below and their respective logs

## Legal

### DIFF
--- a/.github/workflows/composite/action.yml
+++ b/.github/workflows/composite/action.yml
@@ -1,0 +1,23 @@
+name: "Testing given package's unit tests"
+description: 'Builds and tests the package specified by repo-name input'
+inputs:
+  repo-path:
+    required: true
+    description: "Provide the package's path to be tested"
+runs:
+  using: 'composite'
+  steps:
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 16
+        cache: 'npm'
+        cache-dependency-path: '**/package-lock.json'
+
+    - name: Install dependencies
+      shell: bash
+      run: npm run bootstrap
+
+    - name: Run unit tests
+      shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/composite/action.yml
```diff
@@ -0,0 +1,23 @@
...
+  using: 'composite'
+  steps:
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 16
+        cache: 'npm'
+        cache-dependency-path: '**/package-lock.json'
+
+    - name: Install dependencies
+      shell: bash
```
tracy-french, line 17: Why do you need to specify bash shell? We haven't had this before.

tracy-french (changes requested):
Looks nice! Glad we could keep everything within one file. 

Could you please point the PR to `main` instead of `rc` so that we fix the issue for everyone?

.github/workflows/composite/action.yml
tracy-french: Interesting. Didn't know!

mnischay:
@tracy-french : we will have a release mid next week so might as well release it along with all the changes ? just pointing this to main will not work given this is branch is based off of rc ... i dont think rebasing main on to this locally will work... my guess is I need to create an entirely new CR in that case

tracy-french:
> @tracy-french : we will have a release mid next week so might as well release it along with all the changes ? just pointing this to main will not work given this is branch is based off of rc ... i dont think rebasing main on to this locally will work... my guess is I need to create an entirely new CR in that case

Cherry-pick the commit onto `main` and create a new PR.

PR #2515, opened 2024-01-25
fix: kpi add prop bug

mnischay:
## Overview
The issue was that we were not checking the selection properties count
when we add  a kPI / status, we are allowing the user to add more than 1 property

this particular bug was becasue, when we selected kpi, we check if it already has properties added but did not check if more than 1 is selected. given for other widget type we can add more than one, the logic was obviously wrong.

Also, this was done only for modeled data and had missed unmodeled and asset models
## Verification

https://github.com/awslabs/iot-app-kit/assets/135036199/68a9f83f-1734-49ea-a5b7-aab9c5683827

## Legal
This project is available under the [Apache 2.0 License](http://www.apache

.org/licenses/LICENSE-2.0.html).

### DIFF
--- a/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
+++ b/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
@@ -1,11 +1,14 @@
 import { disableAdd } from '~/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd';
+import { SiteWiseQueryConfig } from '~/customization/widgets/types';
 
-describe(disableAdd, () => {
-  it('should return true if no selected widgets', () => {
-    expect(disableAdd([], 0)).toBeTruthy();
-  });
-
-  it('should return true if no results', () => {
+const commonTests = ({
+  objectWithoutProperties,
+  objectWithProperties,
+}: {
+  objectWithoutProperties: { queryConfig: SiteWiseQueryConfig };
+  objectWithProperties: { queryConfig: SiteWiseQueryConfig };
+}) => {
+  it('should return true if no results, with or without properties', () => {
     expect(
       disableAdd(
         [
@@ -15,7 +18,24 @@ describe(disableAdd, () => {
             x: 0,
             y: 0,
             z: 0,
-            properties: {},
+            properties: objectWithoutProperties,
+            height: 500,
+            width: 800,
+          },
+        ],
+        0
+      )
+    ).toBeTruthy();
+    expect(
+      disableAdd(
+        [
+          {
+            id: '1',
+            type: 'status',
+            x: 0,
+            y: 0,
+            z: 0,
+            properties: objectWithProperties,
             height: 500,
             width: 800,
           },
@@ -25,7 +45,7 @@ describe(disableAdd, () => {
     ).toBeTruthy();
   });
 
-  it('should return true if status is selected and has a assets already added', () => {
+  it('should return true if status/kpi is selected and has a property already added', () => {
     expect(
       disableAdd(
         [
@@ -35,7 +55,7 @@ describe(disableAdd, () => {
             x: 0,
             y: 0,
             z: 0,
-            properties: { queryConfig: { query: { assets: [{ id: '1' }] } } },
+            properties: objectWithProperties,
             height: 500,
             width: 800,
           },
@@ -43,8 +63,6 @@ describe(disableAdd, () => {
         10
       )
     ).toBeTruthy();
-  });
-  it('should return true if KPI is selected and has a assets already added', () => {
     expect(
       disableAdd(
         [
@@ -54,7 +72,7 @@ describe(disableAdd, () => {
             x: 0,
             y: 0,
             z: 0,
-            properties: { queryConfig: { query: { assets: [{ id: '1' }] } } },
+            properties: objectWithProperties,
             height: 500,
             width: 800,
           },
@@ -64,7 +82,7 @@ describe(disableAdd, () => {
     ).toBeTruthy();
   });
 
-  it('should return false if status is selected and has NO assets already added', () => {
+  it('should return false if status/kpi is selected and has NO property already added', () => {
     expect(
       disableAdd(
         [
@@ -74,17 +92,15 @@ describe(disableAdd, () => {
             x: 0,
             y: 0,
             z: 0,
-            properties: { queryConfig: { query: { assets: [] } } },
+            properties: objectWithoutProperties,
             height: 500,
             width: 800,
           },
         ],
-        10
+        1
       )
     ).toBeFalsy();
-  });
 
-  it('should return false if KPI is selected and has NO assets already added', () => {
     expect(
       disableAdd(
         [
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.ts
```diff
@@ -12,16 +12,26 @@ export const disableAdd = (
...
     case 'status':
     case 'kpi': {
-      const assets =
+      const modeledProps =
         get(selectedWidget, 'properties.queryConfig.query.assets') ?? [];
-      if (assets.length) {
+      const unmodeledProps =
+        get(selectedWidget, 'properties.queryConfig.query.properties') ?? [];
+      const assetModelProps =
+        get(selectedWidget, 'properties.queryConfig.query.assetModels') ?? [];
```
tracy-french, line 20: Could you change these names to `*Properties` instead of `*Props`. Props is an overloaded term and it adds confusion, especially since there is `collectionProps` in the same function.

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
```diff
@@ -44,37 +44,38 @@ describe(disableAdd, () => {
       )
     ).toBeTruthy();
   });
-  it('should return true if KPI is selected and has a assets already added', () => {
+
+  it('should return false if status/kpi is selected and has NO assets already added', () => {
```
tracy-french, line 48: This test name says status and kpi, but only tests status. Could you test both?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
```diff
@@ -25,7 +25,7 @@ describe(disableAdd, () => {
     ).toBeTruthy();
   });
 
-  it('should return true if status is selected and has a assets already added', () => {
+  it('should return true if status/kpi is selected and has a assets already added', () => {
```
tracy-french, line 28: This test name says status and kpi, but only tests status. Could you test both?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
```diff
@@ -84,7 +85,7 @@ describe(disableAdd, () => {
     ).toBeFalsy();
   });
 
-  it('should return false if KPI is selected and has NO assets already added', () => {
+  it('should return true if KPI is selected and more than 1 property selected', () => {
```
tracy-french, line 88: The test name says more than 1 property selected, yet the properties object is empty?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
```diff
@@ -94,17 +95,17 @@ describe(disableAdd, () => {
...
           },
         ],
         10
       )
-    ).toBeFalsy();
+    ).toBeTruthy();
   });
 
-  it('should return false if x-y-plot is selected and has assets already added', () => {
+  it('should return false if x-y-plot is selected and more than 1 property selected', () => {
```
tracy-french, line 108: The test name says more than 1 property selected, yet the properties object is empty?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
```diff
@@ -123,4 +124,90 @@ describe(disableAdd, () => {
       )
     ).toBeFalsy();
   });
+
+  describe('for unmodeled data', () => {
```
tracy-french, line 128: Missing tests for KPI even though titles say KPI (same comment as above).

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.ts
```diff
@@ -12,16 +12,26 @@ export const disableAdd = (
   switch (currWidgetType) {
     case 'status':
     case 'kpi': {
-      const assets =
+      const modeledProps =
         get(selectedWidget, 'properties.queryConfig.query.assets') ?? [];
-      if (assets.length) {
+      const unmodeledProps =
+        get(selectedWidget, 'properties.queryConfig.query.properties') ?? [];
+      const assetModelProps =
```
tracy-french, line 19: Need tests for asset model properties?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
tracy-french, line 128: Should you also be testing xy-plot here like in the modeled tests?

tracy-french (changes requested):

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
mnischay: I had for both originally, the test file was getting tooo big , and the logic is exactly the same for both, so decided to remove it.. let me add to the same test case

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
mnischay: I had for both originally, the test file was getting tooo big , and the logic is exactly the same for both, so decided to remove it.. let me add to the same test case

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
mnischay: no matter what when we have more than 1 property selected, iif the widget is kpi/status, we disable

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
mnischay: for x-y the properties added doesnt matter

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.ts
mnischay: yes, the last describe block...

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.spec.tsx
mnischay: sure, was trying to keep the file small, no other particular reason

tracy-french (approved):

Issue #2513, opened 2024-01-25
[Accessibility][ConfigPanelThresholds]  Combobox controls rely on option text as a label


jmbuss:
**Describe the bug**
**Standard:** 3.3.2 Labels or Instructions (A)
**Affected Groups:** Cognitively Impaired, Screen Reader Users

**Notes and Advice** 
The "Arithmetic Operator" custom combobox controls rely upon option text as a visible label:
 
<select>
<option disabled>selectable option</option>
<option>1</option>
...
</select>
 
Once an option is selected the control label will no longer be visible.
 Recommendation 
All controls must always have a persistent, visible label (in some cases implicit labels are acceptable).
 
For a HTML <select> element:
 
<label for="octo">selectable option</label>
<select id="octo">
...
</select>
 
For a custom control:
 
<span id="octo">selectable option</span>
<div role="combobox" aria-labelledby="octo">
...
</div>
 Resources 
[Understanding 3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)
[How to Meet 3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/quickref/#labels-or-instructions)

**Acceptance Criteria**
The arithmetic operator should have a label above it using cloudscape which says Operator

Issue #2512, opened 2024-01-25
[Accessibility][ConfigPanelThresholds]  Form controls rely on placeholder text as a label


jmbuss:
**Describe the bug**
**Standard:** 3.3.2 Labels or Instructions (A)
**Affected Groups:** Cognitively Impaired, Screen Reader Users

**Notes and Advice** 
The "Threshold" input field is only labeled by a placeholder attribute.
 
Text marked up using the placeholder attribute will generally be exposed to assistive technologies, so is announced by screen readers. But for other user groups, relying on placeholder text on its own to label form fields is a problematic approach, outlined in [the HTML5 specification](https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#the-placeholder-attribute). Issues include:
 
Placeholder text disappears when a user starts typing. This means that there is no visible label when users are entering text into the input, which could affect users with short-term memory issues.
Placeholder text is not shown when the value of a field is pre-filled. This means that there is no visible label when users are reviewing their input or making corrections.
When placeholder text shows an example value, and does not describe the type of input the field collects, it is not a reliable label and could cause confusion for some users.
People with limited dexterity cannot take advantage of the programmatic association of a <label> element to the control using the for and id attributes, which provides a larger hit area to focus the associated form field.
Placeholder text generally has poor contrast, and when its contrast is increased (and sometimes not even when it is increased) users may skip a field with placeholder text thinking the field has already been filled out.
Placeholder text may be cut off when the length of the text is longer than the input field width.
Placeholder text may become partially obscured or illegible when text is enlarged, e.g., with a screen magnifier.

**Recommendation** 
Form controls should always have a persistent, visible label (with some rare exceptions, such as when there are controls, like submit buttons, immediately adjacent to the control and which arguably act as an implicit label).
 
Only use the placeholder attribute for short hints intended to aid the user with data entry.
 Resources 
[Placeholder is not a label](https://html5accessibility.com/stuff/2021/02/14/re-upped-placeholder-the-piss-take-label/)
[HTML5 Accessibility Chops: the placeholder attribute](https://www.tpgi.com/html5-accessibility-chops-the-placeholder-attribute/)
[Accessible Floating Labels example by Scott Vinkle](https://codepen.io/svinkle/pen/NvazGb)
[Material Design Text Fields Are Badly Designed, by Adam Silver](https://www.smashingmagazine.com/2021/02/material-design-text-fields/)
[...]

PR #2523, opened 2024-01-29
feat: added visible labels for config panel threshold controls

Kausar-HM:
This PR addresses the accessibility issues #2512  and #2513 
1. Added visible label 'Operator' for arithmetic operator
2. Added visible label 'Value' for threshold input

### Verifying changes:
**Before (there were no visible labels):**
image

**After adding visible labels to arithmetic operator and threshold value input:**
**UPDATE**:
image

We have raised single PR for two tickets, as adding only one label made it look incomplete like this:
image

### DIFF
--- a/packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdComponent.tsx
+++ b/packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdComponent.tsx
@@ -10,6 +10,7 @@ import type { InputProps, SelectProps } from '@cloudscape-design/components';
 import {
   Box,
   Button,
+  FormField,
   Input,
   Select,
   SpaceBetween,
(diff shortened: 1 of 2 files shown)


Chandru-HM:
LGTM

packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdComponent.tsx
```diff
@@ -103,34 +104,45 @@ export const ThresholdComponent: FC<{
...
+                />
+              </FormField>
+              <Box margin={{ bottom: 'xs' }} variant='span'>
+                {defaultMessages.showAs}
+              </Box>
               <ColorPicker
                 color={color || DEFAULT_THRESHOLD_COLOR}
                 updateColor={onUpdateColor}
                 data-test-id='threshold-component-color-picker'
+                customStyle={{ marginBottom: `${awsui.spaceScaledXxs}` }}
```
gmuralidharreddy, line 135: Instead of importing all tokens, can you please import whatever you want?

packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdComponent.tsx
```diff
@@ -103,34 +104,45 @@ export const ThresholdComponent: FC<{
...
-                data-test-id='threshold-component-value-input'
-                invalid={!validValue}
-              />
-              {defaultMessages.showAs}
+              <Box margin={{ bottom: 'xs' }} variant='span'>
+                {defaultMessages.if}
+              </Box>
+              <FormField label='Operator'>
+                <Select
+                  expandToViewport={true}
```
gmuralidharreddy, line 112: If this prop is always true, then just use **expandToViewport**

gmuralidharreddy (changes requested):

gmuralidharreddy:
LGTM

tracy-french (changes requested):
I see you're attempting to follow the test patterns in this file, which unfortunately appear to be all wrong and are not using testing-library correctly. Using testing-library correctly, this test is possible. 

Here is an example of the correct usage:

```
import { render, screen } from '@testing-library/react';
import userEvent from '@testing-library/user-event';

it('should focus the input', async () => {
  const user = userEvent.setup();
  render(<TestThresholdComponent />);

  const label = screen.getByText('Operator');
  const input = screen.getByLabelText('Operator');

  expect(input).not.toHaveFocus();
  await user.click(label);
  expect(input).toHaveFocus();
});
```

Could you please update these tests?

packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdComponent.tsx
```diff
@@ -103,34 +104,46 @@ export const ThresholdComponent: FC<{
...
+              <Box variant='span' margin={{ top: 'l' }}>
+                <Button
+                  ariaLabel='delete threshold'
+                  iconName='remove'
+                  variant='icon'
+                  onClick={onDelete}
+                  data-test-id='threshold-component-delete-button'
+                />
+              </Box>
             </div>
```
tracy-french, line 147: There's four separate top margin's here. Is it possible to set the margin once? Maybe on a container element?

tracy-french: `<SpaceBetween />` has an alignItems prop. Could that be used to center the elements instead?

tracy-french (changes requested):

packages/dashboard/src/customization/propertiesSections/thresholdSettings/thresholdComponent.tsx
Kausar-HM: It doesn't work for `<SpaceBetween />`
Screenshot from 2024-02-12 12-39-09

I had tried to adjust the alignment using `alignItems: center` at it's parent `<div className='threshold-display-summary'>`, but it doesn't change the alignment because it is already centered.
If I apply `alignItems: end`, the text appears at the bottom, in that case as well need to apply margin somewhere.
image

I had to apply margin individually to those 4 elements to make text appear in same line.

Kausar-HM:
> I see you're attempting to follow the test patterns in this file, which unfortunately appear to be all wrong and are not using testing-library correctly. Using testing-library correctly, this test is possible.
> 
> Here is an example of the correct usage:
> 
> ```
> import { render, screen } from '@testing-library/react';
> import userEvent from '@testing-library/user-event';
> 
> it('should focus the input', async () => {
>   const user = userEvent.setup();
>   render(<TestThresholdComponent />);
> 
>   const label = screen.getByText('Operator');
>   const input = screen.getByLabelText('Operator');
> 
>   expect(input).not.toHaveFocus();
>   await user.click(label);
>   expect(input).toHaveFocus();
> });
> ```
> 
> Could you please update these tests?

Done, thanks for the help.

tracy-french (approved):

PR #2531, opened 2024-01-30
fix: preserve viewMode onSave

corteggiano:
## Overview
add a viewMode property to the `onSave` function so that parent component can handle preserving state of dashboard onSave

example in dashboard consumer
```
const [initialViewMode, setInitialViewMode] = useState('edit');

const onSave = async (dashboard: DashboardConfiguration, viewMode?: 'preview' | 'edit') => {
    viewMode && setInitialViewMode(viewMode);   // initialViewMode will be stored locally
    ... rest of save functionality
};

// setInitialViewMode is then passed to the dashboard
<Dashboard initialViewMode={setInitialViewMode} ...rest of dashboard config/>
```

https://github.com/awslabs/iot-app-kit/assets/28601414/58ff57f9-b854-49f3-ba40-3ddcb801f342

## Legal

### DIFF
--- a/packages/dashboard/src/components/dashboard/index.test.tsx
+++ b/packages/dashboard/src/components/dashboard/index.test.tsx
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import {
   createMockIoTEventsSDK,
   createMockSiteWiseSDK,
@@ -38,7 +38,7 @@ it('renders', function () {
 });
 
 it('renders in readonly initially', function () {
-  const { queryByText, queryByTestId } = render(
+  const { baseElement, queryByText, queryByTestId } = render(
     <Dashboard
       onSave={() => Promise.resolve()}
       dashboardConfiguration={{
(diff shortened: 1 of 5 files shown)


packages/dashboard/src/types.ts
```diff
@@ -33,8 +33,10 @@ export type DashboardClientConfiguration =
   | DashboardIotSiteWiseClients
   | DashboardClientCredentials;
 
+// OnSave has an optional viewMode value which can be used to persist the dashboard's viewMode after the save action
 export type DashboardSave = (
-  dashboardConfiguration: DashboardConfiguration
+  dashboardConfiguration: DashboardConfiguration,
+  viewMode?: 'preview' | 'edit'
```
corteggiano, line 39: heads up, this is a public API change!

packages/dashboard/src/components/dashboard/index.test.tsx
```diff
@@ -66,6 +66,47 @@ it('renders in readonly initially', function () {
   expect(queryByTestId(/time-selection/i)).toBeInTheDocument();
 });
 
+it('passes the correct viewMode to onSave', function () {
```
tracy-french, line 69: Could you add a test to test the initial view mode without setting `viewMode`?

packages/dashboard/src/components/dashboard/index.test.tsx
```diff
@@ -66,6 +66,47 @@ it('renders in readonly initially', function () {
...
+      dashboardConfiguration={savedConfig}
+      clientConfiguration={{
+        iotEventsClient: createMockIoTEventsSDK(),
+        iotSiteWiseClient:
+          createMockSiteWiseSDK() as unknown as IoTSiteWiseClient,
+        iotTwinMakerClient: {
+          send: jest.fn(),
+        } as unknown as IoTTwinMakerClient,
+      }}
+      initialViewMode='edit'
```
tracy-french, line 95: What about `viewMode`?

tracy-french (changes requested):

packages/dashboard/src/components/dashboard/index.test.tsx
corteggiano: since i didn't do my original solution due to overrequesting data then i thought initialViewMode was still a more fitting name

the parent component does not keep track of the current viewMode, it just knows what the viewMode was onSave

packages/dashboard/src/components/dashboard/index.test.tsx
corteggiano: added!

packages/dashboard/stories/dashboard/sitewise-dashboard.stories.tsx
```diff
@@ -61,6 +68,7 @@ export const Main: ComponentStory<typeof Dashboard> = () => {
     <Dashboard
       clientConfiguration={CLIENT_CONFIGURATION}
       onSave={onSave}
+      initialViewMode={initialViewMode}
```
mnischay, line 71: isnt this just viewMode going forward ?

mnischay (approved):

PR #2537, opened 2024-01-31
feat(Tiles3D): add Tiles3D AssetType and evaluate model type when adding a 3D model to the scene

hwandersman:
## Overview
While the scene composer/viewer supports rendering a 3D Tiles model, it does not support adding a model to the scene. This PR adds support for adding models with the 3D Tiles format.

If the requested file to add to the scene is exactly `tileset.json` then the assumption is that the file is in a folder with all dependent files for a 3D Tiles model. The scene hierarchy node name will default to the name of the parent folder.

Added error handling for adding scene assets that are not supported in the scene composer.

## Verifying Changes

Added unit tests and verified on local storybook. Added a simple mixer model converted to 3D Tiles as an asset to use in storybook.

1. Ability to add a 3D Tiles model to the scene
2. The name of the model is the name of the parent folder

We already support rendering a 3D Tiles models, adding tags to it, and deleting it from the scene. Verified this on storybook as well.

https://github.com/awslabs/iot-app-kit/assets/87385528/c08f9618-9d67-44e6-9f7d-ba0cfa1778da

Updated storybook locally to try to load an invalid file called `PALLET_JACK.invalid`:

Screenshot 2024-01-31 at 4 42 31 PM

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/components/toolbars/floatingToolbar/items/AddObjectMenu.tsx
```diff
@@ -197,7 +199,7 @@ export const AddObjectMenu = ({ canvasHeight, toolbarOrientation }: AddObjectMen
     });
   }, [selectedSceneNodeRef]);
 
-  const handleAddViewCamera = useCallback(() => {
```
hwandersman, line 200: Tested adding a camera view when first loading the scene, and it fails. Can still add a camera view after adding something else.
Need to remove this `useCallback` that I introduced with the feature flag removal.

sheilaXu:
In your video, why is the tiles model shown as having a child, but when you click expand button, nothing show up? should you just not showing the expand button instead?

packages/scene-composer/src/utils/sceneAssetUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+    const dir = path.dirname(filePath);
+    const modelName = dir === '.' ? name : dir.split('/').pop();
+    return {
+      modelName: modelName!,
+      modelType: AssetType.TILES_3D,
+    };
+  }
+
+  if (!ModelFileTypeList.includes(modelType)) {
+    const messageText = `ERROR: Model type ${modelType} not supported`;
```
sheilaXu, line 31: should we localize this string? also, since this modelType may not be a type of 3D model, the message should be something like `File with extension ${modelType} is not a supported model type`

hwandersman: Updated

hwandersman:
> In your video, why is the tiles model shown as having a child, but when you click expand button, nothing show up? should you just not showing the expand button instead?

Updated to disable showing submodels for Tiles3D nodes since it is not supported.

sheilaXu (approved):

PR #2539, opened 2024-02-01
feat(composer): save scene level data to scene root entity

sheilaXu:
## Overview
Save scene level data to scene root entity if exists and gated by a new feature flag DynamicSceneAlpha.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


hwandersman (approved):

packages/scene-composer/src/store/slices/SceneDocumentSlice.spec.ts
```diff
@@ -32,6 +33,10 @@ jest.mock('../../../src/utils/entityModelUtils/deleteNodeEntity');
 
 jest.mock('../../../src/utils/entityModelUtils/createNodeEntity');
 
+jest.mock('../../../src/utils/entityModelUtils/sceneUtils', () => {
```
mumanity, line 36: I know the pattern was already present in this file, but we're trying to move away from `requireActual` in favor of using mocks. Suggest avoiding this pattern in the future.

sheilaXu: what's the alternative to that if we only want to mock one function but need to use the real implementation for some other functions in the same file?

Issue #2379, opened 2023-12-12
[Widgets] widget level loading indicator


diehbria:
This task is to add a widget level loading indicator for the following widgets:
- xy-plot

To allow users to know if there is ongoing data being loaded into the widget.

## UX Notes [updated feb 26 2024]

For the XY-plot, add the loading indicator to the right of the timestamp container
![image](https://github.com/awslabs/iot-app-kit/assets/28601414/d0999c44-fc51-46e7-af0f-6e1a68e133c3)

## When the indicator should appear
The loading indicator should only appear  if both the following conditions are met:
- the "primary" loading indicator is not being shown (the one displayed in the center of the "chart area" )
- if there has been 3 or more seconds of continuous loading of data on the widget. This can be determined by evaluating the `isRefreshed` and `isLoading` properties on the datastream(s) being presented on the widget.

This prevents loading spinners from flashing in and out constantly, which would be distracting to users

```[tasklist]
### Acceptance criteria
- [ ] XY-plot has loading indicator added
- [ ] loading indicators do not appear until one or more datastreams have been in a "loading" state for 3 or more seconds
- [ ] tests added
```

PR #2540, opened 2024-02-01
feat: add a widget level loading indicator for x-y plot

Chandru-HM:
## Overview
Issue: https://github.com/awslabs/iot-app-kit/issues/2379

Verification:
-> Added widget level loading indicator on refresh if it continues for 3 sec or more.
-> Widget level indicator should only show when the "primary" loading indicator is not being shown (the one displayed in the
center of the "chart area" )
-> Changed "isRefreshing" logic in core package and made respective changes.

**Note:** can verify by varying delay time and slow network

## Verifying Changes
**Note:**  tested with 10ms delay and slow network

https://github.com/awslabs/iot-app-kit/assets/142866907/828b8329-b2ab-4992-87e1-9f7dd1700872

## Legal


corteggiano:
i am not seeing the spinner appear when data takes longer to load
image
image

Chandru-HM:
tested by changing throttling and reducing the delay time(IS_REFRESHING_DELAY) in constants file. working as below

https://github.com/awslabs/iot-app-kit/assets/142866907/615e0edc-cb44-435b-8aef-d803a4904a91

corteggiano:
as mentioned before, `useVisualizedDataStreams` should not be exported from react-components
making calls to this hook in dashboard could cause a bunch of duplicate network calls and general performance issues

i will talk to design and ask if we can keep the loading state within react-components so all the changes can be contained within that package

corteggiano:
i've talked with Navita and we have agreed to move the loader into the baseChart component! 

the best way to do this will be by adding a custom toolbox option
https://echarts.apache.org/en/option.html#toolbox.feature

the icon should not do anything on click and should not be visible unless state is loading

Chandru-HM:
Working on moving loader into the baseChart component, so marked this PR as Draft.

Chandru-HM:
@corteggiano changes are updated. icon is added within the echart. no click action and no tooltip. but spinner is not rotating cause echart taking it as image. please review.
to Test: verify by varying delay time and with slow network

Chandru-HM:
@corteggiano changes are updated, and PR is ready for review. Please take look.

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -246,9 +262,22 @@ const BaseChart = ({
             />
           </HotKeys>
         </Resizable>
-        <div className='base-chart-timestamp' style={timestampStyle}>
-          <span>{timestampStart}</span>
-          <span>{timestampEnd}</span>
+        <div className='base-chart-timestamp-container' style={timestampStyle}>
```
corteggiano, line 265: could you move this whole timestamp container section into its own file

corteggiano, line 265: could you move this whole timestamp container section into its own file

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -246,9 +262,22 @@ const BaseChart = ({
...
+          {isPropertiesRefreshing && (
```
corteggiano, line 266: noticed that the spinner shifts over the whole timestamp
ideally there would be padding when the spinner is not there so that it does not shift any elements if it appears
 
Screenshot 2024-02-29 at 10 29 48
Screenshot 2024-02-29 at 10 30 27

corteggiano (changes requested):

packages/react-components/src/components/chart/baseChart.tsx
Chandru-HM: updated

packages/react-components/src/components/chart/baseChart.tsx
Chandru-HM: updated.

corteggiano (approved):

Issue #2511, opened 2024-01-25
[Accessibility][ComponentLibrary] Content displayed is not dismissible


jmbuss:
**Describe the bug**

**Standard:** 1.4.13 Content on Hover or Focus (AA) 
**Affected Groups:** Keyboard Users

**Notes and Advice** 
Content that is displayed when a pointer hovers over the following elements is not dismissible without moving the pointer or focus (typically by pressing ESC)
 
Users with mobility impairments, such as hand tremors, may find it difficult to operate content when the displayed content closes too readily. Similarly, such users may unintentionally trigger the content, or may have no use for it, and need a method to dismiss it that does not require moving their current hover or focus position.

**Recommendation** 
Ensure all displayed content keeps itself visible, which can often be done using CSS focus-within or scripting that adds hover and focus handlers to the content.
 
To allow content to be dismissed, adding support for the ESC key is the most common method and one that many users expect.

![image](https://github.com/awslabs/iot-app-kit/assets/107281089/a9f58455-37f2-408b-bc26-736dcfcb1880)

The tooltips for the component library should be dismissable.

PR #2541, opened 2024-02-01
feat: made widget tooltip dismissable by pressing escape key

Kausar-HM:
In this PR we have added the ability to dismiss the widget tooltip by pressing escape key as per this ticket #2511 

Verifying changes:
[widget-tooltip-dismiss-esc.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/45e6e0c9-84dd-4f78-a29c-d7fb8ea3892d)

Note: when we click on "full screen mode" button on storybook area, first have to click inside the dashboard area, then when we hover over the widget list, tooltip content is dismiss-able by pressing the escape key.


tracy-french (approved):
Looks good! Nice work and thank you, @Kausar-HM!

Issue #2547, opened 2024-02-01
[Accessibility][ConfigPanelText] Color alone conveys information without additional shapes, patterns, or images


jmbuss:
Issue 16.5 [Color alone conveys information without additional shapes, patterns, or images](https://www.tpgarc.com/Analytics/Dashboard/EngagementComponentPageAssertions/1822782/Engagement/1759/ScanLog/106158/Assertion/754/AccessibilityStandard/WCAG21) 
 
 
Assertion ID: 200290
 
Target Count: 3
 
Standard: 1.4.1 Use of Color (A)
 
Affected Groups: Cognitively Impaired, Low Vision
 
Severity: Medium
 
Priority: 3
 
 
**Notes and Advice** 
The only visual means of conveying information for distinguishing visual elements when focused is with color, this is the case for the size control. This may not be perceivable to color-blind and low-vision users, nor be conveyed to users of assistive technology.

**Recommendation** 
Complement the use of color with additional visual cues, such as shapes or patterns, or text, to convey information.
 
In the case of focus indicators we recommend a clearly visible outline or border. Alternately, using a color change alone as a focus indicator can pass [1.4.1 Use of Color](https://www.w3.org/TR/WCAG/#use-of-color), provided that the difference between the focused state and the unfocused state has a contrast ratio of 3:1 or greater. For example, if the focused state of a button is indicated by a change in background color, then the difference between the focused and unfocused backgrounds must have a contrast ratio of 3:1 or greater. High-contrast color inversion (such as black to white) is usually sufficient.
 Resources 
[Well Color Us Surprised—This SC Can Be a Tricky Customer](https://www.tpgi.com/well-color-us-surprised-this-sc-can-be-a-tricky-customer/)
[Understanding Success Criterion 1.4.1: Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)

<img width="497" alt="image" src="https://github.com/awslabs/iot-app-kit/assets/107281089/ccc22bcf-2c4e-421a-88f3-25462d98724b">

<img width="501" alt="image" src="https://github.com/awslabs/iot-app-kit/assets/107281089/62796a11-9d2e-47dc-b43e-c7885833bf11">

Color contrast ratio must be 3:1 for the focus state of these buttons.

PR #2548, opened 2024-02-02
feat: color contrast ratio on tab focus of buttons in config panel text style

Kausar-HM:
Changes to resolve the accessibility issue #2547 
When user focuses on text color style buttons using tab key, the outline color will help user to identify the focused state.

Verifying changes:

Before the outline color contrast was low for those buttons:
[faint-tab-focus-configPanelText-buttons.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/1a7cfdef-86f0-44d1-b6ca-895886f5f0f7)

Now with this PR, color contrast is changed:
[blue-outline-tab-focus.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/4cd5ad11-6782-4383-ac77-a8485feb527d)

image

### DIFF
--- a/packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
+++ b/packages/dashboard/src/customization/propertiesSections/textSettings/text.tsx
@@ -1,5 +1,5 @@
 import type { FC, MouseEventHandler, ReactNode } from 'react';
-import React from 'react';
+import React, { useState } from 'react';
 import type {
   NonCancelableCustomEvent,
   SelectProps,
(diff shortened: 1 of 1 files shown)


corteggiano (approved):

jmbuss:
> #2547

The ticket is about color contrast between hover states. This is because we are only using color to indicate an action. The focus / unfocused states must have a 3:1 contrast ration. For instance, this would work - 

image

these come from - https://refresh.cloudscape.aws.dev/foundation/visual-foundation/colors/

Issue #2551, opened 2024-02-05
[ResourceExplorer] Add "arrow" indicator to resource explorer asset browser


diehbria:
This task is to reduce the confusion users have in knowing that they can click on the asset name to drill down the SiteWise asset heirachy. 

## UX Guidance
- add the cloudscape "arrow" icon to the left of the asset name, as shown in the screenshot below
- Add a tooltip on hover of the asset name that specifies "view children assets of $ASSET_NAME"

![Screenshot 2024-02-05 at 3 27 50 PM (1)](https://github.com/awslabs/iot-app-kit/assets/77755322/963d5c7a-0faa-46f8-8c0a-478edf5306f4)

```[tasklist]
### Acceptance criteria
- [ ] Cloudscape icon is added to asset name per UX guidance
- [ ] Tooltip is added per UX guidance
```

PR #2556, opened 2024-02-06
feat: add arrow to asset name #2551

gmuralidharreddy:
## Overview

This PR is for #2551 
Added right arrow icon as per UX and tooltip on hover of asset name.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/139960352/3638e8cf-c295-4abb-bfe8-f6adb023072d

## Legal


Chandru-HM:
LGTM

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/modeledDataStreamQueryEditor/assetExplorer/assetTable/assetTableNameLink.tsx
```diff
@@ -20,7 +22,13 @@ export function AssetTableNameLink({
         updateParentAssetId(assetId);
       }}
     >
-      {assetName}
+      <span
+        data-testid='asset-name-link'
+        title={`View children assets of ${assetName}`}
```
jmbuss, line 27: Is this the type of tool tip UX wanted? We do have a tool tip component that we use elsewhere in dashboard, like the one used in the component library.

jmbuss (commented):
Looks good, one comment about the UX. I think the tool tip we use elsewhere in dashboard would be a more consistent experience. Can we confirm with UX?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/modeledDataStreamQueryEditor/assetExplorer/assetTable/assetTableNameLink.tsx
gmuralidharreddy: @nadeswal can you please confirm, Is it fine to go with the default tooltip as shown in the above video? Or can we add the tooltip like what we used for the component library?

gmuralidharreddy: @jmbuss I tried with a` <Tooltip /> `component but it was not working as expected, outside layouts are overlaying the tooltip content like in the below screenshot,
image

I tried to use an external tooltip library(https://www.npmjs.com/package/react-tooltip?activeTab=readme) and we are able to show a tooltip like the one below screenshot,
image

Since we are not using the sibling nodes to the root div(React Portals) for tooltips it causing the Z-index issues.

jmbuss: I see, yes it looks like a limitation of the Tooltip component. We will need to update that component to be able to handle cases such as this. We can move forward with the title attribute now to unblock and I will create a ticket to fix the Tooltip component.

jmbuss (approved):

PR #2558, opened 2024-02-06
feat(react-components): calculate min/max and store value in store for chart to consume

ssjagad:
## Overview
Manage data for min/max on chart by using the echarts extension to go through points in each series find the min/max, and store it in the chartStore by id. 

## Verifying Changes

Screenshot 2024-02-06 at 12 47 05 PM

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/store/dataStreamMinMaxStore.ts
+++ b/packages/react-components/src/components/chart/store/dataStreamMinMaxStore.ts
@@ -1,0 +1,30 @@
+import { StateCreator } from 'zustand';
+
+export type MinMaxMap = { [key in string]: number | undefined };
+
+export interface MinMaxData {
+  dataStreamMaxes: MinMaxMap;
+  dataStreamMins: MinMaxMap;
+}
+
+export interface MinMaxState extends MinMaxData {
+  setMaxes: (datastreamMaxes: MinMaxMap) => void;
+  setMins: (datastreamMins: MinMaxMap) => void;
+}
+
+export const createMinMaxSlice: StateCreator<MinMaxState> = (set) => ({
+  dataStreamMaxes: {},
+  dataStreamMins: {},
+  setMaxes: (datastreamMaxes) =>
+    set(() => ({
+      dataStreamMaxes: {
+        ...datastreamMaxes,
+      },
+    })),
+  setMins: (datastreamMins) =>
+    set(() => ({
+      dataStreamMins: {
+        ...datastreamMins,
+      },
+    })),
+});
--- a/packages/react-components/src/echarts/extensions/minMaxDataStreamValues/minMaxDataStreamSync.ts
+++ b/packages/react-components/src/echarts/extensions/minMaxDataStreamValues/minMaxDataStreamSync.ts
@@ -1,0 +1,32 @@
+import SeriesModel from 'echarts/types/src/model/Series';
+import { SeriesOption, DefaultStatesMixin } from 'echarts/types/src/util/types';
+
+export const findMinMax = (
+  series: SeriesModel<SeriesOption<unknown, DefaultStatesMixin>>,
+  start: number,
+  end: number
+) => {
+  let max: number | undefined = undefined;
+  let min: number | undefined = undefined;
+  series.getData().each((dims) => {
+    const dataPoint = series.getData().getValues(dims) as number[];
+    if (contains(start, end, dataPoint[0])) {
+      if (max === undefined || dataPoint[1] > max) {
+        max = dataPoint[1];
+      }
+      if (min === undefined || dataPoint[1] < min) {
+        min = dataPoint[1];
+      }
+    }
+  });
+  return { max, min };
+};
+
+/* We are checking if the data is within the x-axis of the viewport 
+  Currently, the dataZoom also filters the points so that the series 
+  only contains the points seen in the viewport, but this decouples
+  the functinality in case of changes to useDataZoom in the future.
+*/
+const contains = (start: number, end: number, value: number) => {
+  return value > start && value < end;
+};
--- a/packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
+++ b/packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
@@ -1,0 +1,69 @@
+import { EChartsExtensionInstallRegisters } from 'echarts/types/src/extension';
+import { findMinMax } from './minMaxDataStreamSync';
+import useDataStore from '../../../store';
+import throttle from 'lodash.throttle';
+import Grid from 'echarts/types/src/coord/cartesian/Grid';
+import { LifecycleEvents } from 'echarts/types/src/core/lifecycle';
+
+const THROTTLE_RATE = 2000;
+
+// Echarts core use type does not map correctly to the echarts extension type so exporting as any
+// eslint-disable-next-line
+export const dataStreamMinMaxSyncExtension: any = (registers: EChartsExtensionInstallRegisters) => {
+  const dataStreamMinMaxCallback = (
+    ...args: LifecycleEvents['series:afterupdate']
+  ) => {
+    const [ecModel, api, params] = args;
+    const appKitChartId = ecModel.option.appKitChartId as string; // widget-id
+    const grid = api
+      ?.getCoordinateSystems()
+      .find((a) => a.model?.type === 'grid') as Grid;
+
+    if (!grid) return; // grid not found, return early
+
+    const axis = grid?.getAxis('x'); //get X axis object from echarts
+    if (!axis) return; // axis not found, return early
+
+    const startDateXCoordinateValue = axis?.coordToData(
+      axis.toLocalCoord(axis?.getExtent()[0])
+    ); // start date coordinate value for x as number
+    const endDateXCoordinateValue = axis?.coordToData(
+      axis.toLocalCoord(axis?.getExtent()[1])
+    ); // end date coordinate value for x as number
+
+    if (
+      !params.updatedSeries ||
+      startDateXCoordinateValue === undefined ||
+      endDateXCoordinateValue === undefined ||
+      startDateXCoordinateValue > endDateXCoordinateValue
+    )
+      return;
+
+    const minValues: Record<string, number | undefined> = {}; // each value will be {dsId: min}
+    const maxValues: Record<string, number | undefined> = {}; // each value will be {dsId: max}
+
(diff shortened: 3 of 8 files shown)


packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// Echarts core use type does not map correctly to the echarts extension type so exporting as any
+// eslint-disable-next-line
+export const dataStreamMinMaxSyncExtension: any = (registers: EChartsExtensionInstallRegisters) => {
+  registers.registerUpdateLifecycle(
+    'series:afterupdate',
+    (_ecModel, _api, params) => {
+      if (!params.updatedSeries) return;
+
+      params.updatedSeries.forEach((series) => {
```
jmbuss, line 13: have you tested adding / removing datastreams? Sometimes echarts can have null series values.

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/minMaxDataStreamSync.ts
```diff
@@ -0,0 +1,57 @@
...
+  model: LineSeriesModel,
+  max: number,
+  min: number
+) => {
+  const appKitChartId = model.ecModel.option.appKitChartId as string;
+  const { option, coordinateSystem } = model;
+  const id = `${option.id}`; // echart option id can also be a number
+
+  // Cannot determine yAxis unless the coordinateSystem is cartesion2d
+  if (
```
jmbuss, line 34: is this required?

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,20 @@
...
+        const { max, min } = findMinMax(series);
+        if (max == null || min == null) return;
+        handleMinMaxSync(series as unknown as LineSeriesModel, max, min);
```
jmbuss, line 16: can we write this as a function which does not depend on echarts values? seems like it just needs the id and the values?

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/minMaxDataStreamSync.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const storeState = useDataStore.getState();
+  const stores = storeState.chartStores;
+  const store = stores[appKitChartId];
+
+  if (!store) return;
+
+  const state = store.getState();
+
+  if (!isEqual(max, state.dataStreamMaxes[id])) {
```
jmbuss, line 49: is this required?

packages/react-components/src/components/chart/store/dataStreamMinMaxStore.ts
```diff
@@ -0,0 +1,32 @@
...
+  setMax: (datastreamId: string, maxValue: number) => void;
+  setMin: (datastreamId: string, minValue: number) => void;
+}
+
+export const createMinMaxSlice: StateCreator<MinMaxState> = (set) => ({
+  dataStreamMaxes: {},
+  dataStreamMins: {},
+  setMax: (datastreamId, value) =>
+    set((state) => ({
+      dataStreamMaxes: {
```
jmbuss, line 20: what happens when we remove datastreams? do they stick around in the store?

jmbuss: It would be possible to aggregate all of the values at once from the callback, and then overwrite all the mins/ maxes so that we never have a scenario where we leave orphaned datastream mins/maxes

jmbuss (commented):
Nice! Looks good, for testing I think we can write some e2e tests. It will be difficult to determine exact values without refactoring the mock service worker to be deterministic, but at least we can check that the values are being populated in the legend table. If we aren't adding that code yet, let's not install the plugin.

packages/react-components/src/components/chart/store/dataStreamMinMaxStore.ts
ssjagad: thats genius. Done!

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
ssjagad: Yes! I have a case where it is a null value it should store undefined, and then display "No max found" or something in the legend or "-" to indicate its not in the viewport

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,37 @@
...
+// Echarts core use type does not map correctly to the echarts extension type so exporting as any
+// eslint-disable-next-line
+export const dataStreamMinMaxSyncExtension: any = (registers: EChartsExtensionInstallRegisters) => {
+  registers.registerUpdateLifecycle(
+    'series:afterupdate',
+    (_ecModel, _api, params) => {
+      if (!params.updatedSeries) return;
+
+      params.updatedSeries.forEach((series) => {
+        const appKitChartId = series.ecModel.option.appKitChartId as string; // widget-id
```
jmbuss, line 14: You can access this once outside the loop on the _ecModel variable in the callback function.

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        const minValues: Record<string, number | undefined> = {}; // each value will be {dsID: min}
+        const maxValues: Record<string, number | undefined> = {}; // each value will be {dsID: max}
+
+        params.updatedSeries?.forEach((series) => {
```
jmbuss, line 19: why are we doing a nested loop of updatedSeries? can't we just do one loop?

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/minMaxDataStreamSync.ts
```diff
@@ -0,0 +1,19 @@
+import SeriesModel from 'echarts/types/src/model/Series';
+import { SeriesOption, DefaultStatesMixin } from 'echarts/types/src/util/types';
+
+export const findMinMax = (
+  series: SeriesModel<SeriesOption<unknown, DefaultStatesMixin>>
+) => {
+  let max: number | undefined = undefined;
+  let min: number | undefined = undefined;
+  series.getData().each((dims) => {
+    const dataPoint = series.getData().getValues(dims).at(1) as number; // get y value
```
jmbuss, line 10: do we not need to check whether or not the point is contained within the coordinate system? Have you tested all of the different chart types? If we don't need to do this because of our usage with dataZoom, we should note that this function only works because we are filtering data points via dataZoom

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,37 @@
+import { EChartsExtensionInstallRegisters } from 'echarts/types/src/extension';
+import { findMinMax } from './minMaxDataStreamSync';
+import useDataStore from '../../../store';
+
+// Echarts core use type does not map correctly to the echarts extension type so exporting as any
+// eslint-disable-next-line
+export const dataStreamMinMaxSyncExtension: any = (registers: EChartsExtensionInstallRegisters) => {
+  registers.registerUpdateLifecycle(
+    'series:afterupdate',
+    (_ecModel, _api, params) => {
```
jmbuss, line 10: How often is this function called? should we think about throttling it since it does iterator all of the data points in every series?

jmbuss (commented):
Great updates! I had a few more improvements / items to consider since this is using the experimental echarts api

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,47 @@
...
+      (_ecModel, _api, params) => {
+        const appKitChartId = _ecModel.option.appKitChartId as string; // widget-id
+        const grid = _api
+          ?.getCoordinateSystems()
+          .find((a) => a.model?.type === 'grid') as Grid;
+        const axis = grid?.getAxis('x');
+        const start = axis?.coordToData(axis.toLocalCoord(0));
+        const end = axis?.coordToData(axis.toLocalCoord(axis?.getExtent()[1]));
+
+        if (!params.updatedSeries || !start || !end) return;
```
diehbria, line 23: what if start or end is zero?

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,54 @@
...
+        const maxValues: Record<string, number | undefined> = {}; // each value will be {dsID: max}
+
+        params.updatedSeries?.forEach((series) => {
+          const id = `${series.option.id}`; // datastream-id
+          const { max, min } = findMinMax(series, start, end);
+          minValues[id] = min;
+          maxValues[id] = max;
+        });
+
+        console.log(maxValues);
```
jmbuss, line 40: don't forget to remove this one!

jmbuss (commented):
Adding some more comments!

packages/react-components/src/echarts/extensions/minMaxDataStreamValues/updateDataStreamMinMax.ts
```diff
@@ -0,0 +1,54 @@
...
+        const grid = _api
+          ?.getCoordinateSystems()
+          .find((a) => a.model?.type === 'grid') as Grid;
+        const axis = grid?.getAxis('x');
+        const start = axis?.coordToData(axis.toLocalCoord(0));
+        const end = axis?.coordToData(axis.toLocalCoord(axis?.getExtent()[1]));
+
+        if (
+          !params.updatedSeries ||
+          (!start && start !== 0) ||
```
jmbuss, line 25: what is this guard for? It doesn't make sense for start or end to be 0, but why do we need to check against it?

jmbuss (approved):

PR #2568, opened 2024-02-09
feat(scene-composer): enable accelerated raycasting for 3D Tiles

hwandersman:
## Overview
The raycasting for 3D Tiles has not been optimized like other GLTF models. This PR enables three-mesh-bvh `acceleratedRaycasting` to improve the calculation of intersections for raycasting in the scene.

The [3D Tiles renderer](https://github.com/NASA-AMMOS/3DTilesRendererJS) that we use calls a couple raycast methods `raycastTraverse()` and `raycastTraverseFirstHit()`. These methods use the helper `intersectTileScene()` which forces the built-in `raycast` method of the Object3D / Mesh class. This helper must be overwritten to use the object instance's `raycast` method, which we overwrite in `TilesModelComponent`.

I copied over the version of the `raycastTraverse.js` script that AppKit uses, version `0.3.16`.
* [Reference](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/fada8f7394b7a98c5ab259afb793d3353b13e555/src/three/raycastTraverse.js)

## Verifying Changes

Can see the clear difference between the scene composer on prod Console vs with this change on storybook:

https://github.com/awslabs/iot-app-kit/assets/87385528/b7dd291e-e326-4776-916c-34f13bb55a74

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/three/tiles3d/raycastTraverse.js
+++ b/packages/scene-composer/src/three/tiles3d/raycastTraverse.js
@@ -1,0 +1,237 @@
+import { Matrix4, Sphere, Ray, Vector3 } from 'three';
+
+const _sphere = new Sphere();
+const _mat = new Matrix4();
+const _vec = new Vector3();
+const _vec2 = new Vector3();
+const _ray = new Ray();
+
+const _hitArray = [];
+
+function distanceSort( a, b ) {
+
+	return a.distance - b.distance;
+
+}
+
+function intersectTileScene( scene, raycaster, intersects ) {
+
+  // + Amazon --------
+  // Call the object's overridden raycast method
+	scene.traverse(c => c.raycast.call(c, raycaster, intersects));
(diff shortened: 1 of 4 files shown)


packages/scene-composer/src/three/tiles3d/raycastTraverse.js
```diff
@@ -0,0 +1,238 @@
+import { Matrix4, Sphere, Ray, Vector3 } from 'three';
+import { acceleratedRaycast, computeBoundsTree, disposeBoundsTree } from 'three-mesh-bvh';
```
haweston, line 2: It doesn't look like these imports are used in this file anymore.

haweston (approved):
Just remove unneeded import

PR #2571, opened 2024-02-12
feat: adding docker setup for scene-composer ui tests

mitchlee-amzn:
## Overview
 feat: adding docker setup for scene-composer ui tests
    
    - Moves all snapshots to use git lfs (replaces with pointers, you will need git lfs to checkout from this point on)
    - Updates snapshots for scene-composer ui tests, now just chromium-linux for now (you can technically run darwin if you like, but our master snapshots are linux based from now on).
    - Add new Github workflow anyone can add to a pull request to trigger a reliability check. Adding the "test:reliability" label to a pull request will now force all UI tests to run 5 times, useful for checking for flaky tests when new tests are being added in a PR

Basically, now when we run UI tests (currently just for scene-composer, other packages will be separate CR), it will run them in a docker container instead of on the user's local machine by default. This way, snapshots will be reliably compareable on both local development, and the CI server.

More importantly, when we generate new shapshots, they will pass on the CI server when we checkin, rather than requiring us to download the CI specific versions for our tests.

Finally, I also added a github workflow that authors/reviewers can choose to trigger on any Pull Request, that triggers a test reliability process. This essentially will run all UI tests 5 times each to protect against flaky tests being checked in.

## Verifying Changes

Changes should be verified automatically by the test-ui workflows on this PR, and the test:reliability label.

## Legal


package.json
```diff
@@ -27,14 +27,16 @@
...
     "test:stylelint": "stylelint '**/*.css' --max-warnings 0",
     "test:git": "git diff --exit-code",
     "release": "npm run build",
     "pack": "turbo run pack",
     "versionup:patch": "turbo run version --no-git-tag-version patch",
     "versionup:minor": "turbo run version --no-git-tag-version minor",
     "versionup:major": "turbo run version --no-git-tag-version major",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "test:ci": "if test \"$NODE_ENV\" = \"production\"; then echo \"Yes!\"; fi"
```
mitchlee-amzn, line 39: nit: This is test code that needs to be removed.

packages/scene-composer/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+services:
+  # storybook:
+  #   image: node:18
+  #   build: .
+  #   working_dir: '/iot-app-kit/packages/scene-composer'
+  #   volumes:
+  #     - ../../:/iot-app-kit
+  #   ports:
+  #     - 7006:7006/tcp
+  #   entrypoint: 'npm start'
```
mitchlee-amzn, line 12: Dead code needs cleanup.

packages/scene-composer/e2e/tests/local-scene/scene-composer--local-scene.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { test, expect } from '@playwright/test';
+
+import { PlaywrightHelper } from '../utils/sceneHelpers';
+
+const localScene = '/iframe.html?args=&id=developer-scene-composer--local-scene';
```
mitchlee-amzn, line 5: This code is not part of this CR, need proper rebase with @mumanity's changes so these don't show up in diff

tracy-french (approved):

PR #2582, opened 2024-02-13
feat(react-components): add max column to legend

ssjagad:
## Overview
Adding ability to toggle max column from config panel. The max value is pulled from the data store and displayed in a dataStreamMax cell.

## Verifying Changes
https://github.com/awslabs/iot-app-kit/assets/145582655/1fc6534b-5b20-4422-abbf-6334a8f99bd0

## Legal

### DIFF
--- a/packages/dashboard/e2e/tests/constants.ts
+++ b/packages/dashboard/e2e/tests/constants.ts
@@ -21,3 +21,6 @@ export const REMOVE_TREND_CURSOR = 'Delete trend cursor';
 export const TREND_CURSOR_TABLE_HEADER =
   '.base-chart-legend-tc-header-container';
 export const TREND_CURSOR_TABLE_CELL = '.trend-cursor-value';
+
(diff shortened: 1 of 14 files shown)


packages/dashboard/e2e/tests/constants.ts
```diff
@@ -21,3 +21,6 @@ export const REMOVE_TREND_CURSOR = 'Delete trend cursor';
 export const TREND_CURSOR_TABLE_HEADER =
   '.base-chart-legend-tc-header-container';
 export const TREND_CURSOR_TABLE_CELL = '.trend-cursor-value';
+
+export const MAX_VALUE_TABLE_HEADER = '.base-chart-legend-max-header-container';
```
corteggiano, line 25: could these be `data-testid` instead of classes? that way we guarantee they stick around since classes can change names

corteggiano (approved):

Issue #2439, opened 2024-01-03
[Widget] panel UX: On-select/hover widget-level actions


diehbria:
This task is to add a on-select and on-hover action state for the widget. This will be applicable for all the widgets.

This task is to update the "widget panel" and how it exposes the delete action.
UPDATE: this should include download button as well (01/19/24)

Currently, the "delete action" is always presents on some widgets while in edit mode.

This task is to change the panel UX so that the "delete" button only displays when a widget is selected or is hovered on while in edit mode.

(Note that the "download" icon won't be part of this implementation)
UPDATE : download button is now available (01/19/24)

UX guidelines - 
![Screenshot 2024-01-03 at 12 47 44 PM](https://github.com/awslabs/iot-app-kit/assets/77755322/6cda634a-d371-4106-ad20-5ba34176c5f5)

- Add a 5px rounded corner radius - 58x24px rectangle (this is as per the designs - feel free to play with the size to match the - design and the widget size which will be relative to the user making it small or large)
- Add 10px padding on both side and in the middle of the icon
- Icon size 14px (buttons should use the cloudscape inline icon button)

```[tasklist]
### Acceptance criteria
- [ ] Update "panel UX" to expose delete on hover/select per UX guidance
- [ ] New panel UX is accessibile
- [ ] Utilize new panel UX for xy-plot
- [ ] Utilize new panel UX for KPI
- [ ] Utilize new panel UX for Status
- [ ] Utilize new panel UX for table
- [ ] Utilize new panel UX for bar chart
- [ ] Utilize new panel UX for status timeline
- [ ] Get UX signoff from UX (Navita Deswal)

```

Kausar-HM:
@mnischay 
We need to update the ticket description to be inclusive of download button as well since now the download button is available in latest rc, it looks like this
![image](https://github.com/awslabs/iot-app-kit/assets/154328164/51abf8c2-8cbe-4470-8932-692a8c49f451)

We are already working on updating the buttons UX

gmuralidharreddy:
@nadeswal
Can you please provide the UX for download icon placement in the preview mode?

mnischay:
@Kausar-HM : @corteggiano  mentioned that `the download button should be disabled when a table kpi or status  is not in live mode`

nadeswal:
### Story continuation --
**View mode:**
The view mode will support the hover state which will enable the user to download the widget.

Refer screenshot -
[...]

PR #2589, opened 2024-02-14
feat: widget tool box on mouse hover and selection state

gmuralidharreddy:
## Overview
This PR is for the ticket #2439 

- Removed widget header from widgetTile component.
- Added functionality to display remove and download buttons on hover and widget selected state.
- Moved widget title for line-scatter to echart component and for bar and timeline to inside widget tile and truncated the title when widget was resized.
- Displayed only the download button on hover in live mode for line-scatter(xy-plot), bar-chart, and status-timeline widgets and hidden widget actions toolbox for the rest of all widgets.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/139960352/b9a0b94f-b4cd-4584-84c4-1f9b3f5fe6cb

## Legal

### DIFF
--- a/packages/dashboard/src/components/widgets/widgetActions.css
+++ b/packages/dashboard/src/components/widgets/widgetActions.css
@@ -1,0 +1,7 @@
+.widget-actions-container {
+  display: flex;
+  align-items: center;
+  position: absolute;
+  top: -20px;
+  z-index: 999; /* Widget actions toolbox should be above the widget container */
+}
--- a/packages/dashboard/e2e/tests/dashboard/DashboardPage.ts
+++ b/packages/dashboard/e2e/tests/dashboard/DashboardPage.ts
@@ -39,6 +39,7 @@ export class DashboardPage {
     this.gridArea = page.locator('#container');
     this.confirmModalDeleteButton = this.page.getByRole('button', {
       name: 'Delete',
+      exact: true,
     });
     this.contextMenuDeleteButton = this.gridArea.getByText('Delete');
   }
--- a/packages/dashboard/src/components/csvDownloadButton/index.tsx
+++ b/packages/dashboard/src/components/csvDownloadButton/index.tsx
@@ -18,9 +18,14 @@ import {
 } from './constants';
 import { convertViewportToHistoricalViewport } from '../util/dateTimeUtil';
 
-const canOnlyDownloadLiveMode: readonly string[] = ['table', 'kpi', 'status'];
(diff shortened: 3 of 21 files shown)


dpportet:
The PR says it adds remove and download buttons but in the verification video I just see the remove button and no download button, is that an old video?

packages/dashboard/src/components/widgets/widgetActions.css
```diff
@@ -0,0 +1,7 @@
+.widget-actions-container {
+  display: flex;
+  align-items: center;
+  position: absolute;
+  top: -20px;
+  z-index: 999;
```
dpportet, line 6: Is there a reason to chose `999` here? Maybe we can add a comment to explain which elements it overlaps with to need this z-index value.

gmuralidharreddy:
> The PR says it adds remove and download buttons but in the verification video I just see the remove button and no download button, is that an old video?

Actually, I recorded a full-length video which covers all test scenarios, but GitHub have restrictions on uploading a 100mb file, so trim the initial part of the video and uploaded it, now I removed and uploaded a new video which covers almost all scenarios with quick mode.

dpportet (changes requested):
Overall looks good, I think we should add the comment to the z-index since we have a bunch of z-index values throughout the codebase.

dpportet:
> > The PR says it adds remove and download buttons but in the verification video I just see the remove button and no download button, is that an old video?
> 
> Actually, I recorded a full-length video which covers all test scenarios, but GitHub have restrictions on uploading a 100mb file, so trim the initial part of the video and uploaded it, now I removed and uploaded a new video which covers almost all scenarios with quick mode.

I didn't realize at first that we only wanted it in live mode, that makes sense thanks.

packages/dashboard/src/components/widgets/widgetActions.css
gmuralidharreddy: Screenshot 2024-02-17 at 3 44 38 AM

Added comment for the z-index, if we give z-index value like '99' then it shows like the above screenshot

dpportet (approved):

dpportet:
@gmuralidharreddy In doing some testing we noticed some differences from the UX - too much vertical padding and the border radius is too small. Also, the hover icons can hide the chart gesture icons which we don't want. Could you follow up with these fixes to match the expected UX? Included screenshot of difference (UX guidance is the right one).

Screenshot 2024-02-22 at 8 58 04 AM

PR #2600, opened 2024-02-16
feat(dashboard): make component library keyboard accessible

jmbuss:
## Overview
Updates to the component library for accessibility. 

1. Change markup to a button so that the element is focusable via the keyboard instead of directly setting the tabIndex.
2. Add click and keyboard handlers so that widgets can be added with a single action.
3. Add draggable attribute to indicate item functionality correctly

## Legal


chejimmy (approved):

chejimmy:
LGTM! Awesome idea to help with accessibility 🙏

PR #2609, opened 2024-02-21
fix: update public API for status and KPI

corteggiano:
## Overview
- some minor style fixes and 
- add new API values to KPI and Status settings includes: 
   - showTimestamp
   - showAssetName
   - backgroundColor
values not yet configurable in dashboard

image

## Legal


packages/react-components/src/common/dataTypes.ts
```diff
@@ -63,6 +63,9 @@ export type WidgetSettings = {
   name?: string;
```
corteggiano, line 63: moved resolution up to this level and out of KPI/Status settings

ssjagad (approved):

Issue #2565, opened 2024-02-09
[Dashboard] dashboard configuration viewport setting is ignored


diehbria:
the viewport setting in the dashboard configuration is ignored. The expected behavior is for the dashboard to utilize the viewport specified in the dashboard configuration on initial load

```[tasklist]
### acceptance criteria
- [x] add test case to prevent regression
- [x] dashboard on initial load utilizes the specified viewport
- [ ] dashboard only saves relative viewports (eg: 5m, 15m, 1hour)
- [ ] dashboard does NOT save absolute viewports
```

PR #2611, opened 2024-02-22
fix: updated the viewport settings for dynamic input

Chandru-HM:
## Overview
This PR is for ticket #2565 and updated viewport settings for the dynamic input timestamp

## Verifying Changes
https://github.com/awslabs/iot-app-kit/assets/142866907/c55e94c2-6846-442d-8c77-2c72d7675216

## Legal


corteggiano (approved):
ran locally but found some issues

corteggiano:
reverting commit because all viewport are saved, we dont want absolute viewports to persist (aka, we should save `last 5 minutes` but we should not save a viewport that is `feb 21 - feb22`

PR #2614, opened 2024-02-23
fix: updated the ux for widget actions

Chandru-HM:
## Overview
updated the UX as per comments in https://github.com/awslabs/iot-app-kit/pull/2589 for the ticket [#2439](https://github.com/awslabs/iot-app-kit/issues/2439). and updated the vertical padding and the border radius to 2px and 8px.

## Verifying Changes
old: 
image
  
new update:
image

## Legal

### DIFF
--- a/packages/dashboard/src/components/widgets/widgetActions.tsx
+++ b/packages/dashboard/src/components/widgets/widgetActions.tsx
@@ -7,9 +7,8 @@ import {
   colorBackgroundButtonNormalDefault,
   spaceStaticXl,
   spaceStaticL,
-  spaceStaticXxs,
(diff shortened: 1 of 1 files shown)


dpportet (changes requested):
The styling fix looks good, thanks for the update.

Could you also check when the legend is hidden and make sure the hover icons aren't covering the chart gesture icons? We did a bug bash yesterday and noticed the chart gestures would get covered by the hover icons and couldn't be used. Attached a screenshot. 

image

dpportet (approved):
Approving and will follow up on the hover icon fix in a separate PR.

PR #2616, opened 2024-02-23
feat: add threshold settings to KPI

corteggiano:
## Overview
Add threshold style settings to KPI and Status (gated only!) 

this is what the different settings look like
with/without timestamp + with line or filled widget
Screenshot 2024-02-23 at 12 11 27

video of threshold being added to new KPI widget 

https://github.com/awslabs/iot-app-kit/assets/28601414/c2a06a01-2563-45b2-ad2b-503831bc37c9

Also cleaned up some types 

## Legal

### DIFF
--- a/packages/core/src/common/types.ts
+++ b/packages/core/src/common/types.ts
@@ -102,6 +102,13 @@ export interface Threshold<T extends ThresholdValue = ThresholdValue>
   dataStreamIds?: DataStreamId[];
 }
 
+export type ThresholdStyleType = {
+  visible?: boolean;
+  fill?: string;
+};
+
+export type StyledThreshold = Threshold & ThresholdStyleType;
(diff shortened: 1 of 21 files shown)


packages/core/src/common/types.ts
```diff
@@ -102,6 +102,13 @@ export interface Threshold<T extends ThresholdValue = ThresholdValue>
   dataStreamIds?: DataStreamId[];
 }
 
+export type ThresholdStyleType = {
```
corteggiano, line 105: moved the new types here, since they are referenced in both dashboard and react-components

tracy-french (approved):
Looks great! Discussed with @corteggiano offline we plan to complete comprehensive testing pre-release using Playwright.

PR #2627, opened 2024-02-27
fix(dashboard): style fixes to make the dashboard accessible at small screensizes

jmbuss:
## Overview
In order to make the dashboard accessible, the content needs to reflow correctly at small or zoomed in screen sizes. It should be accessible at a minimum viewport width of 320px without requiring horizontal scrolling.

These fixes are so that the toolbar and resource / config panels do not cause overflow when the viewport is small.

## Legal


chejimmy (approved):
LGTM! Making the UX responsive <3

PR #2636, opened 2024-02-28
fix(dashboard): revert the title change for line chart

jmbuss:
## Overview
A recent change to the widget tile moved the title of the xy-plot widget from the tile component into the chart. The tile title acted as a drag area for the widget, meaning the widget drag UX degraded. As a temporary fix, this change adds the title back to the tile.

## Legal


dpportet (approved):

chejimmy (commented):
Nice fix 🙏 TY

PR #2640, opened 2024-02-29
feat: user selected dashboard refresh-rate

ssjagad:
## Overview
Allow the user to select a refresh rate is used by the dashboard to get latest data. 

## Verifying Changes

Persistance + Warning
https://github.com/awslabs/iot-app-kit/assets/145582655/ccde2300-efde-4fd3-9b1b-e5de2934f8ac

KPI/Status

https://github.com/awslabs/iot-app-kit/assets/145582655/80aa2982-bfc9-4af4-9ca9-7ea910e5e234

All widgets except Line

https://github.com/awslabs/iot-app-kit/assets/145582655/b6399201-1dbe-4cc7-8322-210e259d34a5

## Legal

### DIFF
--- a/packages/core/src/data-module/TimeSeriesDataModule.ts
+++ b/packages/core/src/data-module/TimeSeriesDataModule.ts
@@ -26,7 +26,7 @@ import type {
 
 export const DEFAULT_CACHE_SETTINGS = {
   ttlDurationMapping: {
-    [1.2 * MINUTE_IN_MS]: SECOND_IN_MS * 1.5,
+    [1.2 * MINUTE_IN_MS]: SECOND_IN_MS,
     [3 * MINUTE_IN_MS]: 30 * SECOND_IN_MS,
     [20 * MINUTE_IN_MS]: 5 * MINUTE_IN_MS,
   },
--- a/packages/dashboard/src/components/dashboard/index.test.tsx
+++ b/packages/dashboard/src/components/dashboard/index.test.tsx
@@ -20,6 +20,7 @@ it('renders', function () {
         },
         widgets: [],
         viewport: { duration: '5m' },
+        querySettings: {},
       }}
       clientConfiguration={{
         iotEventsClient: createMockIoTEventsSDK(),
@@ -48,6 +49,7 @@ it('renders in readonly initially', function () {
         },
         widgets: [],
         viewport: { duration: '5m' },
+        querySettings: {},
       }}
       clientConfiguration={{
         iotEventsClient: createMockIoTEventsSDK(),
@@ -84,6 +86,7 @@ it('renders in edit initially', function () {
         },
         widgets: [],
         viewport: { duration: '5m' },
+        querySettings: {},
       }}
       clientConfiguration={{
         iotEventsClient: createMockIoTEventsSDK(),
@@ -117,6 +120,9 @@ it('passes the correct viewMode to onSave', function () {
     },
     widgets: [],
     viewport: { duration: '5m' },
+    querySettings: {
+      refreshRate: 5000,
+    },
   };
 
   render(
@@ -157,6 +163,7 @@ it('renders dashboard name', function () {
         },
         widgets: [],
         viewport: { duration: '5m' },
+        querySettings: {},
       }}
       name='Test dashboard'
       clientConfiguration={{
@@ -183,6 +190,7 @@ it('renders without dashboard name', function () {
         },
         widgets: [],
         viewport: { duration: '5m' },
+        querySettings: {},
       }}
       clientConfiguration={{
         iotEventsClient: createMockIoTEventsSDK(),
--- a/packages/dashboard/src/components/internalDashboard/dashboardHeader.tsx
+++ b/packages/dashboard/src/components/internalDashboard/dashboardHeader.tsx
@@ -2,6 +2,7 @@ import React, { memo } from 'react';
 
 import { Box, Header, SpaceBetween } from '@cloudscape-design/components';
 import { TimeSelection } from '@iot-app-kit/react-components';
+import { RefreshRateDropDown } from '../querySettingsSync/refreshRateDropDown';
 
 import {
   colorChartsLineGrid,
@@ -11,7 +12,11 @@ import {
 } from '@cloudscape-design/design-tokens';
 
 import Actions from '../actions';
-import type { DashboardSave, DashboardWidget } from '~/types';
+import type {
+  DashboardSave,
+  DashboardTimeSeriesSettings,
+  DashboardWidget,
+} from '~/types';
 
 type DashboardHeaderProps = {
   name?: string;
@@ -19,6 +24,7 @@ type DashboardHeaderProps = {
   readOnly: boolean;
   dashboardConfiguration: {
     widgets: DashboardWidget<Record<string, unknown>>[];
+    querySettings?: DashboardTimeSeriesSettings;
   };
   grid: {
     enabled: boolean;
@@ -57,6 +63,7 @@ const DashboardHeader = ({
     <Box float='right'>
       <SpaceBetween size='s' direction='horizontal' alignItems='end'>
         <TimeSelection isPaginationEnabled />
+        <RefreshRateDropDown />
         {editable && (
           <>
             <Divider key='2' />
--- a/packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.spec.tsx
+++ b/packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.spec.tsx
@@ -1,0 +1,56 @@
+import React, { PropsWithChildren } from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { RefreshRateDropDown } from './refreshRateDropDown';
+import { Provider } from 'react-redux';
+import { configureDashboardStore } from '~/store';
+
+describe('Refresh rate drop down', () => {
+  const ProviderWrapper = ({ children }: PropsWithChildren) => (
+    <Provider store={configureDashboardStore()}>{children}</Provider>
+  );
+
+  it('selecting option changes on screen', async () => {
+    const dropDownElement = <RefreshRateDropDown />;
+    const user = userEvent.setup();
+    render(dropDownElement, { wrapper: ProviderWrapper });
+    const dropDown = screen.getByLabelText('Refresh rate');
+    expect(screen.queryByText('1s')).not.toBeInTheDocument();
+    expect(screen.getByText('5s')).toBeVisible();
+    await user.click(dropDown);
+    const oneSecondOption = screen.getByText('1s');
+    await user.click(oneSecondOption);
+    expect(screen.queryByText('5s')).not.toBeInTheDocument();
+    expect(screen.getByText('1s')).toBeVisible();
+  });
+
+  it('all options are visible on dropdown click', async () => {
+    const dropDownElement = <RefreshRateDropDown />;
+    const user = userEvent.setup();
+    render(dropDownElement, { wrapper: ProviderWrapper });
+    const dropDown = screen.getByLabelText('Refresh rate');
+    await user.click(dropDown);
+    expect(screen.getByText('1s')).toBeVisible();
+    const fiveSecondOptions = screen.getAllByText('5s');
+    fiveSecondOptions.forEach((option) => expect(option).toBeVisible()); //default option + dropdown option
+    expect(screen.getByText('10s')).toBeVisible();
+    expect(screen.getByText('1m')).toBeVisible();
+    expect(screen.getByText('5m')).toBeVisible();
+  });
+
+  it('triggers modal when 1s option is selected', async () => {
+    const dropDownElement = <RefreshRateDropDown />;
+    const user = userEvent.setup();
+    render(dropDownElement, { wrapper: ProviderWrapper });
+    expect(screen.getByText('5s')).toBeVisible();
+    const dropDown = screen.getByLabelText('Refresh rate');
+    await user.click(dropDown);
+    const oneSecondOption = screen.getByText('1s');
+    await user.click(oneSecondOption);
+    expect(
+      screen.getByText(
+        'You may experience some lag by selecting the 1 second refresh rate.'
+      )
+    ).toBeVisible();
+  });
+});
--- a/packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.tsx
+++ b/packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.tsx
@@ -1,0 +1,46 @@
+import { Alert, FormField, Modal, Select } from '@cloudscape-design/components';
+import React, { useState } from 'react';
+import {
+  DEFAULT_REFRESH_RATE,
+  DEFAULT_OPTION,
+  refreshRateOptionMap,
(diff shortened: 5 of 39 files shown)


packages/core/src/data-module/TimeSeriesDataModule.ts
```diff
@@ -26,7 +26,7 @@ import type {
 
 export const DEFAULT_CACHE_SETTINGS = {
   ttlDurationMapping: {
-    [1.2 * MINUTE_IN_MS]: SECOND_IN_MS * 1.5,
+    [0.5 * MINUTE_IN_MS]: SECOND_IN_MS,
```
chejimmy, line 29: will want to update the corresponding doc under docs/TimeSeriesDataFeatures.md

packages/dashboard/src/components/internalDashboard/dashboardHeader.tsx
```diff
@@ -11,14 +12,19 @@ import {
...
+  DashboardWidget,
+} from '~/types';
 
 type DashboardHeaderProps = {
   name?: string;
   editable?: boolean;
   readOnly: boolean;
   dashboardConfiguration: {
     widgets: DashboardWidget<Record<string, unknown>>[];
+    querySettings: DashboardTimeSeriesSettings;
```
chejimmy, line 27: does `querySettings` need to be optional to maintain backward compatibility?

packages/dashboard/src/components/dashboard/index.test.tsx
```diff
@@ -20,6 +20,9 @@ it('renders', function () {
         },
         widgets: [],
         viewport: { duration: '5m' },
+        querySettings: {
```
chejimmy, line 23: is it necessary to provide the `querySettings.refreshRate`? it should default to 5s without it

packages/core/src/data-module/TimeSeriesDataModule.ts
tracy-french, line 29: If I am reading this right, this would probably halve the amount of data requested?

packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <FormField label='Refresh rate' data-testid='refresh-rate'>
+      <Select
+        selectedOption={selectedRefreshRate}
+        onChange={({ detail }) => {
+          setSelectedRefreshRate(detail.selectedOption);
+          dispatch(
+            onChangeRefreshRateAction({
+              refreshRate: parseInt(detail.selectedOption.value ?? '5000'),
```
tracy-french, line 30: Magic number needs a constant?

packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const RefreshRateDropDown = () => {
+  const dispatch = useDispatch();
+  const refreshRate = useSelector(
+    (state: DashboardState) =>
+      state.dashboardConfiguration.querySettings.refreshRate
+  );
+  const [selectedRefreshRate, setSelectedRefreshRate] =
+    useState<SelectedOption>(selectIntialRefreshRate(refreshRate));
+
```
tracy-french, line 21: Why do you need this state to store the option? It seems like additional state for the selected option could be avoided by creating the option each render.

packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.tsx
```diff
@@ -0,0 +1,44 @@
...
+            })
+          );
+        }}
+        options={[
+          { label: '1s', value: '1000' },
+          { label: '5s', value: '5000' },
+          { label: '10s', value: '10000' },
+          { label: '1m', value: '60000' },
+          { label: '5m', value: '300000' },
+        ]}
```
tracy-french, line 40: Use constants please. You have this list of numbers duplicated in `selectInitialRefreshRate`.

packages/dashboard/src/components/querySettingsSync/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+    return { label: '5s', value: '5000' };
+  } else if (refreshRate === 10000) {
+    return { label: '10s', value: '10000' };
+  } else if (refreshRate === 60000) {
+    return { label: '1m', value: '60000' };
+  } else if (refreshRate === 300000) {
+    return { label: '5m', value: '300000' };
+  }
+  return { label: '5s', value: '5000' };
+};
```
tracy-french, line 14: Try using a map object instead of a bunch of conditionals to clean this code up.

packages/dashboard/src/components/querySettingsSync/utils.ts
```diff
@@ -0,0 +1,14 @@
+export const selectIntialRefreshRate = (refreshRate?: number) => {
```
tracy-french, line 1: The name for this function `select*` sounds like it is mutating and initializing something. It's more it should be called `createRefreshRateOption`.

packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.tsx
```diff
@@ -0,0 +1,44 @@
...
+  label?: string;
+  value?: string;
+};
+
+export const RefreshRateDropDown = () => {
+  const dispatch = useDispatch();
+  const refreshRate = useSelector(
+    (state: DashboardState) =>
+      state.dashboardConfiguration.querySettings.refreshRate
+  );
```
tracy-french, line 18: Should be using your custom hook?

packages/dashboard/src/customization/hooks/useRefreshRate.ts
```diff
@@ -0,0 +1,13 @@
...
+export const useRefreshRate = () => {
+  const refreshRate = useSelector(
+    (state: DashboardState) =>
+      state.dashboardConfiguration.querySettings.refreshRate
+  );
+
+  return {
+    refreshRate,
+  };
+};
```
tracy-french, line 13: Improve this by also encapsulating the dispatch to update state. You could return a function called `updateRate` or similar.

packages/dashboard/src/store/actions/changeRefreshRate/index.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export const onChangeRefreshRateAction = (
+  payload: ChangeRefreshRateActionPayload
+): ChangeRefreshRateAction => ({
+  type: 'CHANGE_REFRESH_RATE',
+  payload,
+});
+
+export const changeRefreshRate = (
```
tracy-french, line 20: I think update is a better word than change.

packages/core/src/data-module/TimeSeriesDataModule.ts
ssjagad: Yup, this is where the 30 seconds of data that we were seeing in the bug bash is coming from.

ssjagad: there should be another interval added [1.2 * MINUTE_IN_MS]: SECOND_IN_MS * 5, to maintain what was there before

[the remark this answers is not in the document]
ssjagad: yes, the cache logic already has tests that have been written.

chejimmy (dismissed):
LGTM! great addition to app kit 🚀

packages/core/src/data-module/TimeSeriesDataModule.ts
ssjagad: I chose to stick with the intervals we already had since they are tested, and just reduced the right hand value to one second so we can support 1 second rate

packages/dashboard/src/customization/widgets/kpi/component.tsx
```diff
@@ -13,13 +13,17 @@ import { getAggregation } from '../utils/widgetAggregationUtils';
...
 
 const KPIWidgetComponent: React.FC<KPIWidget> = (widget) => {
   const { viewport } = useViewport();
   const dashboardSignificantDigits = useSelector(
     (state: DashboardState) => state.significantDigits
   );
 
+  const { refreshRate } = useRefreshRate();
+
+  //write unit test here
```
tracy-french, line 26: Dangling comment?

packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { SECOND_IN_MS } from '@iot-app-kit/core';
+
+export const RefreshRateDropDown = () => {
+  const { refreshRate, updateRefreshRate } = useRefreshRate();
+  const [visible, setVisible] = useState(refreshRate === SECOND_IN_MS);
+
+  const option = refreshRateOptionMap.get(refreshRate ?? DEFAULT_REFRESH_RATE);
+
+  return (
+    <div>
```
tracy-french, line 18: Use a fragment instead of a div to avoid unnecessary DOM nodes.

packages/dashboard/src/components/querySettingsSync/refreshRateDropDown.spec.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { PropsWithChildren } from 'react';
+import { render, screen } from '@testing-library/react';
+import { RefreshRateDropDown } from './refreshRateDropDown';
+import { Provider } from 'react-redux';
+import { configureDashboardStore } from '~/store';
+
+describe('Refresh rate drop down', () => {
```
tracy-french, line 7: You're missing a test for changing the drop-down. Could you please add that?

packages/dashboard/src/components/querySettingsSync/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+  label?: string;
+  value?: string;
+};
+
+const SECOND_IN_MS = 1000;
+const MINUTE_IN_MS = 60000;
+export const DEFAULT_REFRESH_RATE = 5000; // 5 seconds
+export const DEFAULT_OPTION = { label: '5s', value: '5000' };
+
+export const refreshRateOptionMap = new Map<number, SelectedOption>([
```
tracy-french, line 11: I would avoid using and exporting a raw `Map` object. It's an implementation detail which is hard to notice without visiting the file. You can see how the abstraction leaks in `refreshRateDropDown` as you're having to call specific `Map` APIs (i.e., `.get()` and `.values()`).

If you want to use a map, encapsulate it does not require the user to know it's a map. You can do that by creating an API like `findOption` and `getOptions`.

packages/dashboard/src/components/querySettingsSync/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+  [SECOND_IN_MS, { label: '1s', value: '1000' }],
+  [5 * SECOND_IN_MS, { label: '5s', value: '5000' }],
+  [10 * SECOND_IN_MS, { label: '10s', value: '10000' }],
+  [MINUTE_IN_MS, { label: '1m', value: '60000' }],
+  [5 * MINUTE_IN_MS, { label: '5m', value: '300000' }],
+]);
```
tracy-french, line 17: I think there's room to improve this handling. Let's talk offline about the following:

```ts
// in types.ts
export const SECOND_IN_MS = 1000;
export const FIVE_SECONDS_IN_MS = 5 * SECOND_IN_MS;
export const TEN_SECONDS_IN_MS = 10 * SECOND_IN_MS;
export const MINUTE_IN_MS = 60 * SECOND_IN_MS;
export const FIVE_MINUTES_IN_MS = 5 * MINUTE_IN_MS;

// use the type everywhere there is a refresh rate (instead of number)
export type RefeshRate = 
  | SECOND_IN_MS 
  | FIVE_SECONDS_IN_MS 
  | TEN_SECONDS_IN_MS 
  | MINUTE_IN_MS 
  | FIVE_MINUTES_IN_MS;

// in utils.ts
import { type SelectProps } from '@cloudscape-design/components';
import { SECOND_IN_MS, FIVE_SECONDS_IN_MS, TEN_SECONDS_IN_MS, MINUTE_IN_MS, FIVE_MINUTES_IN_MS, type RefreshRate } from './types';

interface RefreshRateOption extends SelectProps['options'][number] { 
  value: RefreshRate;
};

const refreshRateOptions = [
  { label: '1s', value: SECOND_IN_MS.toString() },
  { label: '5s', value: FIVE_SECONDS_IN_MS.toString() },
  { label: '10s', value: TEN_SECONDS_IN_MS.toString() },
  { label: '1m', value: MINUTE_IN_MS.toString() },
  { label: '5m', value: FIVE_MINUTES_IN_MS.toString() },
] as const satisfies RefreshRateOption[];

const DEFAULT_REFRESH_RATE_OPTION = refreshRateOptions[1];

const refreshRateOptionMap = refreshRateOptions.reduce((optionMap, { label, value }) => ({ ...optionMap, [parseInt(value, 10)]: { label, value }}), {});

// in component.tsx

// this should work and return a defined value because refresh rate is a discrete set
const currentOption = refreshRateOptionMap[refreshRate];

..

options={refreshRateOptions}
```

tracy-french (changes requested):
A few minor things. The main thing preventing approval is missing the test to change the drop-down. The rest can be followed up.

packages/dashboard/src/customization/hooks/useRefreshRate.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const refreshRate = useSelector(
+    (state: DashboardState) =>
+      state.dashboardConfiguration.querySettings?.refreshRate
+  );
+
+  return {
+    refreshRate,
+    updateRefreshRate,
+  };
```
tracy-french, line 24: return [refreshRate, updateRefreshRate] as const;

packages/dashboard/src/components/querySettingsSync/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  SECOND_IN_MS,
+  FIVE_SECONDS_IN_MS,
+  TEN_SECONDS_IN_MS,
+  MINUTE_IN_MS,
+  FIVE_MINUTES_IN_MS,
+  type SelectedOption,
+} from './types';
+
+export const DEFAULT_REFRESH_RATE = 5000; // 5 seconds
+export const DEFAULT_OPTION = { label: '5s', value: '5000' };
```
tracy-french, line 11: Are these are duplicates?

tracy-french (commented):
Awesome refactor and excellent tests! You're thinking like a user! Well done.

Swap out the div for a fragment and we'll be good to ship!

packages/dashboard/src/customization/hooks/useRefreshRate.ts
```diff
@@ -0,0 +1,22 @@
+import { useDispatch, useSelector } from 'react-redux';
+import { onUpdateRefreshRateAction } from '~/store/actions';
+import { DashboardState } from '~/store/state';
+
+export const useRefreshRate = () => {
+  const dispatch = useDispatch();
+
+  const updateRefreshRate = (refreshRate: number) => {
```
tracy-french, line 8: This should use your RefreshRate type.

tracy-french (approved):

PR #2642, opened 2024-02-29
fix: updated the viewport settings for dynamic input

Chandru-HM:
## Overview
This PR is for ticket https://github.com/awslabs/iot-app-kit/issues/2565 and updated viewport settings for the dynamic input timestamp

## Verifying Changes
https://github.com/awslabs/iot-app-kit/assets/142866907/1c98f317-af87-41c5-9788-76e5ddaed450

https://github.com/awslabs/iot-app-kit/assets/142866907/480ddeb6-f177-4f7e-b890-505ac3ccb7ea

## Legal

### DIFF
--- a/packages/dashboard/src/components/dashboard/index.tsx
+++ b/packages/dashboard/src/components/dashboard/index.tsx
@@ -77,6 +77,7 @@ const Dashboard: React.FC<DashboardProperties> = ({
                   editable={true}
                   name={name}
                   propertiesPanel={<PropertiesPanel />}
+                  viewport={dashboardConfiguration.viewport}
                 />
               </DndProvider>
             </Provider>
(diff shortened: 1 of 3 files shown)


tracy-french (approved):

packages/dashboard/src/components/dashboard/index.test.tsx
```diff
@@ -19,7 +19,7 @@ it('renders', function () {
           numRows: 100,
         },
         widgets: [],
-        viewport: { duration: '5m' },
```
corteggiano, line 22: instead of changing this test could you create a new test which does the following:

1. check that if no initial viewport is passed in, the dashboard uses a default viewport
2. check that if a viewport is passed in, the dashboard uses that value instead

corteggiano (changes requested):

corteggiano (approved):
great work! thank you :)

PR #2671, opened 2024-03-12
feat(tmdt): support endpoint-url parameter and respect S3 bucket name character limit

hwandersman:
## Overview

Adding support for specifying a different endpoint URL for interacting with IoT TwinMaker. This is an optional parameter for the init, deploy, and destroy commands.

When deploying a TMDT workspace from scratch, if the endpoint url points to TwinMaker gamma then the IAM role must have a trust policy with the gamma service principal.

Was running into a bug where my workspace name caused S3 to throw an error when creating a bucket with too long a name. Added a method `generateS3BucketName` to ensure the S3 bucket name character limit is met.

## Verifying Changes

Tested each command with a gamma workspace successfully.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/tools-iottwinmaker/src/commands/deploy.ts
+++ b/packages/tools-iottwinmaker/src/commands/deploy.ts
@@ -29,6 +29,8 @@ export type Options = {
   region: string;
   'workspace-id': string;
   dir: string;
+  'endpoint-url': string;
+  'execution-role': string;
 };
 
 export const command = 'deploy';
@@ -51,17 +53,33 @@ export const builder: CommandBuilder<Options> = (yargs) =>
       require: true,
       description: 'Specify the project location, directory for tmdt.json file',
     },
+    'endpoint-url': {
+      type: 'string',
+      require: false,
+      description: 'Specify the AWS IoT TwinMaker endpoint.',
+      default: '',
+    },
+    'execution-role': {
+      type: 'string',
+      require: false,
+      description:
+        'The name of the execution role associated with the workspace.',
+      default: '',
+    },
   });
 
 export const handler = async (argv: Arguments<Options>) => {
   const workspaceId: string = argv['workspace-id']; // TODO allow it to be optional (i.e. option to autogenerate workspace for them)
   const region: string = argv.region;
   const dir: string = argv.dir;
+  const tmEndpoint: string = argv['endpoint-url'];
+  const executionRole: string = argv['execution-role'];
+
   console.log(
     `Deploying project from directory ${dir} into workspace ${workspaceId} in ${region}`
   );
 
-  initDefaultAwsClients({ region: region });
+  initDefaultAwsClients({ region, tmEndpoint });
   if (!fs.existsSync(path.join(dir, 'tmdt.json'))) {
     throw new Error('TDMK.json does not exist. Please run tmdt init first.');
   }
@@ -85,7 +103,7 @@ export const handler = async (argv: Arguments<Options>) => {
         workspace role to continue deployment (Y)? Press any other key to abort (n).`,
       });
       if (response.confirmation === 'Y') {
-        await createWorkspaceIfNotExists(workspaceId);
+        await createWorkspaceIfNotExists(workspaceId, executionRole);
       } else {
         console.log('Aborting deployment...');
         return 0;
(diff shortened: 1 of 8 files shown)


packages/tools-iottwinmaker/src/commands/deploy.ts
```diff
@@ -85,7 +93,9 @@ export const handler = async (argv: Arguments<Options>) => {
         workspace role to continue deployment (Y)? Press any other key to abort (n).`,
       });
       if (response.confirmation === 'Y') {
-        await createWorkspaceIfNotExists(workspaceId);
+        // Support gamma endpoint for TwinMaker
+        const isGamma = !!tmEndpoint && tmEndpoint.includes('gamma');
```
johnnyw-aws, line 97: we should probably remove references to "gamma" since those are internal details

i.e. instead of exposing the concept of gamma, have a way for consumers to override the places that would change if pointing to a specific endpoint

packages/tools-iottwinmaker/src/lib/policy/workspace-role-assume-policy.ts
```diff
@@ -1,12 +1,18 @@
...
+export const getWorkspaceRoleAssumePolicy = (isGamma: boolean) => {
+  return {
+    Version: '2012-10-17',
+    Statement: [
+      {
+        Effect: 'Allow',
+        Principal: {
+          Service: [
+            isGamma
+              ? 'iottwinmaker.aws.internal'
```
johnnyw-aws, line 10: similar comment: probably shouldn't publish this in the code

can we instead have a way for to provide a custom workspace role policy if needed?

johnnyw-aws (changes requested):

packages/tools-iottwinmaker/src/commands/deploy.ts
```diff
@@ -51,17 +52,24 @@ export const builder: CommandBuilder<Options> = (yargs) =>
...
   const workspaceId: string = argv['workspace-id']; // TODO allow it to be optional (i.e. option to autogenerate workspace for them)
   const region: string = argv.region;
   const dir: string = argv.dir;
+  const tmEndpoint: string = argv['endpoint-url'];
   console.log(
     `Deploying project from directory ${dir} into workspace ${workspaceId} in ${region}`
   );
 
-  initDefaultAwsClients({ region: region });
+  initDefaultAwsClients({ region, tmEndpoint });
```
gsechkin-aws, line 72: strict types here? e.g. (region: region, tmEndpoint: string)

packages/tools-iottwinmaker/src/commands/destroy.ts
```diff
@@ -64,8 +71,9 @@ export const handler = async (argv: Arguments<Options>) => {
   let deleteWorkspaceFlag: boolean = argv['delete-workspace'];
   let deleteS3Flag: boolean = argv['delete-s3-bucket'];
   const nonDryRun: boolean = argv['non-dry-run'];
+  const tmEndpoint: string = argv['endpoint-url'];
 
-  initDefaultAwsClients({ region: region });
+  initDefaultAwsClients({ region, tmEndpoint });
```
gsechkin-aws, line 76: same as above

packages/tools-iottwinmaker/src/commands/init.ts
```diff
@@ -417,11 +424,12 @@ export const handler = async (argv: Arguments<Options>) => {
   const workspaceId: string = argv['workspace-id'];
   const region: string = argv.region;
   const outDir: string = argv.out;
+  const tmEndpoint: string = argv['endpoint-url'];
   console.log(
     `Bootstrapping project from workspace ${workspaceId} in ${region} at project directory ${outDir}`
   );
 
-  initDefaultAwsClients({ region: region });
+  initDefaultAwsClients({ region, tmEndpoint });
```
gsechkin-aws, line 432: ditto

packages/tools-iottwinmaker/src/commands/deploy.ts
hwandersman: `region` isn't a type, it's the name of the variable and the name of the input field. In Typescript you don't need to specify the value of a named input if the name is the same.

i.e. `{ region: region }` is equivalent and inferred to be `{ region }`

johnnyw-aws (approved):
thanks!

Issue #2663, opened 2024-03-07
[Widgets] Timezone support via `timeZone`


diehbria:
This task is to add timezone support for the shared widgets of IoT App Kit.

Add a new optional property, `timeZone?: string;`. This will be used to format dates into a different timezone than the where the one the user is in.

**Implementation details:**
We want to use https://date-fns.org/ to format dates.

```
const { utcToZonedTime, format } = require("date-fns-tz");

// https://date-fns.org/v3.6.0/docs/Time-Zones#date-fns-tz
// converts a utc date to a formatted string in a specific timeZone
const formatDate = (dateTime: number, { timeZone, pattern }: { timeZone: string; pattern: string; }) => {
  const zonedDate = utcToZonedTime(new Date(dateTime), timeZone);
  const formattedDate = format(zonedDate, pattern, { timeZone });

  return formattedDate;
};
```

How to handle timezones in echarts - https://github.com/apache/echarts/issues/14453

How to handle dates in our React components -
```
// Helper components for use in a React Context
export type DateTimeFormatContextOptions = {
  timeZone: string;
};
export const DateTimeFormatContext = React.createContext<DateTimeFormatContextOptions>({
  timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone,
});

export type DateTimeOptions = {
  dateTime: number;
  pattern?: string;
};
export const DateTime = ({ dateTime, pattern }: DateTimeOptions) => {
  const dateTimeFormatPattern = pattern ?? 'yyy-MM-dd hh:mm:ss a';
  const { timeZone } = useContext(DateTimeFormatContext);

  return <>{ formatDate(dateTime, { timeZone, pattern: dateTimeFormatPattern }) }</>;
};
```

Default: the user's current timeZone as provided by the browser. `Intl.DateTimeFormat().resolvedOptions().timeZone`

For KPI, Status, make sure the bottom timestamp displayed on the widget reflects the updated timezone
![Screenshot 2024-03-08 at 8 06 03 AM](https://github.com/awslabs/iot-app-kit/assets/77755322/1f7edcfb-9a78-4995-bb54-b3ad63477473)

For resource explorer, ensure that all places which display timestamps of data as a result of searches, reflects the timezone offset.

For XY Plot, ensure the bottom timestamp displayed reflects the timezone offset, the data point tooltip, as well as the X-Axis labels, and the timestamps shown within the legend and the trend cursor column headers.

### Expected API behavior
[...]

PR #2680, opened 2024-03-14
feat: timezone support via get time offset

Chandru-HM:
This PR is for ticket https://github.com/awslabs/iot-app-kit/issues/2663 and added the timezone support via get time offset.

## Verifying Changes
https://github.com/awslabs/iot-app-kit/assets/142866907/53db3ac3-e5cf-4b44-a69b-d177413b1b6d

## Legal


packages/react-components/src/components/kpi/kpiBase.tsx
```diff
@@ -129,7 +130,14 @@ export const KpiBase: React.FC<KPIBaseProperties> = ({
                   }}
                 />
                 <div className='timestamp' data-testid='kpi-timestamp'>
-                  {isLoading ? '-' : new Date(point.x).toLocaleString()}
+                  {isLoading ? (
+                    '-'
+                  ) : (
+                    <DateTime
+                      dateTime={point.x}
+                      pattern='dd/MM/yyyy, HH:mm:ss'
```
jmbuss, line 138: Can we make these constants? I believe this exists already here - packages/react-components/src/utils/time.ts

jmbuss: Also, let's default the DateTime componeent to this so we don't need to pass it everywhere.

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/modeledDataStreamQueryEditor/assetExplorer/assetTable/assetTableColumnDefinitionsFactory.tsx
```diff
@@ -87,7 +87,14 @@ export class AssetTableColumnDefinitionsFactory {
       id: 'creationDate',
       header: 'Creation Date',
       cell: ({ creationDate }) =>
-        creationDate ? getFormattedDateTime(creationDate) : '-',
+        creationDate ? (
+          <DateTime
+            dateTime={creationDate.getTime()}
```
jmbuss, line 92: Is it possible to change creationDate to a number prop? I don't think we should do a bunch of conversions number -> date -> number ->date. I think all of the places we have dates (except for probably viewport) should start as numbers.

jmbuss (changes requested):

packages/react-components/src/components/kpi/kpiBase.tsx
Chandru-HM: Done.
Now the default pattern is 'M/d/yy HH:mm:ss' and created a constant ('dd/MM/yyyy, hh:mm:ss' ) as you mentioned to use in react-components.

Chandru-HM:
closed this PR due to many conflicts from the back merge. created the new one for the same https://github.com/awslabs/iot-app-kit/pull/2820

Issue #2667, opened 2024-03-08
[Widgets] Support theming


tracy-french:
To make it possible to get an `iot-app-kit` widget to match the styling of the consuming application, we will expose a hook-based theming API.

Below is the expected API, which should be exported from `@iot-app-kit/react-components';

```ts
import { type Theme } from '@cloudscape-design/components/theming';

/** Collection of Cloudscape design tokens supported for theming. */
type SupportedCloudscapeDesignTokens = Pick<Theme['tokens'], 'fontFamilyBase'>;
/** Collection of design tokens supported for theming. */
type Tokens = SupportedCloudscapeDesignTokens;

export interface UseThemeOptions {
  /**
   * Specifies the mode (i.e., color scheme) of the components.
   * 
   * @defaultValue `light`
   */
  mode?: 'light' | 'dark';
  /**
   * Specifies design token overrides to change the styling of components.
   */
  tokens?: Tokens;
}

/**
 * Use to set the theme of `iot-app-kit` components.
 *
 * @experimental
 * This API is experimental, is not completely functional, and is subject to 
 * breaking changes. Until stabilized, many components will not see styling 
 * changes. Do not use in production.
 */
export function useTheme({ mode = 'light', tokens }: UseThemeOptions): void {
  // implementation
}
```

The implementation should use Cloudscape mechanism for setting mode and theme. Please read the follow documents to understand how Cloudscape does this:
- https://cloudscape.design/foundation/visual-foundation/visual-modes/
- https://cloudscape.design/foundation/visual-foundation/theming/

The implementation should depend on the following imports:

```ts
import { applyMode, Mode } from '@cloudscape-design/global-styles';
import { applyTheme, type Theme } from '@cloudscape-design/components/theming';
```

For the initial implementation, we will only expose a single token (`fontFamilyBase`) and the mode selection. We are only exposing a little at a time as there is additional implementation on our side to ensure all custom components are using the appropriate Cloudscape design tokens, where applicable.

The font family theme setting should be fully functional. The mode setting should use `applyMode`, but making sure it works effectively will be completed in separate changes.

```[tasklist]
### Acceptance criteria
- [ ] `useTheme` and `UseThemeOptions` are exported from `@iot-app-kit/react-components`.
[...]

PR #2683, opened 2024-03-15
feat: support theming using the cloudscape mechanism

Chandru-HM:
This PR is for ticket https://github.com/awslabs/iot-app-kit/issues/2667 and added the support theming using the cloudscape mechanism.

## Verifying Changes
Tested applying in table **useTheme({ mode: 'dark', tokens: { fontFamilyBase: 'cursive' } });**
https://github.com/awslabs/iot-app-kit/assets/142866907/69a18227-bc8e-44f1-aa68-ce3c2c3d2e91

## Legal


packages/react-components/src/hooks/useTheming/index.ts
```diff
@@ -0,0 +1 @@
+export * from './useTheming';
```
tracy-french, line 1: Could you also export the hook and the types from the react-components package?

We want the import to look like:

```
import { useTheme, type UseThemeOptions, type Tokens, type Mode } from '@iot-app-kit/react-components';
```

tracy-french (changes requested):
Looks great! Before merging, let's add the export as commented.

packages/react-components/src/hooks/useTheming/index.ts
Chandru-HM: updated

tracy-french (approved):

PR #2698, opened 2024-03-23
fix(scene): handle bad texture files gracefully and toggle opacity on selection

haweston:
## Overview
1. Opacity doesn't change when a texture is selected for the ground plane which can make it seem like it did not load.  Now if it's zero the opacity will change 1 when a new texture is set. If the opacity is greater than zero it will not change.
2. Bad texture files could fail to load but get set as a material leading to error messages when the bad texture is removed from the ground plane.

## Verifying Changes
1. run story book scene composer
2. select the default composer scene
3. set the scene setting -> ground plane opacity to 0
4. select a texture for the ground plane.  verify the texture appears and the ground plane opacity is now 100%

5. edit the scene composer storybook to use an empty text file, or a non-texture fiile.
6. run storybook select texture for the scene composer ground plane in.  The file should URI should be set and the ground plane color will awy.  A blank white or black texture may be set.
7. remove the texture.  The ground plane color should return. there should be no error message in the scene composer.
8. select the texture again, and then remove it again.  There should still be no texture visible.  The second attempt is to verify that the caching in Three.js's file loader also didn't corrupt the texture.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/hooks/useTwinMakerTextureLoader.tsx
+++ b/packages/scene-composer/src/hooks/useTwinMakerTextureLoader.tsx
@@ -72,10 +72,14 @@ const useTwinMakerTextureLoader: (options?: TwinMakerTextureLoaderOptions) => {
         const oldTexture = material.map;
         material.color = new Color('#FFFFFF');
         loadTexture(uri, (result) => {
-          material.map = result as unknown as Texture;
-          material.needsUpdate = true;
+          if (result instanceof Texture) {
+            material.map = result as Texture;
+            material.needsUpdate = true;
+          } else {
+            console.error('Error parsing Texture file');
+          }
         });
-        if (oldTexture) {
+        if (oldTexture?.dispose) {
           oldTexture.dispose();
         }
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/hooks/useTwinMakerTextureLoader.tsx
```diff
@@ -72,10 +72,12 @@ const useTwinMakerTextureLoader: (options?: TwinMakerTextureLoaderOptions) => {
         const oldTexture = material.map;
         material.color = new Color('#FFFFFF');
         loadTexture(uri, (result) => {
-          material.map = result as unknown as Texture;
-          material.needsUpdate = true;
+          if (result instanceof Texture) {
```
hwandersman, line 75: We should log or show some error if the loaded texture is not valid

hwandersman (approved):

Issue #2567, opened 2024-02-09
[Dashboard] making the decimal place value too large crashes dashboard


diehbria:
When the decimal place settings is set to too large of a number, the dashboard crashes: 
```

RangeError: precision 123 out of range 
round number.js:20

```
![dec](https://github.com/awslabs/iot-app-kit/assets/77755322/e6ce4403-0c17-4d24-b0e8-6c35bc64ff8d)

```[tasklist]
### acceptance criteria
- [ ] prevent user from putting decimal place that will cause browser crash
```

Kausar-HM:
Hi @nadeswal

In this ticket we have to add validation to 'Decimal places' input, so we added the same in way it works for other fields:
![image](https://github.com/awslabs/iot-app-kit/assets/154328164/fee47758-3ff9-4bbb-a9f9-a0aee0d89b5c)

Currently the 'Decimal places' label is to the left of the input, in the config panel all other fields have the label at top, so should we move 'Decimal places' to the top as well?

![image](https://github.com/awslabs/iot-app-kit/assets/154328164/6d593461-277c-4a5f-a764-95d695f17edf)

nadeswal:
Here is the screenshot that you can use to replicate the design -
<img width="258" alt="Screenshot 2024-02-15 at 2 32 59 PM" src="https://github.com/awslabs/iot-app-kit/assets/141063047/c69cf156-8164-4c96-8c5e-22ba35a08c49">

jmbuss:
This is what the team discussed and agreed on - We will keep the decimal places as it is in it's own separate section but modify the section name to "Format data" instead of decimal places. Please refer to the screenshot below -

![Screenshot 2024-03-20 at 2 20 08 PM](https://github.com/awslabs/iot-app-kit/assets/107281089/345d52e2-f1c9-428c-bc6b-5731fc683f6b)

PR #2699, opened 2024-03-25
fix: decimal places input validation

Kausar-HM:
Added validation to the decimal places input to prevent dashboard from crashing as per issue #2567 
Also moved the "Resolution and Aggregation", and "Format data" sections at 2nd and 3rd position in style section
### Verifying changes:
[decimal-places-validation.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/d94a39fc-c768-43d9-961c-8d208303090a)

image

image
image


Kausar-HM:
Hey @jmbuss 
I have one query regarding the section sequence here

This is the sequence I have kept in my PR
image

But in the issue, this is the design:
image

I didn't insert the **Format data** section right below the **Y Axis** section, because those 3 sections are grouped under one component recently: ```packages/dashboard/src/customization/propertiesSections/lineAndScatterStyleSettings/section.tsx```
image

This group of sections is visible only for LineChart:
```
export const LineAndScatterStyleSettingsSection: React.FC = () => (
  <PropertiesSection
    isVisible={isLineAndScatterWidget}
    render={({ useProperty }) => (
      <RenderLineAndScatterStyleSettingsSection useProperty={useProperty} />
    )}
  />
);
```

So the sequence which I have used currently is fine or do we need to think of something else?

jmbuss:
Yes, this is fine to have the format data section after the aggregation / resolution section.

tracy-french (approved):

PR #2700, opened 2024-03-25
feat(react-components): add l4e queries

jmbuss:
## Overview
Tanstack queries for:
* DescribeAssetCommand
* DescribeAssetModelCompositeModelCommand
* DescribeAssetPropertyCommand
* GetAssetPropertyValueHistoryCommand

Hook for combining these for getting L4E data.

Storybook story for visualizing the results of these queries.

## Legal


packages/react-components/src/echarts/unboundedZoom/isDurationViewport.ts
```diff
@@ -0,0 +1,6 @@
+import { DurationViewport, Viewport } from '@iot-app-kit/core';
+
+export const isDurationViewport = (
```
corteggiano, line 3: this also exists in the `components/chart` folder
noting it down so we can dedupe as a followup

corteggiano (approved):

PR #2724, opened 2024-04-04
chore: upgrade dependent package zustand to 4.5.2

cshang2017:
Zustand made breaking changes when moving from version 3 to version 4. This commit upgrades the zustand dependency to the latest version 4.5.2, updates the codebase to accommodate the changes introduced in version 4, and removes obsolete test validation steps related to the previous version.

## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/StateManager.tsx
+++ b/packages/scene-composer/src/components/StateManager.tsx
@@ -328,20 +328,17 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
   // Subscribe to store update
   useEffect(() => {
     if (onSceneUpdated) {
-      return useStore(sceneComposerId).subscribe(
-        (state, old: Pick<RootState, 'document' | 'sceneLoaded'>) => {
-          // Do not call onSceneUpdated when
-          //  - scene is not loaded
-          //  - scene is just loaded
-          //  - document is not changed
-          // Transient document update will also trigger onSceneUpdated, the app side will debounce the actual saving to S3 call.
-          if (!state.sceneLoaded || !old.sceneLoaded || state.document === old.document) {
-            return;
-          }
-          onSceneUpdated(sceneDocumentSnapshotCreator.create({ document: state.document }));
-        },
-        (state) => ({ document: state.document, sceneLoaded: state.sceneLoaded }),
-      );
+      return useStore(sceneComposerId).subscribe((state, old: Pick<RootState, 'document' | 'sceneLoaded'>) => {
+        // Do not call onSceneUpdated when
+        //  - scene is not loaded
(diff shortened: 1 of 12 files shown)


packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -328,20 +328,17 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
...
-          //  - scene is not loaded
-          //  - scene is just loaded
-          //  - document is not changed
-          // Transient document update will also trigger onSceneUpdated, the app side will debounce the actual saving to S3 call.
-          if (!state.sceneLoaded || !old.sceneLoaded || state.document === old.document) {
-            return;
-          }
-          onSceneUpdated(sceneDocumentSnapshotCreator.create({ document: state.document }));
-        },
-        (state) => ({ document: state.document, sceneLoaded: state.sceneLoaded }),
```
haweston, line 343: So this gets  (state) => () streamlined away with the 4.x upgrade?

packages/scene-composer/src/store/slices/ViewOptionStateSlice.spec.ts
```diff
@@ -1,7 +1,12 @@
 import { ITagSettings, KnownComponentType } from '../../interfaces';
 import { Component } from '../../models/SceneModels';
 
-import { createViewOptionStateSlice } from './ViewOptionStateSlice';
+import { createViewOptionStateSlice as _createViewOptionStateSlice } from './ViewOptionStateSlice';
+
+const createViewOptionStateSlice = (set) => {
+  const { noHistoryStates } = _createViewOptionStateSlice(set);
+  return noHistoryStates;
```
haweston, line 8: This just to limit the return scope down to the noHistoryState parameter instead of the full list that _create gives? What's the reason for doing that?

packages/scene-composer/src/store/Store.ts
```diff
@@ -62,24 +63,33 @@ export type RootState = ISharedState &
...
+    ),
+  );
 
 // TODO: currently undoMiddleware will record editor state changes, such as select/deselect object.
 // We may want to fine-tune the undo/redo experience.
-const stores = new Map<string, UseStore<RootState>>();
+const stores = new Map<string, StoreState<RootState>>();
 
-const useStore: (id: string) => UseStore<RootState> = (id: string) => {
+const useStore = (id: string) => {
```
haweston, line 92: Can we not set the type on useStore anymore?  It's not something like StoreState<RootState> instead now?

haweston (approved):

PR #2726, opened 2024-04-08
feat(react-components): add l4e datasource

jmbuss:
## Overview

Create data module for all charts.
This module has a data loader which will be used to handle data ingestion / transformation for all charts, and transformers which will be able to transform specific data shapes into the correct encoding needed for a chart.

Create an abstract data transformer for arbitrary object types.
The first usage of this transformer is for anomaly data objects. These objects are the parsed complex data type for anomaly result composite properties. The only difference is that the name is expected to be the Property Name. The component will work the same if it is the guid, but the tooltip will would show the guid in that case.

Refactor the L4E widget to use the new datasource. Special handling is sometimes required because the format can either be columnar or row based. This is required because we need to map in a hardcoded bar value to control the height, and also have a customized tooltip.

## Legal


packages/react-components/src/components/l4eWidget/index.tsx
```diff
@@ -1,12 +1,20 @@
...
+import {
+  AnomalyObjectDataSourceTransformer,
+  DataSourceLoader,
+} from '../../data';
+
+const AnomalyDataSourceLoader = new DataSourceLoader([
+  new AnomalyObjectDataSourceTransformer(),
+]);
 
 export interface L4EWidgetProps {
```
corteggiano, line 19: delete this if its unused?

packages/react-components/src/components/l4eWidget/index.tsx
```diff
@@ -21,19 +29,24 @@ export interface L4EWidgetProps {
...
-  data,
+  datasources,
+  viewport,
   title,
   decimalPlaces,
-  viewportStart,
-  viewportEnd,
   tooltipSort,
-}: L4EWidgetProps) => {
+}: AnomalyChartOptions) => {
```
corteggiano, line 37: why call them Options instead Props?

packages/react-components/src/data/transformers/anomaly/object/transformer.ts
```diff
@@ -0,0 +1,124 @@
...
+ * calling GetAssetPropertyValueHistory for usage in
+ * an Anomaly Widget
+ *
+ * encoding: {
+ *  timestamp: number;
+ *  diagnostic_n: number;
+ *  diagnostic_n+1: number;
+ * }
+ */
+export class AnomalyObjectDataSourceTransformer extends ObjectDataSourceTransformer<
```
corteggiano, line 25: i love the use of classes for the transformers!

corteggiano (approved):

packages/react-components/src/data/README.md
```diff
@@ -0,0 +1,159 @@
+# Data Sources
```
corteggiano, line 1: awesome!

corteggiano (approved):

PR #2733, opened 2024-04-11
fix(scene): validation scene json on deserialize

haweston:
## Overview
Scene json load invalid values for
color hexcodes
opacity ranges
fog near and far distances.

If an invalid value is detected on load then the value will be replaced with the default value.

## Verifying Changes
Run storybook and load the invalidInputTest.scene.json file to verify that invalid values are converted to defaults

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/store/helpers/serializationHelpers.ts
+++ b/packages/scene-composer/src/store/helpers/serializationHelpers.ts
@@ -10,7 +10,18 @@ import {
   LEGACY_VERSION,
 } from '../../common/constants';
 import { ERROR_MESSAGE_DICT, ErrorCode, ErrorLevel, SceneComposerRuntimeError } from '../../common/errors';
-import { KnownComponentType } from '../../interfaces';
+import {
+  KnownComponentType,
+  KnownSceneProperty,
+  IFogSettings,
+  IGroundPlaneSettings,
+  ISceneBackgroundSetting,
+  DEFAULT_FOG_COLOR,
+  DEFAULT_FOG_NEAR,
+  DEFAULT_FOG_FAR,
+  DEFAULT_SCENE_BACKGROUND_COLOR,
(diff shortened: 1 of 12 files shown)


packages/scene-composer/src/store/helpers/serializationHelpers.ts
```diff
@@ -573,6 +585,53 @@ function createDocumentState(
...
+      backgroundSettings.color = backgroundSettings.color
+        ? isValidHexCode(backgroundSettings.color)
+          ? backgroundSettings.color
+          : DEFAULT_SCENE_BACKGROUND_COLOR
+        : undefined;
+      scene.properties[KnownSceneProperty.SceneBackgroundSettings] = backgroundSettings;
+    }
+    if (scene.properties[KnownSceneProperty.GroundPlaneSettings]) {
+      const groundPlaneSettings = scene.properties[KnownSceneProperty.GroundPlaneSettings] as IGroundPlaneSettings;
+      groundPlaneSettings.color = groundPlaneSettings.color
```
mumanity, line 624: I don't think you need to nest this. If it's `undefined` wouldn't you want to return the default color? 
```
groundPlaneSettings.color = isValidHexCode(groundPlaneSettings.color)
          ? groundPlaneSettings.color
          : DEFAULT_GROUND_PLANE_COLOR

haweston: No, it's allowed to be undefined because there may be a texture assigned instead of a color and that's okay. So it's only invalid if we try to assign a color that isn't a color.

mumanity: Gracias.

mumanity (approved):

PR #2734, opened 2024-04-11
Fix opacity rounding

haweston:
## Overview
At certain numbers the ground plane opacity will show long decimal precision when the +/- 1 clicker is used. This is due to floating point precision because the percentage is stored as 1 to 0 in the file and where it's used in the three.js.  We want to give users whole percents when they use the clicker and have applied a fixed rounding to zero decimal points for the opacity input widget.

## Verifying Changes
run story book
go to scene settings
click the opacity settings for the ground plane up and down from 100 to zero and make sure that only whole number percents are shown.  Try typing a decimal point and see if the input is rounded to a whole percent.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/store/helpers/serializationHelpers.ts
+++ b/packages/scene-composer/src/store/helpers/serializationHelpers.ts
@@ -10,7 +10,18 @@ import {
   LEGACY_VERSION,
 } from '../../common/constants';
 import { ERROR_MESSAGE_DICT, ErrorCode, ErrorLevel, SceneComposerRuntimeError } from '../../common/errors';
-import { KnownComponentType } from '../../interfaces';
+import {
+  KnownComponentType,
+  KnownSceneProperty,
+  IFogSettings,
+  IGroundPlaneSettings,
+  ISceneBackgroundSetting,
+  DEFAULT_FOG_COLOR,
+  DEFAULT_FOG_NEAR,
+  DEFAULT_FOG_FAR,
+  DEFAULT_SCENE_BACKGROUND_COLOR,
+  DEFAULT_GROUND_PLANE_COLOR,
+} from '../../interfaces';
 import DebugLogger from '../../logger/DebugLogger';
 import { Component, DistanceUnit, ModelType, Node, Scene } from '../../models/SceneModels';
 import { isDynamicNode } from '../../utils/entityModelUtils/sceneUtils';
@@ -36,6 +47,7 @@ import {
   isISceneComponentInternal,
   IPlaneGeometryComponentInternal,
 } from '../internalInterfaces';
+import { isValidHexCode } from '../../utils/colorUtils';
 
 import { addComponentToComponentNodeMap } from './componentMapHelpers';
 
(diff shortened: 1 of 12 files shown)


packages/scene-composer/src/store/helpers/serializationHelpers.ts
```diff
@@ -573,6 +585,53 @@ function createDocumentState(
...
+    if (scene.properties[KnownSceneProperty.BackgroundCustomColors]) {
+      scene.properties[KnownSceneProperty.BackgroundCustomColors] = (
+        scene.properties[KnownSceneProperty.BackgroundCustomColors] as string[]
+      ).filter((color) => isValidHexCode(color));
+    }
+    if (scene.properties[KnownSceneProperty.GroundCustomColors]) {
+      scene.properties[KnownSceneProperty.GroundCustomColors] = (
+        scene.properties[KnownSceneProperty.GroundCustomColors] as string[]
+      ).filter((color) => isValidHexCode(color));
+    }
```
mitchlee-amzn, line 603: These all do the same thing, and you could simplify this code. You can likely remove a lot of conditionals if you instead setup a default state.

For instance, instead of checking `if scene.properties`, initialize it.

It's odd that this logic is mutating the original scene object, it would be cleaner if the initial shape of the scene.properties was a known state/structure, and that the mutation happens in the store, instead of correcting it on save. But even barring that, if we need to map it to a different structure for writing, it should be copied instead of mutated:

```
const persistedscene = Object.keys(scene).reduce((persist, property) => {
      persist[property] = property || [] as String[]; // basic example to show
}, {});

return JSON.serialize(persistedscene);
```

packages/scene-composer/src/utils/colorUtils.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { isValidHexCode } from './colorUtils';
```
mitchlee-amzn, line 1: nit: Do we have our own way of doing this? ThreeJS has this this built in: https://threejs.org/docs/#api/en/math/Color

We shouldn't need to have tests for color conversion or validity, we should rely on a library that handles this for us.

packages/scene-composer/src/utils/colorUtils.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * This method uses a regular expression (`hexRegex`) to validate a hex color code.
+ * The regex checks if the hex code starts with a "#" symbol, followed by either a
+ * 6-digit combination of characters from A-F, a-f, and 0-9.
+ * The `test` method is then used to validate the `hexCode` against the regex pattern.
+ * @param hexCode
+ * @returns
+ */
+export const isValidHexCode = (hexCode: string): boolean => {
```
mitchlee-amzn, line 9: In general we should let ThreeJS handle colors, it provides a more flexible interface and allows for color values to be more than just hex codes (less rigid). See previous comment.

Issue #2668, opened 2024-03-08
[Widgets] Support Cloudscape theming options


tracy-french:
This issue is dependent on https://github.com/awslabs/iot-app-kit/issues/2667.

Note: This issue will be updated with additional theming options as we understand which we need.

Extend `useTheme` to support additional Cloudscape theming options. Please update the type `SupportedCloudscapeDesignTokens` to the following:

```ts
type SupportedCloudscapeDesignTokens = Pick<
  Theme['tokens'],
  | 'borderRadiusButton'
  | 'borderRadiusContainer'
  | 'borderRadiusControlCircularFocusRing'
  | 'borderRadiusControlDefaultFocusRing'
  | 'borderRadiusDropdown'
  | 'borderRadiusInput'
  | 'borderRadiusItem'
  | 'borderRadiusToken'
  | 'colorBackgroundButtonNormalActive'
  | 'colorBackgroundButtonNormalDefault'
  | 'colorBackgroundButtonNormalDisabled'
  | 'colorBackgroundButtonNormalHover'
  | 'colorBackgroundButtonPrimaryActive'
  | 'colorBackgroundButtonPrimaryDefault'
  | 'colorBackgroundButtonPrimaryDisabled'
  | 'colorBackgroundButtonPrimaryHover'
  | 'colorBackgroundControlChecked'
  | 'colorBackgroundControlDefault'
  | 'colorBackgroundControlDisabled'
  | 'colorBackgroundDropdownItemDefault'
  | 'colorBackgroundDropdownItemFilterMatch'
  | 'colorBackgroundDropdownItemHover'
  | 'colorBackgroundInputDefault'
  | 'colorBackgroundInputDisabled'
  | 'colorBackgroundItemSelected'
  | 'colorBorderButtonNormalActive'
  | 'colorBorderButtonNormalDefault'
  | 'colorBorderButtonNormalDisabled'
  | 'colorBorderButtonNormalHover'
  | 'colorBorderButtonPrimaryDisabled'
  | 'colorBorderDropdownItemFocused'
  | 'colorBorderDropdownItemHover'
  | 'colorBorderInputDefault'
  | 'colorBorderInputFocused'
  | 'colorBorderItemFocused'
  | 'colorBorderItemSelected'
  | 'colorForegroundControlDefault'
  | 'colorForegroundControlDisabled'
  | 'colorTextAccent'
  | 'colorTextBodyDefault'
  | 'colorTextBodySecondary'
[...]

PR #2715, opened 2024-04-03
fix: migrated tooltip css to styled component for customizing theming

Kausar-HM:
Added dark mode support for Tooltip component as part of #2668 

How to test changes?
Use theming hook in `packages/dashboard/src/components/widgets/list.tsx`
```
import { useTheme } from '../../../../../packages/react-components/src/hooks/useTheming';
useTheme({
    mode: 'dark',
  });
```
Import Tooltip using relative path in below components: ` import { Tooltip } from '../../../../react-components/src/components/tooltip';`
```
packages/dashboard/src/components/internalDashboard/collapsiblePanel.tsx
packages/dashboard/src/components/internalDashboard/dashboardHeader.tsx
packages/dashboard/src/customization/propertiesSections/propertiesAndAlarmsSettings/styledPropertyComponent.tsx
packages/react-components/src/components/time-sync/timeSelection.tsx
```

### Verifying changes:
Tooltip UI in RC:
[tooltip-UI-RC.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/e0e92642-47c8-4e42-85be-f69dbefd57bf)

Tooltip UI after this PR:
[tooltip-dark-mode-changes.webm](https://github.com/awslabs/iot-app-kit/assets/154328164/ff6c3ba0-1f67-429e-9fbc-28a20a9fc4e0)

Also, the border color had to be changed to a lighter shade than before, since we needed that in dark mode to match with other control's border color. We have used this design token `colorBorderControlDefault`
Here are before after screenshots for reference:
image 
image image

PR #2739, opened 2024-04-16
fix: migrated widget tooltip css to styled-components for theming #2668

Kausar-HM:
As part of theming issue #2668 , fixed the UI of widget panel tooltip to support dark mode and theming options

### Verifying changes:

**BEFORE** (RC)
Light mode 
image

Dark mode
image
---

### AFTER
Light mode
image

Dark mode
image

### DIFF
--- a/packages/dashboard/src/components/palette/icons/index.tsx
+++ b/packages/dashboard/src/components/palette/icons/index.tsx
@@ -1,15 +1,16 @@
 import React, { useState, type DragEventHandler, useEffect } from 'react';
 import Box from '@cloudscape-design/components/box';
 import {
-  colorBackgroundLayoutMain,
   colorBorderButtonNormalDisabled,
-  colorTextButtonNormalActive,
-  spaceScaledS,
-  spaceScaledM,
-  spaceStaticXs,
-  spaceScaledXxxs,
+  colorBackgroundSegmentHover,
+  borderRadiusPopover,
+  fontSizeHeadingS,
+  spaceStaticM,
+  spaceStaticS,
+  colorTextBodyDefault,
 } from '@cloudscape-design/design-tokens';
 import './index.css';
+import styled from 'styled-components';
 
 type PaletteComponentIconProps = {
   Icon: React.FC;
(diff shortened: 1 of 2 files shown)


gmuralidharreddy:
lgtm

packages/dashboard/src/components/palette/icons/index.tsx
```diff
@@ -21,16 +22,46 @@ const PaletteComponentIcon: React.FC<PaletteComponentIconProps> = ({
...
-    fontSize: spaceScaledM,
-    color: colorTextButtonNormalActive,
-    backgroundColor: colorBackgroundLayoutMain,
-    padding: `${spaceScaledS} ${spaceScaledM}`,
-    borderRadius: spaceStaticXs,
-    borderWidth: spaceScaledXxxs,
-    borderColor: colorBorderButtonNormalDisabled,
-    boxShadow: `${spaceScaledXxxs} ${spaceScaledXxxs} ${spaceScaledXxxs} ${colorBorderButtonNormalDisabled}`,
-  };
+  const Tooltip = styled.div<{ hover: boolean }>`
```
corteggiano, line 25: Is there a reason we're not reusing the component exported from `packages/react-components/src/components/tooltip/tooltip.tsx` ?  I dont see why we have to duplicate a similar tooltip that was already done in this PR https://github.com/awslabs/iot-app-kit/pull/2715

corteggiano (changes requested):

packages/dashboard/src/components/palette/icons/index.tsx
Kausar-HM: I made a copy of this branch to try out the approach for reusing `packages/react-components/src/components/tooltip/tooltip.tsx` - https://github.com/awslabs/iot-app-kit/pull/2747

we can reuse this component, but it is not keyboard accessible - on pressing escape key, tooltip won't disappear.
This is existing behavior for this component, when we tried to fix it using a `hover` state and registered an event handler for handling escape, somehow it is lagging in that component.
Another issue is, if you rapidly hover over the widget icons, some of the previously visible tooltips won't disappear.
After spending some more time, we still didn't find the reason behind this.

Can we merge this PR for dark mode / theming support for now? with these changes, it is still accessible(dismiss using esc).

**To improve the tooltip:**
Can we not use some library like react-tooltip which has support for props like `closeOnEsc`?
Or rebuild the tooltip component in such a way that it is accessible?

corteggiano: agreed that its better to get this theming support out quickly! thank you for investigating further 🙏 i will add a task to our backlog to consolidate the tooltip implementations!

corteggiano (approved):

PR #2740, opened 2024-04-16
feat: resource explorers

tracy-french:
## Overview
This change includes the initial versions of the new resource explorer components. They are not 100% complete, as we still need documentation, a bug bash, a security review, complete internationalization, and further theming capabilities. They are marked with an `@experimental` doc comment to prevent usage. The remaining items will be resolved as follow-ups. Once resolved, we will move forward with an official release.

## Legal

### DIFF
--- a/packages/react-components/src/components/resource-explorers/types/request-fn.ts
+++ b/packages/react-components/src/components/resource-explorers/types/request-fn.ts
@@ -1,0 +1,144 @@
+import type {
+  BatchGetAssetPropertyValueRequest,
+  BatchGetAssetPropertyValueResponse,
+  ExecuteQueryRequest,
+  ExecuteQueryResponse,
+  ListAssetModelsRequest,
+  ListAssetModelsResponse,
+  ListAssetModelPropertiesRequest,
+  ListAssetModelPropertiesResponse,
+  ListAssetsRequest,
+  ListAssetsResponse,
+  ListAssociatedAssetsRequest,
+  ListAssociatedAssetsResponse,
+  ListAssetPropertiesRequest,
+  ListAssetPropertiesResponse,
+  ListTimeSeriesRequest,
+  ListTimeSeriesResponse,
+} from '@aws-sdk/client-iotsitewise';
+import type { RequestTimeout } from './common';
+
+/** First-class function used to send requests to AWS. */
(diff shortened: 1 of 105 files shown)


packages/react-components/src/components/resource-explorers/types/resource-explorer.ts
```diff
@@ -0,0 +1,96 @@
...
+} from './common';
+import {
+  DropDownResourceDefinition,
+  ResourceDropDownSettings,
+} from './drop-down';
+
+/** Props common to all resource explorers. */
+export type CommonResourceExplorerProps<Resource> = {
+  /** TODO */
+  requestFns?: unknown;
```
jmbuss, line 22: Is this type unknown because it's a commonly used interface across the more specific explorer interfaces?

packages/react-components/src/components/resource-explorers/types/request-fn.ts
```diff
@@ -0,0 +1,144 @@
...
+export type RequestFunction<Request, Response> = (
+  request: Request,
+  options?: {
+    abortSignal?: AbortSignal;
+    requestTimeout?: RequestTimeout;
+  }
+) => PromiseLike<Response>;
+
+/** Utility type to ensure the correct structure of `requestFns`. */
+export type RequestFunctions<T> = T extends Partial<
```
jmbuss, line 31: Not saying we should move these as part of this PR, but these might be useful types to have for other systems that need to make requests. Maybe it should be in a common types in react-components?

packages/react-components/src/components/resource-explorers/types/resource-explorer.ts
tracy-french: Pretty much. This is a little trick to ensure every resource explorer has a requestFns prop defined. The specific explorers override the value. It's unknown here because it's unusable internally without overriding.

packages/react-components/src/components/resource-explorers/types/request-fn.ts
tracy-french: Completely agree. We should have all the components use these types (where it makes sense, of course).

jmbuss (approved):
Nice work!

jmbuss (approved):
Nice work!

Issue #2635, opened 2024-02-28
[Gauge] - Able to add Gauge Widget to Dashboard 


corteggiano:
| :exclamation:  PLEASE MAKE THESE CHANGES UNDER A FEATURE FLAG  | 
|-----------------------------------------|

| :exclamation:  PLEASE MAKE THESE CHANGES IN SMALL INCREMENTAL PRs AND NOT AT ONCE  | 
|-----------------------------------------|

Create a feature flag by setting a local_storage value and referencing it as this:
`const hasGauge = !!localStorage?.getItem('HAS_GAUGE_WIDGET');` 

Implementation Steps:
1. Add the following to `iot-app-kit/packages/dashboard/src/customization/widgets/types.ts`
```
export type GaugeProperties = QueryProperties & {
  gaugeThickness?: number;
  showName?: boolean;
  showUnit?: boolean;
  fontSize?: number;
  labelFontSize?: number;
  unitFontSize?: number;
  yMin?: number;
  yMax?: number;
};

export type GaugeWidget = DashboardWidget<GaugeProperties>;
```
1. Create `Gauge` folder under `iot-app-kit/packages/dashboard/src/customization/widgets`
2. In the folder, you will need the implement the following files (take a look at KPI for an idea of what to do for each file):
    * `component.tsx` - this is where you will use the react-component implementation
    * `icon.tsx` - Represents the icon that will be shown in the widget palette
    * `plugin.tsx` - Represents the initial state of all the settings when the widget is added to a dashboard
                              Initial states should include an `initialSize, showName: true,  showUnit: true`

3. Users should be able to add the gauge widgets to the dashboard
    * Make sure you cannot add more than 1 property to a widget (we already do this for KPI so it should be the same behavior)

```[tasklist]
### Dial Acceptance criteria
- [ ] User can add a gauge component to dashboard
- [ ] User cannot add more than one property to a gauge widget
```

jmbuss:
![Gauge (1)](https://github.com/awslabs/iot-app-kit/assets/107281089/87b30ec1-abcd-49b8-af61-9d4c37d23e93)

svg for the component library section.

corteggiano:
here are the light and dark mode SVGs
![gauge-light](https://github.com/awslabs/iot-app-kit/assets/28601414/c2a928b9-14fc-438c-87be-0db564af8582)
[...]

Issue #2634, opened 2024-02-28
[Gauge] - Customize Gauge Widget in Dashboard


corteggiano:
| :exclamation:  PLEASE MAKE THESE CHANGES UNDER A FEATURE FLAG  | 
|-----------------------------------------|

| :exclamation:  PLEASE MAKE THESE CHANGES IN SMALL INCREMENTAL PRs AND NOT AT ONCE  | 
|-----------------------------------------|

Create the config panels for Gauge, they should have the following sections
    **Styles**
        * Show/hide the unit
        * Show/hide property name
        * Change the y min and y max
        * Change the aggregation and resolution
        * Change font sizes
    **Properties**
        * Can remove a property from here
        * Can change the widget’s color from the color picker
    **Thresholds**
        * Can create a threshold
        * Can remove a threshold
        * Can change a threshold's color and value

<img width="724" alt="image" src="https://github.com/awslabs/iot-app-kit/assets/28601414/d8b8eb47-663c-4c68-9f4d-107a90d6812d">

 
```[tasklist]
### Acceptance criteria
- [ ] User can show/hide unit
- [ ] Users can show/hide property name
- [ ] Users can change the ymin and ymax of the widget
- [ ] Users can change the aggregation and resolution of the widget
- [ ] Users can change the widget property's color
- [ ] Users can remove properties
- [ ] Users can create thresholds
- [ ] Users can remove thresholds
- [ ] Users can customize thresholds
- [ ] Users can customize font sizes
```

corteggiano:
the resolution and aggregation are part of the Gauge widget’s query prop and should be updated using the existing dropdown found at `dashboard/src/customization/propertiesSections/aggregationSettings/index.tsx`

PR #2746, opened 2024-04-23
feat: add gauge component to dashboard

gmuralidharreddy:
This PR is for the ticket #2635 

Now with these changes, we can able to add a Guage component in the dashboard. Since the gauge is to occupy a full widget, I made a few changes to the component.

Verify:

https://github.com/awslabs/iot-app-kit/assets/139960352/fdf2f382-495d-475a-be4c-23ddb53247cb


packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.ts
```diff
@@ -28,6 +28,23 @@ export const disableAdd = (
       }
       break;
     }
+    case 'gauge': {
```
jmbuss, line 31: is this different than the 2 cases above?

packages/react-components/src/components/gauge/gaugeDataQualityText.tsx
```diff
@@ -6,20 +6,15 @@ import { DataQualityText } from '../data-quality/data-quality-text';
...
 }) => {
   if (error || !quality) {
     return null;
   }
   return (
-    <div
-      className='gauge-data-quality'
-      style={{ bottom: showName ? '36%' : '40%' }}
-    >
+    <div className='gauge-data-quality' style={{ bottom: '2%' }}>
```
jmbuss, line 17: since we are using percentages, can you add some screenshots of this widget on the dashboard at various sizes. small / medium / large square, and then similarly, small / medium / large rectangles where we test short but wide rectangles / tall but skinny rectangles.

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/footer/disableAdd.ts
gmuralidharreddy: Sorry, both are same. Resolved

corteggiano:
it looks great locally! thank you for the work
i noticed a couple of issues:
1. dragging the widget around the dashboard is pretty buggy/inconsistent. the widget should be draggable from anywhere on the tile and should perform as well as the other widgets. ( i think this is related to the tooltip bug which is fixed so try rebasing from RC branch)
2. widget title does not appear on widget 
image

3. properties color does not match gauge color and changing the property color doesnt change gauge color
image

4. cant add thresholds
image

gmuralidharreddy:
> it looks great locally! thank you for the work i noticed a couple of issues:
> 
> 1. dragging the widget around the dashboard is pretty buggy/inconsistent. the widget should be draggable from anywhere on the tile and should perform as well as the other widgets. ( i think this is related to the tooltip bug which is fixed so try rebasing from RC branch)
> 2. widget title does not appear on widget
> 
> image
> 3. properties color does not match gauge color and changing the property color doesnt change gauge color
> 
> image
> 4. cant add thresholds
> 
> image

1. Rebased from RC.
2. 2,3 & 4 These three are going to be done as part of Customization in the config panel ticket: (https://github.com/awslabs/iot-app-kit/issues/2634)

PR #2753, opened 2024-04-29
fix(scene composer): correcting state logic for tag settings

mumanity:
## Overview
Cleaning up state management for tag settings panel. 

## Verifying Changes
Before

https://github.com/awslabs/iot-app-kit/assets/6610619/6d508738-2fb3-430b-9404-78e25fdd346a

After

https://github.com/awslabs/iot-app-kit/assets/6610619/3933210c-9508-40bf-a1f4-37fb29ed0990

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.spec.tsx
+++ b/packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.spec.tsx
@@ -98,26 +98,6 @@ describe('SceneTagSettingsEditor', () => {
     });
   });
 
-  it('should update input value when store value changed', async () => {
-    useStore('default').setState(baseState);
-    const { container, rerender } = render(<SceneTagSettingsEditor />);
-    const polarisWrapper = wrapper(container);
-    const input = polarisWrapper.findInput();
-
-    expect(input).toBeDefined();
-    expect(input?.findNativeInput().getElement().value).toEqual(String(DEFAULT_TAG_GLOBAL_SETTINGS.scale));
-
-    (useTagSettings as jest.Mock).mockReturnValue({
-      ...DEFAULT_TAG_GLOBAL_SETTINGS,
-      scale: 6,
-    });
-    act(() => {
-      rerender(<SceneTagSettingsEditor />);
-    });
-
-    expect(polarisWrapper.findInput()?.findNativeInput().getElement().value).toEqual('6');
-  });
-
   it('should update store when checkbox clicked', async () => {
     useStore('default').setState(baseState);
     const { container } = render(<SceneTagSettingsEditor />);
@@ -169,9 +149,6 @@ describe('SceneTagSettingsEditor', () => {
       // update slider value should update store
       fireEvent.change(slider[0], { target: { value: '3' } });
     });
-
-    // set value only after mouse up
-    expect(setScenePropertyMock).not.toBeCalled();
 
     act(() => {
       fireEvent.mouseUp(slider[0]);
--- a/packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
+++ b/packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
@@ -19,12 +19,18 @@ export const SceneTagSettingsEditor: React.FC = () => {
   const isViewing = useStore(sceneComposerId)((state) => state.isViewing());
   const setTagSettings = useViewOptionState(sceneComposerId).setTagSettings;
   const tagSettings: ITagSettings = useTagSettings();
-  const [dirty, setDirty] = useState(false);
   const [focusInput, setFocusInput] = useState(false);
   const [focusSlider, setFocusSlider] = useState(false);
   const [draggingSlider, setDraggingSlider] = useState(false);
-
   const [internalScale, setInternalScale] = useState(tagSettings.scale);
+
+  const handleInputFocus = useCallback(() => setFocusInput(true), []);
+  const handleInputBlur = useCallback(() => setFocusInput(false), []);
+
+  const handleSliderFocus = useCallback(() => setFocusSlider(true), []);
+  const handleSliderBlur = useCallback(() => setFocusSlider(false), []);
+  const handleSliderMouseUp = useCallback(() => setDraggingSlider(false), []);
+  const handleSliderMouseDown = useCallback(() => setDraggingSlider(true), []);
 
   const updateSettings = useCallback(
     (settingsPartial: Partial<ITagSettings>) => {
@@ -37,7 +43,6 @@ export const SceneTagSettingsEditor: React.FC = () => {
           ...getSceneProperty(KnownSceneProperty.ComponentSettings),
           [KnownComponentType.Tag]: newTagSettings,
         };
-
         setSceneProperty(KnownSceneProperty.ComponentSettings, newComponentSettings);
       }
       setTagSettings(newTagSettings);
@@ -48,20 +53,9 @@ export const SceneTagSettingsEditor: React.FC = () => {
   const onSliderChange = useCallback(
     (event) => {
       setInternalScale(Number(event.target.value));
-      setDirty(true);
     },
-    [setInternalScale, setDirty],
+    [internalScale],
   );
-
-  const onInputBlur = useCallback(() => {
-    // If slider is getting focus, this makes sure to execute setFocusSlider(true) first to keep showing it.
-    const id = setTimeout(() => {
-      setFocusInput(false);
-    }, 1);
-    return () => {
-      clearTimeout(id);
-    };
-  }, [setFocusInput]);
 
   const onInputChange = useCallback(
     (event) => {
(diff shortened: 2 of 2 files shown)


packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.spec.tsx
```diff
@@ -98,26 +98,6 @@ describe('SceneTagSettingsEditor', () => {
     });
   });
 
-  it('should update input value when store value changed', async () => {
```
mumanity, line 101: CALLOUT
This test failed since the `tagSettings` aren't triggering a stateful update when the document is updated. Refresh behavior works fine & I don't believe this impacts existing prod state, however, would like to confirm that this is non-intrusive to dynamic scene work. 

Would ideally like to bring this back if possible.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -19,7 +19,6 @@ export const SceneTagSettingsEditor: React.FC = () => {
   const isViewing = useStore(sceneComposerId)((state) => state.isViewing());
   const setTagSettings = useViewOptionState(sceneComposerId).setTagSettings;
   const tagSettings: ITagSettings = useTagSettings();
-  const [dirty, setDirty] = useState(false);
```
mumanity, line 22: Removing to let React handle state management.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.spec.tsx
```diff
@@ -170,9 +150,6 @@ describe('SceneTagSettingsEditor', () => {
       fireEvent.change(slider[0], { target: { value: '3' } });
     });
 
-    // set value only after mouse up
```
mumanity, line 173: Cleaned up since the state no longer waits for `onBlur` to trigger update.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -37,62 +36,38 @@ export const SceneTagSettingsEditor: React.FC = () => {
           ...getSceneProperty(KnownSceneProperty.ComponentSettings),
           [KnownComponentType.Tag]: newTagSettings,
         };
-
         setSceneProperty(KnownSceneProperty.ComponentSettings, newComponentSettings);
       }
       setTagSettings(newTagSettings);
     },
-    [tagSettings, getSceneProperty, setSceneProperty],
+    [internalScale, tagSettings, getSceneProperty, setSceneProperty],
```
hwandersman, line 43: `internalScale` is not used in this function

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -37,62 +36,38 @@ export const SceneTagSettingsEditor: React.FC = () => {
...
   );
 
   const onSliderChange = useCallback(
     (event) => {
       setInternalScale(Number(event.target.value));
-      setDirty(true);
     },
-    [setInternalScale, setDirty],
+    [internalScale, tagSettings],
```
hwandersman, line 50: `tagSettings` is not used in this function

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -37,62 +36,38 @@ export const SceneTagSettingsEditor: React.FC = () => {
...
         value = 0;
       }
-
       if (value !== internalScale) {
         setInternalScale(Number(value));
-        setDirty(true);
       }
     },
-    [setInternalScale, setDirty],
+    [internalScale, tagSettings],
```
hwandersman, line 63: Why do you add `tagSettings` to all of these useCallback functions as a dependency?

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
mitchlee-amzn: nit: I am pretty sure there are lint rules to do this, can we have a task at least to go turn those on so we don't need to rely on Harry to tell us? 🥇

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -37,62 +36,38 @@ export const SceneTagSettingsEditor: React.FC = () => {
...
-    }
-  }, [updateSettings, setDirty, dirty, focusInput, internalScale, draggingSlider, isViewing]);
-
-  // Update internal when scale in store is changed
-  useEffect(() => {
-    if (!dirty && tagSettings.scale !== internalScale) {
-      setInternalScale(tagSettings.scale);
     }
-  }, [dirty, draggingSlider, tagSettings.scale, internalScale]);
+  }, [internalScale, tagSettings]);
```
mitchlee-amzn, line 70: I don't think you need this.

"If internal scale doesn't equal scale, set it to scale" is the same as "set internal scale to scale". You don't need a condition, therefore you don't need a check, and you don't need the useEffect at all.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -103,7 +78,7 @@ export const SceneTagSettingsEditor: React.FC = () => {
               value={String(internalScale)}
               type='number'
               onFocus={() => setFocusInput(true)}
-              onBlur={onInputBlur}
+              onBlur={() => setFocusInput(false)}
```
mitchlee-amzn, line 81: You should useCallback, anonymous functions in this case get redefined in every render, creating additional waste for the GC.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
mumanity: This monitors `internalScale` for updates & only triggers the update to settings on scene properties if there is an appropriate change to be made. This covers updates via input, slider, & keystroke. It is necessary to keep them all in sync.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
mumanity: Tested this suggestion with the snippet below & it crashed the app due to excessive rerenders. Guessing that I misunderstood your suggestion? 

`onBlur={useCallback(() => setFocusInput(false), [focusInput])}`

mitchlee-amzn: Yeah it needs to be named:

```
const handleBlur = useCallback(() => setFocusInput(false), []);

onBlur={handleBlur}
```

mumanity: And clearly I missed the established pattern that you were referring to... fixing now.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.spec.tsx
mitchlee-amzn: Yeah, this is a dumb test.

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
mumanity: Linter implemented in a separate effort (#2779)

packages/scene-composer/src/components/panels/scene-settings/SceneTagSettingsEditor.tsx
```diff
@@ -19,13 +19,19 @@ export const SceneTagSettingsEditor: React.FC = () => {
...
   const [focusInput, setFocusInput] = useState(false);
   const [focusSlider, setFocusSlider] = useState(false);
   const [draggingSlider, setDraggingSlider] = useState(false);
-
   const [internalScale, setInternalScale] = useState(tagSettings.scale);
 
+  const handleInputFocus = useCallback(() => setFocusInput(true), []);
+  const handleInputBlur = useCallback(() => setFocusInput(false), []);
+
+  const handleSliderFocus = useCallback(() => setFocusSlider(true), []);
```
mumanity, line 30: Unfortunately we need all of these state managers. `mouseUp` & `mouseDown` are required in order for the slider to work & `focus`/`blur` are necessary for each independently as well.

mitchlee-amzn (commented):
LGTM

PR #2767, opened 2024-05-02
feat: ds ga

cshang2017:
1. enable loading query results with child nodes.
2. remove dependent layer concept.

## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/StateManager.tsx
+++ b/packages/scene-composer/src/components/StateManager.tsx
@@ -108,6 +108,7 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
     () => createStandardUriModifier(sceneContentUri || '', undefined),
     [sceneContentUri],
   );
+  const [sceneRootEntityId, setSceneRootEntityId] = useState(null);
 
   const isViewing = config.mode === 'Viewing';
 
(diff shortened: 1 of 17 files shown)


packages/scene-composer/src/components/StateManager.tsx
```diff
@@ -245,22 +246,21 @@ const StateManager: React.FC<SceneComposerInternalProps> = ({
...
 
       sceneMetadataModule
-        .getSceneEntity({ entityId: rootId })
+        .getSceneEntity({ entityId: sceneRootEntityId })
         .then((res) => {
-          const document = parseSceneCompFromEntity(res);
-          if (!document) {
-            throw new Error('Failed to parse scene metadata');
+          if (res?.components?.length) {
+            const document = parseSceneCompFromEntity(res);
```
haweston, line 258: So right now this executes and doesn't fail out until if(res?.components?.length)  realizes we have no scene component?  Does that still avoid the early return  on 269?

cshang2017: Original code assume there is scene Node entity.  It's hard to remove completely. So, the extra check is to ensure the sceneInfo is from scene.json, not from a sceneNode. I added this line after saw crash issue when updating a property of scene object in the composer.

haweston: Okay, looks like the setSceneContentUri on line 251 ensures we have a way to load scene content if there is no scene component on the scene entity.

haweston (approved):

PR #2769, opened 2024-05-02
feat(widgets): add name style settings for line/table + edit label in config panel

ssjagad:
## Overview
- Add name property to style settings object for non-line widgets
- Add name property style chart options for line chart
- created label editing component for config panel
- users can now edit names of individual properties from config panel or revert to original name

https://github.com/awslabs/iot-app-kit/assets/145582655/e5653326-5aed-4d6b-840e-6998ffc0e7e6

## Legal

### DIFF
--- a/packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
+++ b/packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
@@ -9,7 +9,7 @@ import {
   useGroupableEChart,
   useLoadableEChart,
 } from '../../../hooks/useECharts';
-import { ChartDataQuality, ChartOptions } from '../types';
+import { ChartDataQuality, ChartOptions, ChartStyleSettings } from '../types';
 import { useXAxis } from './axes/xAxis';
 import { useTooltip } from './tooltip/convertTooltip';
 import { useEffect, useMemo, useRef, useState } from 'react';
@@ -71,7 +71,8 @@ const toDataStreamIdentifiers = (dataStreams: DataStream[]) =>
 
 const toDataStreamMetaData = (
   datastreams: ReturnType<typeof toDataStreamIdentifiers>,
-  series: SeriesOption[]
+  series: SeriesOption[],
+  styleSettings?: ChartStyleSettings
 ) => {
   return datastreams.map(
     ({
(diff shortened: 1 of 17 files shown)


packages/react-components/src/components/chart/chartOptions/useChartConfiguration.ts
```diff
@@ -89,9 +90,15 @@ const toDataStreamMetaData = (
         ({ id: seriesId }) => seriesId === id
       ) ?? { appKitColor: color };
       const colorUsed = (foundSeries as GenericSeries).appKitColor;
+
+      let styledName;
+      if (refId && styleSettings) {
+        styledName = styleSettings[refId]?.name;
```
jmbuss, line 96: We can probably turn this into a ternary and get rid of some code!

packages/react-components/src/components/chart/baseChart.tsx
```diff
@@ -98,7 +98,7 @@ const BaseChart = ({
     thresholds,
     utilizedViewport,
     visibleData,
-  } = useVisualizedDataStreams(queries, viewport);
+  } = useVisualizedDataStreams(queries, viewport, undefined);
```
jmbuss, line 101: is it important to pass undefined here? Are we using this hook in any other components? Maybe we can just remove that param entirely?

jmbuss (commented):
Looks good to me! I think there are a few things commented out that you can just delete, and then I had a few questions.

packages/react-components/src/components/chart/baseChart.tsx
ssjagad: Nope, that should be removed since im not passing styleSettings as a parameter there

corteggiano (approved):

packages/react-components/src/components/chart/hooks/useVisualizedDataStreams.ts
```diff
@@ -30,13 +31,13 @@ export const useVisualizedDataStreams = (
...
     (lastUpdatedBy ? viewport : passedInViewport || viewport) ??
     DEFAULT_VIEWPORT;
-
   const { dataStreams, thresholds: queryThresholds } = useTimeSeriesData({
     viewport: utilizedViewport,
     queries,
     settings: {
       fetchFromStartToEnd: true,
     },
+    // styles: styleSettings,
```
jmbuss, line 40: 1 more comment :D

packages/react-components/src/components/chart/hooks/useVisualizedDataStreams.ts
```diff
@@ -11,6 +11,7 @@ import isEqual from 'lodash.isequal';
 import { useTimeSeriesData } from '../../../hooks/useTimeSeriesData';
 import { useViewport } from '../../../hooks/useViewport';
 import { DEFAULT_VIEWPORT, StreamType } from '../../../common/constants';
+// import { ChartStyleSettings } from '../types';
```
jmbuss, line 14: A little cleanup here :D

PR #2771, opened 2024-05-03
fix(react-components): add error state

corteggiano:
## Overview
Handle other states -- empty dataset, loading, failed, and error

Screenshot 2024-05-03 at 09 33 11

## Legal


packages/react-components/src/components/anomaly-widget/index.tsx
```diff
@@ -37,6 +41,27 @@ export const AnomalyWidget = (options: AnomalyWidgetOptions) => {
     loading,
   });
 
+  if (error) {
+    return (
```
jmbuss, line 45: can we make this it's own component?

jmbuss (approved):

PR #2778, opened 2024-05-06
feat: ds ga

cshang2017:
1. enable drag and drop.

## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
+++ b/packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
@@ -4,7 +4,7 @@ import { HTML5Backend } from 'react-dnd-html5-backend';
 import { isEmpty } from 'lodash';
 
 import { useSceneComposerId } from '../../../common/sceneComposerIdContext';
-import { findComponentByType, getFinalNodeTransform } from '../../../utils/nodeUtils';
+import { findComponentByType, getRelativeTransform } from '../../../utils/nodeUtils';
 import { ISceneNodeInternal, useNodeErrorState, useSceneDocument, useStore } from '../../../store';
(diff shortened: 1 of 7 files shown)


packages/scene-composer/src/components/panels/SceneHierarchyPanel/SceneHierarchyDataProvider.tsx
```diff
@@ -221,12 +221,12 @@ const SceneHierarchyDataProvider: FC<SceneHierarchyDataProviderProps> = ({ selec
...
           hierarchyParent = getSceneNodeByRef(hierarchyParent.parentRef);
         }
-        newHierarchyParentObject = getObject3DBySceneNodeRef(hierarchyParent?.ref);
+        newParentObject3D = getObject3DBySceneNodeRef(hierarchyParent?.ref);
       }
 
       let maintainedTransform: ITransform | null = null;
       if (originalObject3D) {
-        maintainedTransform = getFinalNodeTransform(originalObject, originalObject3D, newHierarchyParentObject);
+        maintainedTransform = getRelativeTransform(originalObject!, newParent!, getSceneNodeByRef);
```
haweston, line 229: So this shouldn't be the relative transform (assuming relative equal nodes local transform)? This is instead the transform need to keep the object in the same place even though it has a new parent.   This function was working before for moving nodes and should still work for dynamic entities as written.  The issue is that https://github.com/awslabs/iot-app-kit/blob/main/packages/scene-composer/src/components/panels/SceneHierarchyPanel/components/SceneHierarchyTree/SceneHierarchyTreeItem.tsx#L113 prevents the drag from happening if it's dynamic. But it looks like our loader isn't setting the reserved layer id on scene load so isDynamicNode is returning false which lets you drag anyways...

PR #2779, opened 2024-05-06
fix(scene composer): implemented react-hooks linter & converted useSt…

mumanity:
…ore to accessStore

## Overview
Implemented the `react-hook` linter for scene-composer package & updated the naming convention of `useStore` to `accessStore` to adhere to React best practices. 

- `useStore` has been updated to `accessStore` (name change only)
- updating `jest` & adding the new required dependency of `jest-environment-jsdom`
- updating `jest.config.ts` as the `globals` prop was deprecated with the `jest` update

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/package.json
```diff
@@ -46,7 +46,7 @@
     "convert-svg": "npx @svgr/cli --out-dir src/assets/auto-gen/icons/ --typescript --index-template tools/index-template.js -- src/assets/icons/",
     "release": "run-s compile copy-assets",
     "copy-assets": "copyfiles -e \"**/*.tsx\" -e \"**/*.ts\" -e \"**/*.snap\" -e \"**/*.js\" -e \"**/*.jsx\" -e \"**/*.json\" \"src/**/*\" dist/",
-    "lint": "eslint . --max-warnings=485",
+    "lint": "eslint . --max-warnings=712",
```
hwandersman, line 49: Isn't this supposed to decrease the number of warnings? Why are we increasing the limit here?

packages/scene-composer/src/__snapshots__/SceneViewer.spec.tsx.snap
```diff
@@ -1,25 +1,12 @@
...
-}
-
-.c0 canvas {
-  outline: none;
-  -webkit-tap-highlight-color: rgba(255,255,255,0);
-}
-
 <div
-  className="c0 sceneViewer"
+  className="sc-cWSHoV icqfWN sceneViewer"
```
hwandersman, line 5: Is this going to lead to inconsistent failures? The random classnames have caused issues in the past

packages/scene-composer/src/components/toolbars/common/__snapshots__/ItemContainer.spec.tsx.snap
```diff
@@ -1,250 +1,40 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`ItemContainer should render properly and trigger onClick when clicking on item 1`] = `
-.c0 {
```
hwandersman, line 4: Randomly jumped to this file - looks like snapshots are not updating correctly. This content probably shouldn't be removed, and I see classnames are changed to something more random

packages/scene-composer/src/__snapshots__/SceneViewer.spec.tsx.snap
mumanity: I don't think it would cause any more issues than it did before as the previous class names were random.

hwandersman: It looks more deterministic now. Classes are named `c0`, then children are `c1`, etc.

packages/scene-composer/src/components/toolbars/common/__snapshots__/ItemContainer.spec.tsx.snap
mumanity: It's an intentional change from the jest version update to improve readability. We could re-enable it, however, since we use cloudscape I would suggest that we let them test their css. We're currently covered with e2e visual regression tests to ensure visual consistency.

packages/scene-composer/package.json
mumanity: We are adding a new linter that is increasing the warnings as we now include warnings from the increased coverage. While this might appear bad, it's actually really good to help us improve our code.

packages/scene-composer/src/__snapshots__/SceneViewer.spec.tsx.snap
mumanity: Are we opposed to trying it & seeing if it works before removing functionality that is designed to make us more light weight? I'd really like to only turn this off if it is determined to be a problem, especially since I think it's much more readable.

hwandersman: That's fine, I just want to make sure we have some confidence and testing done to see that it won't immediately cause failures in the future

hwandersman (approved):

packages/scene-composer/src/components/toolbars/common/__snapshots__/ItemContainer.spec.tsx.snap
mumanity: Validated locally that running the test for `SceneHierarchyTreeItem` six times did not create issues with mismatched snapshots.

mitchlee-amzn: This snapshot is problematic because it's shouldn't be rendering the cloudscape component at all, the snapshot should look something like: "<div data-mocked="Container"><StuffWeCareAbout /></div>".

So somethings not mocking right, which means our tests can fail just because cloudscape updates their side.

packages/scene-composer/package.json
mitchlee-amzn: It's not supposed to decrease the number of warnings, it's enforcing better practices in the code through increasing warnings. imo we should fix those, the point of adding the linter logic was to find other areas we may have issues.

PR #2783, opened 2024-05-06
feat: ds ga

cshang2017:
1. enable drag and drop.

## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/utils/entityModelUtils/updateNodeEntity.ts
+++ b/packages/scene-composer/src/utils/entityModelUtils/updateNodeEntity.ts
@@ -35,7 +35,7 @@ export const updateEntity = async (
 
   const nodecomp = updateNodeEntityComponent(node, undefined);
 
-  const updateEntity: UpdateEntityCommandInput = {
+  let updateEntity: UpdateEntityCommandInput = {
     workspaceId: undefined,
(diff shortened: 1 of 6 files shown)


packages/scene-composer/src/utils/entityModelUtils/updateNodeEntity.ts
```diff
@@ -44,6 +44,16 @@ export const updateEntity = async (
     },
   };
 
+  if (node.parentRef) {
```
haweston, line 47: I think this might need to be more robust?  If the node had a parentRef before but it's gone now we also need to do a parentEntityUpdte?  But also if the parentRef exists and hasn't changed then parentEntityUpdate is an extra API call that's redundant. Probably will require passing a little more info on old state to the function call though.

cshang2017: It is true that adding extra redundant during the api call. However, it does not need change updateEntity parameters, which has other places to be called.

cshang2017: will do a patch.

PR #2788, opened 2024-05-07
fix(scene): fix regression in updateSceneNode for reparenting nodes

haweston:
## Overview
Dynamic scene changes introduced a regression in updating scene nodes that change the nodes parent/child relationship such as when doing drag in drop. This impacts static & dynamic nodes equally.  

Observed behavior:  When you drag a node with a visual 3D element to a new parent the node is reparented in the scene hierarchy correctly but a second visual representing appears in the scene in the wrong location.

## Verifying Changes
Use a storybook scene with a hierarchy like
root -
  Model
     Model 2
         Tag
  Model 3

Drag Model 2 to be parented to Model 3.

No new models or tags should appear in the scene.  The existing models and tags should not change location.

You should be able to update the tag transform and only the tag changs.
You should be able to update the Model 2 transform and have both the Model 2 and Tag location update.
You should be able to update the Model 3 transform and have it update the location of Model 3, Model 2, and Tag.

https://github.com/awslabs/iot-app-kit/assets/109186219/b5e447d0-a064-41e9-a909-784ecabde960

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/store/helpers/sceneDocumentHelpers.ts
+++ b/packages/scene-composer/src/store/helpers/sceneDocumentHelpers.ts
@@ -163,15 +163,20 @@ export const updateSceneNode = (
   partial: RecursivePartial<ISceneNodeInternal>,
   skipEntityUpdate?: boolean,
 ): void => {
+  // cache these values before merge deep overwrites them
+  const nodeToMove = draft.document.nodeMap[ref];
+  const oldParentRef = nodeToMove?.parentRef;
+  const oldParent = draft.document.nodeMap[oldParentRef || ''];
+
+  // Ignore childRefs from partial since it's determined by parentRef and handled by this function
+  const filteredPartial = partial;
+  delete partial.childRefs;
+
   // update target node
-  mergeDeep(draft.document.nodeMap[ref], partial);
+  mergeDeep(draft.document.nodeMap[ref], filteredPartial);
 
   // Reorder logics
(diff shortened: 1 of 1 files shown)


hwandersman:
Looks like this is reverting a piece of this [commit](https://github.com/awslabs/iot-app-kit/commit/ce10aba15c135384cba66f018fb2f502aac1d4eb#diff-fa8c814d8236a62ceb29662aad5d63a16a012d5ff55f0f6ffa0e8ea73e3c7573)

haweston:
Yes it is.  I made a breaking change in that commit that I didn't realize the impact of.  I moved the copy of oldParent info commands inside the if for if there was work to done on reparenting, but this put the copy after the mergeDeep which updates the parentRef information on the node so it's no longer available to copy.  So for example if a node got moved to scene root oldParentRef was getting set to undefined instead of the old parent ref which means the opposite branches got executed for updating the childRefs etc.

> Looks like this is reverting a piece of this [commit](https://github.com/awslabs/iot-app-kit/commit/ce10aba15c135384cba66f018fb2f502aac1d4eb#diff-fa8c814d8236a62ceb29662aad5d63a16a012d5ff55f0f6ffa0e8ea73e3c7573)

hwandersman (approved):

cshang2017 (approved):

PR #2800, opened 2024-05-09
feat: ds ga

cshang2017:
1. add copy scene nodes feature
2. increase query depth to 10 levels

## Overview

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/utils/entityModelUtils/copySceneNodes.spec.ts
+++ b/packages/scene-composer/src/utils/entityModelUtils/copySceneNodes.spec.ts
@@ -1,0 +1,47 @@
+import { TwinMakerSceneMetadataModule } from '@iot-app-kit/source-iottwinmaker';
+
+import { setTwinMakerSceneMetadataModule } from '../../common/GlobalSettings';
+
+import { copySceneNodes } from './copySceneNodes';
+import { cloneSceneNodes, createSceneRootEntity, fetchSceneNodes, saveSceneNodes } from './sceneUtils';
+
+jest.mock('./sceneUtils', () => ({
+  cloneSceneNodes: jest.fn(),
+  createSceneRootEntity: jest.fn(),
+  fetchSceneNodes: jest.fn(),
+  saveSceneNodes: jest.fn(),
+}));
+
+const srcNodes = [{ ref: 'node1' }, { ref: 'node2', parentRef: 'node1' }];
+const destNodes = [{ ref: 'node3' }, { ref: 'node4', parentRef: 'node3' }];
+
+describe('copySceneNodes', () => {
+  const createSceneEntity = jest.fn();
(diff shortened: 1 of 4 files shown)


packages/scene-composer/src/utils/entityModelUtils/sceneUtils.ts
```diff
@@ -97,6 +99,46 @@ export const isDynamicScene = (document?: ISceneDocumentInternal): boolean => {
   return !isEmpty(document?.properties?.[KnownSceneProperty.SceneRootEntityId]);
 };
 
+export const useSceneRootEntityId = (document?: ISceneDocumentInternal): string => {
```
haweston, line 102: May have linter issues as use<functionName> implies it's a react hook but this isn't.

packages/scene-composer/src/utils/entityModelUtils/sceneUtils.ts
```diff
@@ -97,6 +99,46 @@ export const isDynamicScene = (document?: ISceneDocumentInternal): boolean => {
...
+  const sceneRootEntityId = document?.properties?.[KnownSceneProperty.SceneRootEntityId];
+  return sceneRootEntityId;
+};
+
+export const saveSceneNodes = async (
```
haweston, line 107: We still have coverage on this because it's used in even more places now, but might be worth adding a unit test for this function now that's it's broken out and more testable.

packages/scene-composer/src/utils/entityModelUtils/copySceneNodes.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  it('should call related functions', async () => {
+    const sourceSceneRootEntityId = 'src-scene-root-entity-id';
+    await copySceneNodes(sourceSceneRootEntityId);
+
+    expect(createSceneRootEntity).toBeCalledTimes(1);
+    expect(fetchSceneNodes).toBeCalledTimes(1);
+    expect(cloneSceneNodes).toBeCalledTimes(1);
+    expect(saveSceneNodes).toBeCalledTimes(1);
```
haweston, line 45: Could upgrade to toBeCalledWith the return from moved cloneSceneNodes on the save.

haweston (approved):
None of the comments are blockers.

packages/scene-composer/src/utils/entityModelUtils/copySceneNodes.spec.ts
cshang2017: good point.

cshang2017:
If none of blocker, How about having another PR to address comments?

PR #2822, opened 2024-05-21
feat: customize gauge in dashboard config

gmuralidharreddy:
## Overview
This PR is for #2634 
Added config panel changes for gauge component as per UX.

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/139960352/e6fe5bc5-8318-46c3-a8f1-1da7488a5a49

## Legal


packages/dashboard/src/customization/propertiesSections/gaugeStyleSettings/displaySettingsSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import { Box, Checkbox, FormField } from '@cloudscape-design/components';
+
+import { PropertyLens } from '~/customization/propertiesSection';
+import { GaugeProperties } from '~/customization/widgets/types';
+import { DashboardWidget } from '~/types';
+import { StyledExpandableSection } from '../components/styledComponents';
+import { maybeWithDefault } from '~/util/maybe';
+
+export const DisplaySettingsSection = ({
```
corteggiano, line 11: instead of writing a new `DisplaySettings` section specific to the Gauge component could we update the existing one 
which is found at `dashboard/src/customization/propertiesSections/displaySettingsSection/index.tsx`

packages/dashboard/src/customization/propertiesSections/gaugeStyleSettings/axisSettingsSection.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { PropertyLens } from '~/customization/propertiesSection';
+import { GaugeProperties } from '~/customization/widgets/types';
+import { DashboardWidget } from '~/types';
+import {
+  FormLabel,
+  StyledExpandableSection,
+} from '../components/styledComponents';
+import { maybeWithDefault } from '~/util/maybe';
+
+export const AxisSettingsSection = ({
```
corteggiano, line 14: instead of a new YAxis settings section could you reuse the one found at `dashboard/src/customization/propertiesSections/lineAndScatterStyleSettings/yAxis.tsx` 

you will need to move it out of its current folder under LineAndScatterStyles into its own and add in some logic to only show the Y-axis label input for the line chart

corteggiano (changes requested):

corteggiano:
looks so good! tested it out next to some other widgets and you can see how every time the data changes, the Gauge will animate filling up form 0% 

could you try to animate from its previous value instead and if that's not possible then disable the animations entirely

https://github.com/awslabs/iot-app-kit/assets/28601414/d679b20e-2d54-4c93-8915-f9ba97636a92

gmuralidharreddy:
> looks so good! one minor bug and one slightly bigger one
> 
> 1. the `Label font size` doesn't appear to be doing anything to change the `property name`'s font size
> 2. tested it out next to some other widgets and you can see how whenever the data changes, the Gauge will animate filling up from 0% every time. here's a video showing the bug:
>    https://github.com/awslabs/iot-app-kit/assets/28601414/d679b20e-2d54-4c93-8915-f9ba97636a92
> 
> could you change that to animate from its previous value instead and if that's not possible then disable the animations entirely

1. Previously label font size is works for lables(10, 20, 30, 40...etc) which are outside arc when there are thresholds. Now I updated it to title(Properties name).
2. In my local it was working for me but anyway I make valueAnimation as false. Please check and confirm.

corteggiano:
**Deleting a threshold doesn't remove it from the widget:** 
image

**And then interacting with any widget settings after removing the widget (changing from raw data to average count in this image below) makes double text appear:**
image

**Changing the property's label through `Properties Panel => Property => Label` makes the widget rerender every few seconds. You can see the issue in the video below:**
https://github.com/awslabs/iot-app-kit/assets/28601414/04a68c5a-63a6-4020-89db-1298f2f188fb

corteggiano:
LGTM! will approve and merge once conflicts are solved :)

gmuralidharreddy:
> LGTM! will approve and merge once conflicts are solved :)

Resolved conflicts, thanks :)

corteggiano (approved):

PR #2823, opened 2024-05-22
feat(dashboard): default viewport setting

corteggiano:
## Overview
Add viewport settings in the Dashboard Settings panel
Had to stringify the viewport in the redux actions since it doesnt like Date objects

https://github.com/awslabs/iot-app-kit/assets/28601414/cf4dc8b7-d7c9-4f4e-a453-1e3088e12507

Screenshot 2024-05-24 at 09 44 45
Screenshot 2024-05-24 at 09 44 49

## Legal


packages/dashboard/src/util/viewport.ts
```diff
@@ -0,0 +1,34 @@
...
+  if ('duration' in viewport) {
+    return viewport;
+  } else {
+    return {
+      start: new Date(viewport.start),
+      end: new Date(viewport.end),
+      group: viewport.group,
+    };
+  }
+};
```
tracy-french, line 34: Neither of these functions appear to be doing what their name implies. Shouldn't there be some JSON.stringify/JSON.parse involved?

packages/dashboard/src/components/defaultViewport/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+  applyButtonLabel: 'Apply',
+  formatRelativeRange: (e: DateRangePickerProps.RelativeValue) =>
+    e.amount === 1 ? `Last ${e.unit}` : `Last ${e.amount} ${e.unit}s`,
+  formatUnit: (e: DateRangePickerProps.TimeUnit, n: number) =>
+    1 === n ? e : `${e}s`,
+  dateRangeIncompleteError:
+    'The selected date range is incomplete. Select a start and end date for the date range.',
+  dateRangeInvalidError:
+    'The selected date range is invalid. The start date must be before the end date.',
+};
```
tracy-french, line 34: Do we need to do this if we can use this? https://cloudscape.design/get-started/dev-guides/internationalization/

Also, is this how react-intl handles internationalization?

packages/dashboard/src/util/viewport.ts
corteggiano: oh yeah that woudl've been easier lol. updated code :)

packages/dashboard/src/components/defaultViewport/constants.ts
corteggiano: dashboard doesn't have `react-intl` support yet 🤔 can definitely bring in the default cloudscape messages for this though!

packages/dashboard/src/components/internalDashboard/index.tsx
```diff
@@ -10,6 +10,8 @@ import {
   spaceScaledXxxs,
 } from '@cloudscape-design/design-tokens';
 import { ContentLayout } from '@cloudscape-design/components';
+import messages from '@cloudscape-design/components/i18n/messages/all.all';
+import { I18nProvider } from '@cloudscape-design/components/i18n';
```
corteggiano, line 14: places provider here because we export both a View and Edit Dashboard component and they both implement `InternalDashboard`

tracy-french (approved):

PR #2830, opened 2024-05-23
chore(scene): detect dynamic scenes missing their root entity

haweston:
## Overview
When a dynamic scene has it's root entity deleted the scene no longer functions because it doesn't have a parent entity to attach other entities to.  The scene composer should detect this and provide an error message alerting the users instead of letting them attempt changes and then giving an error.

## Verifying Changes
Create a Dynamic Scene
Manually Delete the Sceen Root Entity
Open the scene in Storybook and observe the error message
Screenshot 2024-05-23 at 11 57 24 AM

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/components/SceneLayers.tsx
+++ b/packages/scene-composer/src/components/SceneLayers.tsx
@@ -1,12 +1,16 @@
-import React, { useContext, useEffect } from 'react';
+import React, { useCallback, useContext, useEffect } from 'react';
 import { useQuery } from '@tanstack/react-query';
+import { useIntl } from 'react-intl';
 
+import { getGlobalSettings } from '../common/GlobalSettings';
 import { sceneComposerIdContext } from '../common/sceneComposerIdContext';
 import { accessStore } from '../store';
-import { fetchSceneNodes } from '../utils/entityModelUtils/sceneUtils';
+import { checkIfEntityExists, fetchSceneNodes } from '../utils/entityModelUtils/sceneUtils';
 import { KnownSceneProperty } from '../interfaces';
+import { DisplayMessageCategory } from '../store/internalInterfaces';
 
 export const SceneLayers: React.FC = () => {
+  const { formatMessage } = useIntl();
   const sceneComposerId = useContext(sceneComposerIdContext);
   const isViewing = accessStore(sceneComposerId)((state) => state.isViewing());
   const autoUpdateInterval = accessStore(sceneComposerId)(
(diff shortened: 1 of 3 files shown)


packages/scene-composer/src/components/SceneLayers.tsx
```diff
@@ -31,11 +36,32 @@ export const SceneLayers: React.FC = () => {
...
+          },
+        ]);
+      }
+    }
+  }, [sceneRootEntityId, addMessages, formatMessage]);
+
   useEffect(() => {
     if (nodes.data) {
-      renderSceneNodes(nodes.data);
+      if (nodes.data.length > 0) {
```
hwandersman, line 58: Wouldn't you need to check if there's a scene root entity no matter what? What is the `nodes` object going to look like if `sceneRootEntityId` doesn't exist?

mumanity: nit-pick: you can just put `if (nodes.data.length)` because any length is more than 0. 👍

hwandersman:
I assume we want the "Learn more" button link to link to our docs? How is that set up?

packages/scene-composer/src/components/SceneLayers.tsx
haweston: So if  the scene has a sceneRootEntityID the query gets run.  So if(nodes.date) guards against static scenes where we didn't run the query and also don't need to check for the root entity.  When the scene is dynamic the query gets run and nodes.data exists but may be length zero.  If it's not length zero it means the query based around the root so we don't need to manually check for it. But if it's zero then we need to go check and see if the query came back empty because there was no root.  It could come back length zero because there is a valid scene root but nothing placed in the scene to be child of it.

haweston: That nit pick is a logic error because length can be zero because data is an array so it's length can be present and set to zero.

haweston:
> I assume we want the "Learn more" button link to link to our docs? How is that set up?

I checked and that's set in https://github.com/awslabs/iot-app-kit/blob/main/packages/scene-composer/src/components/modals/MessageModal.tsx#L88 without currently an input variable to let it link to different things.  I think that's maybe a worthwhile upgrade, but out of scope for this feature.

hwandersman (approved):

PR #2846, opened 2024-06-04
feat(dashboard): move dashboard migration utility to app-kit

ssjagad:
## Overview
Moving over the migration script used by centurion into app-kit so it can be used by multiple consumers.

## Verifying Changes

All tests passing should verify this change.

## Legal

### DIFF
--- a/packages/dashboard/src/migration/convert-monitor-to-app-defintion.ts
+++ b/packages/dashboard/src/migration/convert-monitor-to-app-defintion.ts
@@ -1,0 +1,318 @@
+import {
+  DashboardConfiguration,
+  DashboardWidget,
+  MigrateDashboard,
+} from '../types';
+import {
+  DashboardWidgetType,
+  MonitorWidgetType,
+  MonitorMetric,
+  SiteWiseMonitorDashboardDefinition,
+  MonitorAnnotations,
+  MonitorWidget,
+  QueryConfig,
+  AssetMap,
+} from './types';
+import { nanoid } from 'nanoid';
+import {
+  appCellPerMonitorSquareHeight,
+  appCellsPerMonitorSquareWidth,
+  defaultDisplaySettings,
+  defaultResolution,
+  minHeight,
+  minWidth,
+} from './constants';
+import {
+  getKPIAndGridData,
+  getProperty,
+  getStaticProperties,
+  getStyleSettings,
+} from './getters';
+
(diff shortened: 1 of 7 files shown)


packages/dashboard/src/migration/convert-monitor-to-app-defintion.ts
```diff
@@ -0,0 +1,509 @@
+import {
```
corteggiano, line 1: this file is pretty big! would love if it could be split up into smaller files (for example: convertLineChart, convertKPI)

tracy-french: +1

packages/dashboard/src/migration/convert-monitor-to-app-defintion.ts
```diff
@@ -0,0 +1,509 @@
...
+}
+
+interface ApplicationQueryConfig {
+  source: string;
+  query: ApplicationQuery;
+}
+
+interface QueryConfig {
+  queryConfig: ApplicationQueryConfig;
+}
```
tracy-french, line 43: Why are some types in this file and some in the types.ts file? How do you choose where they go?

ssjagad: fixed, moved everything to one type file for anything being used in folder

packages/dashboard/src/migration/convert-monitor-to-app-defintion.ts
ssjagad: I think it makes sense to leave all the conversion logic in one file. It all gets used by convertWidget which in turn in used by migrateDashboard. I think a better way is to split by conversion logic, and functions that set defaults. Happy to do it the other way though if we don't agree.

ssjagad: I could maybe change the file name from just getters to widget-data-getters to be more specific though?

corteggiano (approved):

packages/dashboard/src/migration/convert-monitor-to-app-defintion.ts
ssjagad: merged in for now, but i can change this on the follow up PR with the overlapping bug fix

PR #2859, opened 2024-06-12
fix(scene): fix dynamic scene entity binding reload on query

haweston:
## Overview
Dynamic Scenes don't load the entity binding on a node in the Scene Composer even though it's saved on the scene entity

## Verifying Changes
In scene composer for a dynamic scene add an entity binding
verify that the entity binding got saved to the isVisualOf property of the scene nodes 3d. node component
refresh the scene composer
the entity binding should load when you inspect the scene node again 

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


cshang2017 (commented):
one concern on default query rate limit 10 queries/sec

mumanity (approved):

PR #2865, opened 2024-06-14
fix(dashboard): make onSave optional

corteggiano:
## Overview
make on save optional so custom toolbar users dont need to provide it

## Legal

### DIFF
--- a/packages/dashboard/src/components/dashboard/index.tsx
+++ b/packages/dashboard/src/components/dashboard/index.tsx
@@ -33,8 +33,8 @@ import { TimeSync } from '@iot-app-kit/react-components';
 import { debounce } from 'lodash';
 
 export type DashboardProperties = {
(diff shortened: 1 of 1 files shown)


jmbuss (approved):

packages/dashboard/src/components/dashboard/index.tsx
```diff
@@ -33,8 +33,8 @@ import { TimeSync } from '@iot-app-kit/react-components';
 import { debounce } from 'lodash';
 
 export type DashboardProperties = {
-  onSave: DashboardSave;
   onDashboardConfigurationChange?: DashboardConfigurationChange;
+  onSave?: DashboardSave;
```
jmbuss, line 37: do we need to add a noop implementation in the header if they are using the default one?

PR #2868, opened 2024-06-14
fix: anomaly data zoom

jmbuss:
## Overview
Fix for 2 bugs related to anomaly chart datazoom
1. bar flickering on tick. To fix this we apply a sampling function when the data density is larger than the pixel size. This ensures that there are not z-index issues when rendering the points.
2. toolbox datazoom does nothing. To fix this, all zoom handling in the datazoom hook is throttled using request animation frame. This ensures the correct ordering of events and allows us to cancel zoom handling that no longer applies.

## Legal

### DIFF
--- a/packages/react-components/src/components/anomaly-chart/constants.ts
+++ b/packages/react-components/src/components/anomaly-chart/constants.ts
@@ -65,6 +65,7 @@ export const ANOMALY_LEGEND_PADDING = [
 export const ANOMALY_BAR_SERIES_CONFIGURATION = {
   barMinWidth: 2,
   barMaxWidth: 10,
+  sampling: 'lttb',
   stack: 'Total',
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/anomaly-chart/constants.ts
```diff
@@ -65,6 +65,7 @@ export const ANOMALY_LEGEND_PADDING = [
 export const ANOMALY_BAR_SERIES_CONFIGURATION = {
   barMinWidth: 2,
   barMaxWidth: 10,
+  sampling: 'lttb',
```
chejimmy, line 68: nice 👍

chejimmy (approved):
LGTM 🚀
Nice UX improvement

PR #2880, opened 2024-06-20
fix(react-components): thresholds properly add and remove series

corteggiano:
## Overview
Add state reducer to properly rerender chart when number of series changes
and prevent rerenders when just the data changes :) 

using similar [reducer and state management as the anomaly widget](https://github.com/awslabs/iot-app-kit/blob/main/packages/react-components/src/components/anomaly-chart/hooks/useAnomalyEchart.ts)

## before: deleting all thresholds still left a "ghost" of the threshold series and value
Screenshot 2024-05-24 at 09 21 16

## after
https://github.com/awslabs/iot-app-kit/assets/28601414/fcad0a1b-baf2-4a6c-aeba-3c2947469c70

## Legal


chejimmy (approved):
Code LGTM! Could you add some (snap) tests for the person who need to make changes? 🙏

ssjagad (approved):

PR #2882, opened 2024-06-20
feat(anomaly-chart): added `gestures` enablement option

chejimmy:
## Overview

This PR adds the `gestures` option to the Anomaly Chart, the same `gestures` from [Chart](https://awslabs.github.io/iot-app-kit/?path=/docs/components-chart--docs).

## Verifying Changes

https://github.com/awslabs/iot-app-kit/assets/50635800/238cb12c-c0af-4ab6-851a-7ceddc6beae6

## Legal

### DIFF
--- a/packages/react-components/src/echarts/unboundedZoom/constants.ts
+++ b/packages/react-components/src/echarts/unboundedZoom/constants.ts
@@ -1,4 +1,8 @@
-import { DataZoomComponentOption, XAXisComponentOption } from 'echarts';
+import {
+  DataZoomComponentOption,
+  ToolboxComponentOption,
+  XAXisComponentOption,
+} from 'echarts';
 
 export const ECHARTS_GESTURE = 'echarts-gesture';
 
@@ -8,6 +12,14 @@ export const DEFAULT_DATA_ZOOM: DataZoomComponentOption = {
   zoomOnMouseWheel: true,
   moveOnMouseMove: 'shift',
   moveOnMouseWheel: false,
+};
+
+export const DEFAULT_DATA_ZOOM_GESTURES_ENABLED: DataZoomComponentOption = {
+  disabled: false,
+};
+
+export const DEFAULT_DATA_ZOOM_GESTURES_DISABLED: DataZoomComponentOption = {
+  disabled: true,
 };
 
 // this is the chart live mode refresh rate, this should be inline with the animation props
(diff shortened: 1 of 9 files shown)


packages/react-components/src/echarts/unboundedZoom/constants.ts
```diff
@@ -32,15 +44,20 @@ export const DEFAULT_X_AXIS: XAXisComponentOption = {
...
-export const DEFAULT_TOOLBOX = {
+export const DEFAULT_TOOLBOX_GESTURES_ENABLED: ToolboxComponentOption = {
   show: true,
   top: 0,
   right: 8,
   feature: {
     dataZoom: {
       title: { back: 'Undo\nzoom' },
       filterMode: 'none',
+      xAxisIndex: 'all',
```
jmbuss, line 55: Curious what this does? allow events to trigger off any x axis in the chart instance?

packages/react-components/src/echarts/unboundedZoom/useUnboundedDataZoom.ts
```diff
@@ -87,13 +94,51 @@ const reducer = (state: TickState, action: TickAction): TickState => {
...
+    dataZoomOption = {
+      ...dataZoomOption,
+      ...DEFAULT_DATA_ZOOM_GESTURES_DISABLED,
+    };
+    toolboxOption = DEFAULT_TOOLBOX_GESTURES_DISABLED;
+  }
+  return { dataZoomOption, toolboxOption };
+};
+
+const dispatchGesturesEnablement = (
```
jmbuss, line 120: is this so that things work if the gestures flag changes on the fly?

packages/react-components/src/echarts/unboundedZoom/constants.ts
chejimmy: not needed, it was resetting it back to the default value.
> Defines which [xAxis](https://echarts.apache.org/en/option.html#xAxis) should be controlled. By default, it controls all x axes.

packages/react-components/src/echarts/unboundedZoom/useUnboundedDataZoom.ts
chejimmy: yup, setting config does not deactivate the datazoon selection.

PR #2890, opened 2024-06-26
fix(scene-composer): reverting zustand upgrade due to Grafana issues related to zustand, react, r3f

hwandersman:
## Overview
The latest scene viewer panel in Grafana fails to load the scene canvas:
image (15)

After digging deeper into this error I found that the latest AppKit version we added to Grafana upgrade zustand from v3 to v4 which uses `useSyncExternalStore` internally. Found a [thread online](https://github.com/pmndrs/zustand/issues/1164) that explained how this version of zustand, along with state from multiple renderers like React and react-three/fiber (r3f) causes issues. 

Since this is breaking existing Grafana customers I downgraded zustand back to v3. This should not introduce any breaking or impactful changes to the consuming packages of the AppKit (TwinMaker Console, Grafana plugin, etc.). Reverting this commit: https://github.com/awslabs/iot-app-kit/pull/2724

## Verifying Changes

1. Tested in storybook - loaded a dynamic scene
2. Tested in Grafana - copied over the dist build of the scene composer and the downgraded version of zustand in the plugin node_modules. Saw a dynamic scene load into the panel successfully.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/scene-composer/src/store/slices/SceneDocumentSlice.ts
+++ b/packages/scene-composer/src/store/slices/SceneDocumentSlice.ts
@@ -1,5 +1,6 @@
 import { isDataBindingTemplate } from '@iot-app-kit/source-iottwinmaker';
 import { cloneDeep, isEmpty, isString } from 'lodash';
+import { GetState, SetState } from 'zustand';
 import { ComponentUpdateType } from '@aws-sdk/client-iottwinmaker';
 
 import {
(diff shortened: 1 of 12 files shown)


packages/scene-composer/src/store/slices/SceneDocumentSlice.ts
```diff
@@ -202,7 +202,7 @@ export const createSceneDocumentSlice: SliceCreator<keyof ISceneDocumentSlice> =
               ]);
             });
         } else {
-          appendSceneNode(draft as RootState, node, disableAutoSelect);
```
hwandersman, line 205: Should I not have made this change? Is this cast important?

mitchlee-amzn (approved):

packages/scene-composer/src/store/slices/SceneDocumentSlice.ts
haweston: If it still loads saved static and dynamic scenes then this cast removal should be fine.

PR #2898, opened 2024-06-28
fix(react-components): improve axis styling and add labels

corteggiano:
## Overview

- Removed the vertical bars in the chart grid
- Removed y-axis border
- Removed ticks from x-axis
- Added labels for each axis -- can be a user provided string or will default to "Time" and "Total contribution" 

### Showing x-axis, y-axis, and timestamp bar
Screenshot 2024-06-27 at 12 35 03

### Hiding y-axis
Screenshot 2024-06-27 at 12 34 44

### Hiding x-axis and y-axis
Screenshot 2024-06-27 at 12 34 18

### Hiding timestamp bar and y-axis
image

### Hiding x-axis, y-axis, and timestamp bar
Screenshot 2024-06-27 at 12 34 27

### Dark mode
image

## Legal


packages/react-components/src/echarts/themes/registerCloudscapeThemes.ts
```diff
@@ -255,7 +255,7 @@ export const registerCloudscapeThemes = () => {
       axisLine: {
         show: true,
         lineStyle: {
-          color: '#e0e2e7',
+          color: '#404d5c',
```
mumanity, line 258: Shouldn't we use the [tokens](https://cloudscape.design/foundation/visual-foundation/design-tokens/#tokens) so that these don't break with cloudscape updates?

corteggiano: sadly we cant directly use the tokens because they are CSS variables and echarts will only accept the actual color as hex or name :( this is a big issue that we've run into when integrating echarts w cloudscape

mumanity (approved):

PR #2907, opened 2024-07-09
feat: use time series data request manager

jmbuss:
## Overview
This is the first piece of code for the useTimeSeriesData implementation. This includes a request manager which ensures that requests are deduped, and that the correct time intervals are picked. 

Time intervals are created from:
1. a static list of "live data" intervals. These are created in order to account for any late data points and configure different refresh rates.
2. any existing cached intervals, so that we don't need to re-request those intervals.
3. any gaps in the viewport and the above 2 lists.

There are 2 parts to this PR. the generic abstraction for resolving these intervals for any set of time series data. These classes are contained in the useTimeSeriesData hook. The concrete implementations will be in their own hooks, for example, the useAssetPropertyValueHistory hook. This will be the implementation of useTimeSeriesData for raw historical data. There will also be an implementation for useAssetPropertyAggregates and useGetAssetPropertyValue. Those will follow after completion of useAssetPropertyValueHistory as their implementations will be roughly the same with a few modifications to the cache client and request functions.

## Legal


packages/react-components/src/queries/queryClient.ts
```diff
@@ -9,5 +9,5 @@ export const queryClient = new QueryClient({
       staleTime: STALE_TIME,
     },
   },
-  queryCache: new QueryCache({}),
+  queryCache: new QueryCache(),
```
corteggiano, line 12: is there a difference or is it just cleaner this way?

PR #2919, opened 2024-07-19
feat: use time series data

jmbuss:
## Overview

Initial implementation for useTimeSeriesData. Currently it only supports raw data requests, aggregation / resolutions will be added in the next PR.

Below are some diagrams that show the flow of a request.

### General component flow and responsibilities
image

### Component dependencies
image

### Picking resolver strategy
image

## Legal


packages/react-components/stories/utils/query.ts
```diff
@@ -77,14 +77,14 @@ export const getTimeSeriesDataQuery = (
           {
             refId: '1',
             propertyId: propertyId1,
-            aggregationType: 'AVERAGE',
-            resolution: '1m',
+            // aggregationType: 'AVERAGE',
```
corteggiano, line 80: will this break the story for anyone using the current implementation of useTimeSeriesData>

jmbuss: No, this is just for storybook. I was testing raw vs aggregate data with the new implementation.

corteggiano (approved):

PR #2921, opened 2024-07-24
fix(react-components): gauge widget properly shows property name

corteggiano:
## Overview
Before: property name was not displayed anywhere
Now: property name shown above data quality, font size customizable by `labelFontSize` setting

name, no data quality:
Screenshot 2024-07-24 at 15 22 35

name with data quality:
Screenshot 2024-07-24 at 15 22 49

larger font size:
Screenshot 2024-07-24 at 15 23 18

## Legal


dpportet (dismissed):
Looks good, but should the default text size be bigger? It looks small compared to the value.

corteggiano:
> Looks good, but should the default text size be bigger? It looks small compared to the value.

updated! 
image
image

PR #2941, opened 2024-08-07
chore(dashboard): add tests for text widget to include url sanitization

ssjagad:
## Overview
Adding unit and e2e tests for href sanitation change

## Verifying Changes
Tests pass

## Legal


packages/dashboard/e2e/tests/widgets/text.spec.ts
```diff
@@ -280,5 +280,38 @@ test.describe('Test Text Widget', () => {
...
     });
+
+    test('when link is enabled, it should sanitize href and not link if dangerous', async ({
+      dashboardWithTextWidget,
+      configPanel,
+      page,
+    }) => {
+      const widget = dashboardWithTextWidget.gridArea.locator(
+        '[data-gesture=widget]'
+      );
```
tracy-french, line 291: [Out of scope]: I do not think this is a good selector. `.locator` should be used as a last resort and I do not understand `[data-gesture=widget]`. If I was a user, I would not find my text widget by looking for `[data-gesture=widget]`. Using a selector like this is brittle and disconnected from how a user uses the dashboard. We should improve this. I realize the other tests are using this, so no need to address here.

packages/dashboard/src/customization/widgets/text/component.test.tsx
tracy-french, line 1: These tests are so much better. Thank you, Sahil!

packages/dashboard/src/customization/widgets/text/component.test.tsx
```diff
@@ -13,88 +11,101 @@ jest.mock('~/store/actions', () => ({
...
+    const href = 'javascript://%0Aalert(1)';
+    render(<TestReadonlyTextWidget value={text} href={href} isUrl />);
+
+    const textWidget = screen.getByText(text);
+    expect(textWidget).toBeVisible();
+    expect(textWidget).not.toHaveAttribute('href');
+    expect(screen.queryByRole('link', { name: text })).not.toBeInTheDocument();
+  });
+
+  it('should santize and then render unsafe link 2', () => {
```
tracy-french, line 91: [Nit]: It is hard to tell what specifically the tests are testing for beyond "not unsafe" due to the duplication of the name and the usage of a number (2) to distinguish. It would be better if the test names were more explicit.

packages/dashboard/src/customization/widgets/text/component.test.tsx
```diff
@@ -13,88 +11,101 @@ jest.mock('~/store/actions', () => ({
...
+    const text = 'text';
+    const href = '\x6A\x61\x76\x61\x73\x63\x72\x69\x70\x74\x3aalert(1)';
+    render(<TestReadonlyTextWidget value={text} href={href} isUrl />);
+
+    const textWidget = screen.getByText(text);
+    expect(textWidget).toBeVisible();
+    expect(screen.queryByRole('link', { name: text })).not.toBeInTheDocument();
```
tracy-french, line 98: [Super nit]: This assertion is not in the same place as the other tests (i.e., 2nd expect vs last in the others). Very minor thing, but it sticks out as "different", when it is not.

tracy-french (approved):
Good work! Prevents an issue and increases the quality of our unit tests. Thank you, Sahil!

PR #2943, opened 2024-08-09
chore(react-components): kpi tests

corteggiano:
## Overview
Add tests for KPI component

## Legal


packages/react-components/src/components/kpi/tests/kpi.spec.tsx
```diff
@@ -0,0 +1,226 @@
...
+    expect(screen.queryByTestId('kpi-timestamp')).toBeNull();
+    expect(screen.queryByTestId('kpi-aggregation')).toBeNull();
+
+    // KPI value should be visible
+    expect(screen.getByTestId('kpi-value').textContent).toContain(
+      `${DATA_STREAM.data[0].y.toFixed(4)}`
+    );
+  });
+
+  it('kpi settings dont impact error state', () => {
```
tracy-french, line 133: What is the goal of this test? Do you need it?

packages/react-components/src/components/kpi/tests/kpi.spec.tsx
```diff
@@ -0,0 +1,226 @@
...
+  it('does not render the kpi + side threshold in an error state', () => {
+    render(
+      <KPI query={ERROR_QUERY} viewport={VIEWPORT} thresholds={[THRESHOLD]} />
+    );
+    expect(screen.queryByTestId('kpi-side-threshold')).toBeNull();
+  });
+});
+
+describe('kpi with custom settings', () => {
+  it('renders a kpi without any extra text', () => {
```
tracy-french, line 101: What about tests where the settings are `true`? Those would be more valuable than the false equivalents.

packages/react-components/src/components/kpi/tests/kpi.spec.tsx
```diff
@@ -0,0 +1,226 @@
...
+      <KPI query={SUCCESS_QUERY} settings={{ backgroundColor: '#00ff00' }} />
+    );
+
+    // background color is set
+    expect(screen.getByTestId('kpi-base-component')).toHaveStyle(
+      'background-color: #00ff00'
+    );
+
+    // font color is calculated to use the highest contrast color (in this case it is black)
+    expect(screen.getByTestId('kpi-value')).toHaveStyle('color: black');
```
tracy-french, line 198: Assert the color of the text like:
```ts
expect(screen.getByText(DATA_STREAM.data[0].y)).toHaveStyle('color: black');
```

packages/react-components/src/components/kpi/tests/kpi.spec.tsx
```diff
@@ -0,0 +1,226 @@
...
+  });
+
+  it('kpi widget in error state with custom green background color', () => {
+    render(
+      <KPI query={ERROR_QUERY} settings={{ backgroundColor: '#00ff00' }} />
+    );
+
+    // background color is set to the custom background color
+    expect(screen.getByTestId('kpi-error-component')).toHaveStyle(
+      'background-color: #00ff00'
```
tracy-french, line 223: Create variables for any values you setup at the start of a test which you are writing assertions for.

packages/react-components/src/components/kpi/tests/mockData.tsx
tracy-french, line 1: If you are only using this data in the one unit test file, I would keep it all in the unit test file. It is going to be a lot easier to change if you have everything in one place. Not only that, it's easier to understand the tests since it's easier to see what the values are at the same time.

packages/react-components/stories/kpi/default-kpi-error.stories.tsx
tracy-french, line 1: Do you need a separate file for each story?

packages/react-components/stories/kpi/kpi-mock-data.ts
```diff
@@ -39,6 +39,42 @@ export const MOCK_TIME_SERIES_DATA_QUERY = mockTimeSeriesDataQuery([
   },
 ]);
 
+export const MOCK_TIME_SERIES_DATA_QUERY_ERROR = mockTimeSeriesDataQuery([
+  {
+    dataStreams: [
+      {
+        ...DATA_STREAM,
+        data: [],
+        error: { msg: 'Error: Oh no!!! Something has gone wrong' },
```
tracy-french, line 48: What is a more realistic error message?

packages/react-components/src/components/kpi/tests/mockData.tsx
```diff
@@ -0,0 +1,69 @@
...
+  name: 'some name',
+  unit: 'mph',
+};
+
+export const THRESHOLD = {
+  comparisonOperator: COMPARISON_OPERATOR.GT,
+  value: 120,
+  visible: true,
+  color: '#ff0000', // red threshold
+} as StyledThreshold;
```
tracy-french, line 37: What does this need to be cast? Are there missing fields?

packages/react-components/src/components/kpi/tests/kpi.spec.tsx
tracy-french: Also include tests are the different permutations. You don't have to do everything, yet you should have tests for:
1. All true.
2. Tests for each option is true and the rest are false.
3. Tests where defining true for two or more has any special behavior.

packages/react-components/src/components/kpi/tests/kpi.spec.tsx
```diff
@@ -0,0 +1,226 @@
...
+      ERROR_DATA_STREAM.error.msg
+    );
+  });
+
+  it('kpi widget default decimal places', () => {
+    render(<KPI query={SUCCESS_QUERY} />);
+
+    // KPI value should be visible with the defaut 4 decimal places
+    expect(screen.getByTestId('kpi-value').textContent).toContain(
+      `${DATA_STREAM.data[0].y.toFixed(4)}`
```
tracy-french, line 165: Use constant for default 4.

packages/react-components/src/components/kpi/tests/kpi.spec.tsx
```diff
@@ -0,0 +1,226 @@
...
+  ERROR_DATA_STREAM,
+  ERROR_QUERY,
+  FILLED_THRESHOLD,
+  LOADING_QUERY,
+  SUCCESS_QUERY,
+  THRESHOLD,
+  VIEWPORT,
+} from './mockData';
+
+describe('default kpi widget', () => {
```
tracy-french, line 15: What about tests varying `viewport` or `styles` props?

tracy-french, line 15: I would add a few tests where the `useViewport` hook return value is changed to show how https://github.com/awslabs/iot-app-kit/blob/main/packages/react-components/src/components/kpi/kpi.tsx#L41 works.

packages/react-components/src/components/kpi/tests/kpi.spec.tsx
corteggiano: the true settings are the default which is tested by the entire first group of tests
will add some which test the permutations here !

packages/dashboard/jest-setup.js
```diff
@@ -20,9 +20,9 @@ afterAll(() => {
 
 /** Disable logging during tests to keep the test runner clean. Comment out lines as needed. */
 function disableLogging() {
-  // console.error = () => {};
```
corteggiano, line 23: this was mistakenly commented out a few months ago

packages/react-components/src/components/kpi/tests/kpiBase.spec.tsx
```diff
@@ -127,13 +127,14 @@ describe('data quality', () => {
...
     const Y_VALUE = 123445;
+    const Y_VALUE_DECIMALS = Y_VALUE.toFixed(4);
     render(
       <KpiBase
         propertyPoint={{ x: new Date().getTime(), y: Y_VALUE }}
         settings={{ showName: false }}
       />
     );
-    expect(screen.queryByText(Y_VALUE)).not.toBeNull();
+    expect(screen.queryByText(Y_VALUE_DECIMALS)).not.toBeNull();
```
tracy-french, line 137: [Nit]: Use `.not.toBeInTheDocument()`.

tracy-french (approved):
Looks great, Luci! Thank you for raising the bar on these tests!

PR #2946, opened 2024-08-19
chore: vend as library instead of a bundle

mitchlee-amzn:
## Overview
Removes bundling from the dashboard package, and instead vends it like the other libraries.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/dashboard/package.json
```diff
@@ -132,8 +128,16 @@
     "@aws-sdk/client-iot-events": "^3.354.0",
     "@aws-sdk/client-iotsitewise": "^3.456.0",
     "@aws-sdk/client-iottwinmaker": "^3.354.0",
+    "@iot-app-kit/charts-core": "^2",
+    "@iot-app-kit/components": "^10",
+    "@iot-app-kit/core": "^10",
+    "@iot-app-kit/core-util": "^10",
+    "@iot-app-kit/react-components": "^10",
```
tracy-french, line 135: Are these actually peerDependencies? We don't make anyone install these themselves. Or am I seeing this wrong?

mitchlee-amzn: I think they should be peerDependencies, but it's a judgement call. Yes they will need to install these themselves if they want to use it, but we'll get a consistent version across the system this way. It's also much more convenient when you are testing internally, you don't have to pack all these packages up and file reference them everytime you want to debug something.

Either way is fine with me though.

PR #2948, opened 2024-08-19
fix: filter component models from query editor

dpportet:
## Overview
* Since component asset models are not associated with any properties, they cannot be used in our query editor to add to dashboards, so this change filters component asset models from the selection in the dynamic assets tab.

## Verifying Changes
* Use the query editor when your account has a component asset model created. Check the dynamic assets tab and make sure no component asset models are displayed there.

## Legal


chejimmy (approved):
LGTM! 🚢

PR #2949, opened 2024-08-19
fix: prevent creating new chart when icon is dragged over different …

dpportet:
## Overview
* Bug - when dragging a new widget, if widget is dropped over a different widget icon, a chart is added to the dashboard when it shouldnt. This changes fixes this.
* This bug resulted from adding keyboard shortcuts. To fix I changed the mouse event from onPointerUp to onClick which prevents the bug from occurring. From my testing this seems to work as intended but let me know if there's other reasons to use onPointerUp.

## Verifying Changes
* Start dragging a new chart onto the grid. Move your mouse back to another chart type. Lift mouse and confirm that a new chart isn't created

https://github.com/user-attachments/assets/539c6d3c-17cb-4c7c-b52f-a34ec311b916

https://github.com/user-attachments/assets/e437ac25-fcc4-4998-8568-92595fde4083

## Legal


chejimmy (approved):
LGTM! 🚢

PR #2963, opened 2024-09-04
feat(react-components): implement request functions and hooks to build clients

hwandersman:
## Overview
Building off of this previous [commit](https://github.com/awslabs/iot-app-kit/commit/86a8e64ffce69ef6fe6a08680ca1c7594824aa69) (PR #2959 )

We need to be able to accept the `IoTSiteWiseClient` and `IoTSiteWise` client object types from the AWS SDK into our hooks and components. Internally we want to use the promise pattern `IoTSiteWise` but externally it is the best practice for customers to input the command pattern `IoTSiteWiseClient`. In the future we may expand on this by narrowing types based on Tracy's proposal:

> The hook creates every API, even if they are not used, relying on type narrowing. This allows us to create a simple and generic implementation. Later, we may optimize to only create the APIs needed.

For now we will accept the entire client types for the following reasons:
1. The resource explorer has already been designed to use the `IoTSiteWise` client and will be passing that to future SW alarm hooks
2. The SW alarm hooks have already been designed to use at least the `IoTSiteWiseClient` client.

We will not yet push for narrowed request functions which is helpful for customers who have custom implementations of API requests. Instead we will accept `IoTSiteWiseClient | IoTSiteWise` as a client, and return `IoTSiteWise` from the `useIoTSiteWiseClient()` hook. 

Note: added comments clarifying that these hooks and functions are experimental and used internally. We will not expose any of these in our public API until further review.

## Verifying Changes

Unit tests and local builds passed.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


ssjagad (approved):

jmbuss (approved):

packages/react-components/src/hooks/requestFunctions/useIoTEventsClient.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+      ...new IoTEvents({
+        credentials: {
+          accessKeyId: '',
+          secretAccessKey: '',
+          sessionToken: '',
+        },
+        region: 'us-east-1',
+      }),
+      describeAlarmModel: describeAlarmModelMock,
+    } as unknown as IoTEvents;
```
fpauer, line 40: instead can we use `satisfies` ?

packages/react-components/src/hooks/requestFunctions/useIoTEventsClient.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @experimental Do not use in production.
+ */
+export function useIoTEventsClient({
+  iotEventsClient,
+}: UseIoTEventsClientOptions): IoTEvents {
+  const iotEvents = useMemo(() => {
+    if (iotEventsClient instanceof IoTEventsClient) {
+      return {
+        describeAlarmModel: createDescribeAlarmModel(iotEventsClient),
+      } as IoTEvents;
```
fpauer, line 26: instead can we use `satisfies` ?

packages/react-components/src/hooks/requestFunctions/useIoTSiteWiseClient.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ *
+ * @param iotSiteWiseClient is an AWS SDK client type for IoT SiteWise
+ * @returns an IoTSiteWise type, promise pattern client
+ *
+ * @experimental Do not use in production.
+ */
+export function useIoTSiteWiseClient({
```
fpauer, line 33: just to understand better, for any Sitewise or Events operation called withint react-components, we would need to add a requestFunction implementation of the operation and return in the hooks ?

Also would not be better to move `useIoTSiteWiseClient` and `useIoTEventsClient` to `core-utils` instead ? so we can use in other packages if necessary.

packages/react-components/src/hooks/requestFunctions/useIoTEventsClient.spec.ts
hwandersman: Just tried to use `satisfies` and it doesn't work. Probably because of the mock.

packages/react-components/src/hooks/requestFunctions/useIoTEventsClient.ts
hwandersman: Not sure if `satisfies` applies here either. We want to explicitly type cast this because we use a different type for the apis. The type is a specific RequestFunction type that doesn't seem to cleanly satisfy the client type.

packages/react-components/src/hooks/requestFunctions/useIoTSiteWiseClient.ts
hwandersman: The team decided offline that we'd put this in `react-components` for now. I was considering putting it in `core` or `core-utils`, but we don't want to make this part of the public API yet. We have more to discuss as a team about the direction of these request functions, and don't want that to block us.

fpauer: if we dont want to make this part of the public API yep makes sense to be in react-components, also considering that it will be only used in the react-components scope

fpauer (approved):

PR #2970, opened 2024-09-09
feat(react-components): add hook for latest asset property value

jmbuss:
## Overview
Tanstack wrapper around batch / get asset property value.

The implementation uses `dataloader` in order to batch requests across multiple hooks into the same batch call. 

## Legal


packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
```diff
@@ -0,0 +1,184 @@
...
+  const { getAssetPropertyValue, batchGetAssetPropertyValue } =
+    useIoTSiteWiseClient({ iotSiteWiseClient });
+
+  const queries = useMemo(() => {
+    return requests.map((request) => {
+      return {
+        enabled:
+          syncFlag &&
+          enabled &&
+          clientIsValid({
```
hwandersman, line 85: Why do you need to validate the input here?

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
```diff
@@ -0,0 +1,184 @@
...
+      clientIsValid(client),
+      'Expected getAssetPropertyValue or batchGetAssetPropertyValue to be defined as required by the enabled flag.'
+    );
+
+    invariant(
+      requestIsValid({ assetId, propertyId, propertyAlias }),
+      'Expected assetId and propertyId to be defined or propertyAlias to be defined as required by the enabled flag.'
+    );
+
+    if (batchGetAssetPropertyValue) {
```
hwandersman, line 151: So we inherently batch calls only if that requestFn exists? We may want to make this more explicit in the hook options.

Or something to consider: do we even want to use the non-batch function at all? Why not only support BatchGet*?

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
```diff
@@ -0,0 +1,184 @@
...
+          propertyId,
+          propertyAlias,
+        },
+        {
+          abortSignal: signal,
+        }
+      );
+    }
+
+    throw 'Expected request to be handled by one of the two client functions as required by the enabled flag.';
```
hwandersman, line 182: Shouldn't this be handled by the invariant check for the client?

packages/react-components/src/queries/useLatestAssetPropertyValues/requestExecution/getLatestAssetPropertyValueRequest.ts
```diff
@@ -0,0 +1,36 @@
...
+    settings: { abortSignal: AbortSignal }
+  ) {
+    try {
+      return await this.#getAssetPropertyValue(request, settings);
+    } catch (error) {
+      this.#handleError(request, error);
+    }
+  }
+
+  #handleError(
```
hwandersman, line 26: Is this something we want to standardize across our hooks? Any reason for this more precise error handling?

packages/react-components/src/queries/useLatestAssetPropertyValues/requestExecution/batcher.ts
```diff
@@ -0,0 +1,159 @@
...
+    console.table(requests);
+
+    throw error;
+  }
+
+  private async batchLoaderFn(
+    requests: readonly LoaderRequest[]
+  ): Promise<LoaderResponse[]> {
+    let nextToken = undefined;
+    console.log(requests);
```
hwandersman, line 67: nit: remove unnecessary console logs

packages/react-components/src/queries/useLatestAssetPropertyValues/requestExecution/getLatestAssetPropertyValueRequest.ts
jmbuss: Yes, it's been the pattern in our request execution handling. This is also used to handle general errors across the entire batch or for a single error entry in the batch result.

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
jmbuss: Yes, it is handled by the invariant, but because it's an either or check they're both technically undefined. This was the easiest way to ensure that the function doesn't optionally return undefined since it's impossible for it to do so.

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
jmbuss: Yes, we may not have access to batch calls, this is the case with for edge devices. The non batch apis ensures it still works in those contexts, but we of course prefer to use batch if available.

hwandersman: Right edge! Good callout

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
jmbuss: The hook should only execute the request function if the inputs are valid. If there are no clients or the request is malformed, the hook is disabled.

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
```diff
@@ -0,0 +1,177 @@
...
+ * getAssetPropertyValue or batchGetAssetPropertyValue
+ * on a list of GetAssetPropertyValueRequest requests
+ *
+ * @param iotSiteWiseClient is an AWS SDK IoT SiteWise client
+ * @param enabled will manually disable the hook
+ * @param requests list of GetAssetPropertyValueRequest
+ * @param refreshRate refresh rate for GetAssetPropertyValueRequest in ms
+ * @param batchSechedulerTimeout configures the timeout window in ms to accept
+ * batch requests https://github.com/graphql/dataloader?tab=readme-ov-file#batch-scheduling
+ * (Note: can only be set once.)
```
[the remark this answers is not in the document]
jmbuss: Yes correct, to update the dataloader instance, it requires us to recreate it, which is a complexity I didn't think was worth handling at the moment. Maybe the right thing to do is just not expose these settings until there's a clear need for them.

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
```diff
@@ -0,0 +1,177 @@
...
+  propertyAlias,
+}: LatestAssetPropertyValueRequest) =>
+  (isAssetId(assetId) && isPropertyId(propertyId)) ||
+  isPropertyAlias(propertyAlias);
+
+const clientIsValid = ({
+  getAssetPropertyValue,
+  batchGetAssetPropertyValue,
+}: QueryFnClient = {}) =>
+  getAssetPropertyValue != null || batchGetAssetPropertyValue != null;
```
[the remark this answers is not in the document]
jmbuss: 👍 yes forgot to update this one! thanks!

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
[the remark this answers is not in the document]
hwandersman: Yeah we should set the default for now and not expose it

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
jmbuss: Added a note in the comments

packages/react-components/src/queries/useLatestAssetPropertyValues/useLatestAssetPropertyValues.ts
hwandersman: Oh got it there isn't an `isEnabled()` function, it's inline here

hwandersman (approved):

PR #2971, opened 2024-09-10
feat(react-components): use historical asset property values

jmbuss:
## Overview
Tanstack wrapper around `getAssetPropertyValueHistory`, `batchGetAssetPropertyValueHistory`. These can be used to fetch historical asset property values by assetId + propertyId or propertyAlias for a given viewport.

Note: Caching works based on an entire request in this implementation. Partial request caching is not supported. Late data configuration is also not supported. Instead, a refreshRate can be given, or one will be determined based on a percentage of the provided viewport, and will refresh the whole viewport range.

This hook will work if the client provided implements either / both getAssetPropertyValueHistory, batchGetAssetPropertyValueHistory. The hook will prefer to use the batch api if both are provided.

## Legal

### DIFF
--- a/packages/react-components/src/queries/useHistoricalAssetPropertyValues/historicalAssetPropertyValueKeyFactory.ts
+++ b/packages/react-components/src/queries/useHistoricalAssetPropertyValues/historicalAssetPropertyValueKeyFactory.ts
@@ -1,0 +1,52 @@
+import { Viewport } from '@iot-app-kit/core';
+import { HistoricalAssetPropertyValueRequest } from './types';
+import { Quality, TimeOrdering } from '@aws-sdk/client-iotsitewise';
+
+export const HISTORICAL_ASSET_PROPERTY_VALUE_RESOURCE_KEY =
+  'historical asset property value';
+
+type HistoricalAssetPropertyValueKeyFactoryOptions =
+  HistoricalAssetPropertyValueRequest & { refreshRate?: number };
+
+export class HistoricalAssetPropertyValueKeyFactory {
+  #assetId?: string;
+  #propertyId?: string;
+  #propertyAlias?: string;
+  #qualities?: Quality[];
+  #timeOrdering?: TimeOrdering;
+
+  #viewport?: Viewport;
+
+  constructor({
+    assetId,
+    propertyId,
+    propertyAlias,
+    qualities,
+    timeOrdering,
+    viewport,
+  }: HistoricalAssetPropertyValueKeyFactoryOptions) {
+    this.#assetId = assetId;
+    this.#propertyId = propertyId;
+    this.#propertyAlias = propertyAlias;
+    this.#qualities = qualities;
+    this.#timeOrdering = timeOrdering;
+
+    this.#viewport = viewport;
+  }
+
+  create() {
+    const cacheKey = [
+      {
+        resource: HISTORICAL_ASSET_PROPERTY_VALUE_RESOURCE_KEY,
+        assetId: this.#assetId,
+        propertyId: this.#propertyId,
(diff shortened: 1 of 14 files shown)


packages/react-components/src/queries/useHistoricalAssetPropertyValues/sync.ts
```diff
@@ -0,0 +1,77 @@
...
+    set((state) => ({
+      ...state,
+      enabled: {
+        ...state.enabled,
+        [refreshRate]: false,
+      },
+    })),
+});
+
+export const useHistoricalAssetPropertyValueSyncStore =
```
hwandersman, line 37: Can we reuse the same sync store between the Latest and Historical syncs? They are identical

packages/react-components/src/queries/useHistoricalAssetPropertyValues/historicalAssetPropertyValueKeyFactory.ts
```diff
@@ -0,0 +1,56 @@
...
+    refreshRate,
+  }: HistoricalAssetPropertyValueKeyFactoryOptions) {
+    this.#assetId = assetId;
+    this.#propertyId = propertyId;
+    this.#propertyAlias = propertyAlias;
+    this.#qualities = qualities;
+    this.#timeOrdering = timeOrdering;
+
+    this.#viewport = viewport;
+    this.#refreshRate = refreshRate;
```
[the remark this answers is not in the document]
jmbuss: Ya, good point. This is a code remnant from trying to sync queries using the client directly. I can remove it!

packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/batcher.ts
```diff
@@ -0,0 +1,160 @@
...
+  private batchSize: number;
+
+  private batchGetAssetPropertyValueHistory: BatchGetAssetPropertyValueHistory;
+
+  private loader: DataLoader<LoaderRequest, LoaderResponse>;
+
+  private constructor({
+    batchGetAssetPropertyValueHistory,
+    batchSchedulerTimeout = 100,
+    batchSize = 128,
```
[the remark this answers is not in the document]
jmbuss: 👍 good catch

packages/react-components/src/queries/useHistoricalAssetPropertyValues/sync.ts
jmbuss: Ya, moved this into a separate package.

packages/react-components/src/queries/utils/useTimeSeriesDataQuerySync.ts
```diff
@@ -2,18 +2,17 @@ import { useEffect } from 'react';
 import { StateCreator, create } from 'zustand';
 import { devtools } from 'zustand/middleware';
 
-export interface LatestAssetPropertyValueSync {
+export interface TimeSeriesDataSync {
```
hwandersman, line 5: There are two separate data loaders / batches for the latest and historical time series requests. Shouldn't we have two separate sync states managed in the same store here?

jmbuss: I considered that, but I thought that it is best that all time series data queries with the same refresh rate, refresh at the same time. I think that is what consumers would expect to happen. But you're right, since they are using different loaders, it doesn't actually improve performance in any way because they are different batch apis. I just wanted to ensure that data flows into charts at the same time.

hwandersman (approved):

PR #2944, opened 2024-08-12
feat: adding timezone support to dashboard/widgets

dpportet:
## Overview
* Completing implementation for https://github.com/awslabs/iot-app-kit/issues/2663
* Adding dashboard and widget level support for timeZone. Adding `timeZone` as an optional param to Dashboard, Line chart, Anomaly chart, date picker and KPI. If this param is provided, times will be displayed in that timeZone instead of the browsers time zone. If none is provided, it will default to use the users browser time zone.
* Affected components:
   * Line Chart
      * Trend Cursor header is in timezone
      * Legend time is in timezone
      * Tooltip timestamp is in timezone
      * X axis label is in timezone
   * Anomaly chart
      * Same as line chart
   * Asset explorer
      * Dates displayed in table
   * KPI
      * Timestamp is in correct timezone
   * Date picker widget
      * Display offset in provided timezone in absolute mode

## Verifying Changes
* Create a dashboard component or widget and see if the times are displayed in your local timezone. The line chart shows the UTC offset which can be useful to test this. Provide timeZone as a prop and validate the new timeZone is used.

https://github.com/user-attachments/assets/11b39d7c-23bd-413e-a52c-69bb748212be

## Legal
This project is available under the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0.html).

PR #2972, opened 2024-09-10
fix(dashboard): add timezone support for new RE

ssjagad:
## Overview
We had added timezone support in the old RE [#2944](https://github.com/awslabs/iot-app-kit/pull/2944/files). Making the new RE component have feature parity with the old.

## Verifying Changes
Local:
Screenshot 2024-09-10 at 1 37 03 PM

With Tokyo timezone:
Screenshot 2024-09-10 at 1 35 27 PM

## Legal

### DIFF
--- a/packages/react-components/src/components/resource-explorers/explorers/asset-property-explorer/internal-asset-property-explorer.tsx
+++ b/packages/react-components/src/components/resource-explorers/explorers/asset-property-explorer/internal-asset-property-explorer.tsx
@@ -26,10 +26,12 @@ import {
 import { AssetPropertyResource } from '../../types/resources';
 import {
   DEFAULT_ASSET_PROPERTY_TABLE_DEFINITION,
-  DEFAULT_ASSET_PROPERTY_WITH_LATEST_VALUES_TABLE_DEFINITION,
+  createDefaultLatestValuesTableDefinition,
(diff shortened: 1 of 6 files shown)


dpportet (approved):

packages/react-components/src/components/resource-explorers/explorers/asset-property-explorer/internal-asset-property-explorer.tsx
```diff
@@ -26,10 +26,12 @@ import {
 import { AssetPropertyResource } from '../../types/resources';
 import {
   DEFAULT_ASSET_PROPERTY_TABLE_DEFINITION,
-  DEFAULT_ASSET_PROPERTY_WITH_LATEST_VALUES_TABLE_DEFINITION,
+  createDefaultLatestValuesTableDefinition,
 } from '../../constants/table-resource-definitions';
```
corteggiano, line 30: would think of moving the file outside of the constants folder since it has functions now

ssjagad: makes sense. Ill do it for the decimal places change that follows, bc it uses the same function :)

corteggiano (approved):

PR #2976, opened 2024-09-11
fix(dashboard): fast follow to clean up messy code for table cell render

ssjagad:
## Overview
Moved render functions defined in internal property explorers into default file, wrapped in a higher order function that takes in either timezone/significant digits to render latest value or latest value time.

## Verifying Changes

https://github.com/user-attachments/assets/b52b89b8-19af-4609-89e3-424ebc6fd5b9

## Legal


corteggiano (approved):
nice! ty for cleaning these up :)

PR #2978, opened 2024-09-12
feat(react-components): add useLatestAlarmPropertyValue hook to fetch alarm prop vals in useAlarms

hwandersman:
## Overview
Add `useLatestAlarmPropertyValue` hook as an abstraction over `useLatestAssetPropertyValues` to fetch the latest alarm property value for "state", "type", or "source" properties. Hook injects the property data into the running AlarmData list.

Decided to make `useLatestAlarmPropertyValue` generic with a field for the `alarmPropertyFieldName` (`'state' | 'type' | 'source'`). There is a predicate to check which AlarmData of the running list has that property field, and fetches the latest data for each. The hook is self-contained and will only cause a re-render if that property field changes.

`useAlarms` builds AlarmData by passing the output of one hook into the other.

* Minor: renamed the input for `useDescribeAssets` and `useDescribeAssetModels` to be `requests` to standardize across all hooks

## Verifying Changes

Added unit tests and ran in storybook to see alarm output successfully.

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal

### DIFF
--- a/packages/react-components/src/hooks/useAlarms/utils/queryUtils.ts
+++ b/packages/react-components/src/hooks/useAlarms/utils/queryUtils.ts
@@ -2,6 +2,28 @@ import { UseQueryResult } from '@tanstack/react-query';
 import { AlarmDataStatus } from '../types';
 
 /**
+ * Combine two query statuses
+ *
+ * If any one status isLoading/Refetching/Error is true then the running status is also true
+ * If any one status isSuccess is false then running status is also false
+ * Combine all errors into a running list
+ */
+// Combine two query statuses
+const combineStatuses = ({
+  oldStatus,
+  newStatus,
+}: {
+  oldStatus: AlarmDataStatus;
+  newStatus: AlarmDataStatus;
+}): AlarmDataStatus => ({
+  isLoading: oldStatus.isLoading || newStatus.isLoading,
+  isRefetching: oldStatus.isRefetching || newStatus.isRefetching,
+  isSuccess: oldStatus.isSuccess && newStatus.isSuccess,
+  isError: oldStatus.isError || newStatus.isError,
+  errors: [...(oldStatus.errors ?? []), ...(newStatus.errors ?? [])],
+});
+
+/**
  * getStatusForQuery builds the AlarmDataStatus for a tanstack query result.
  * Since AlarmData is built on multiple queries the status can be combined with a previous query.
  *
@@ -13,7 +35,7 @@ export const getStatusForQuery = (
   query: UseQueryResult,
   oldStatus?: AlarmDataStatus
 ): AlarmDataStatus => {
-  let errors: Error[] | undefined =
+  const errors: Error[] | undefined =
     query.error !== null ? [query.error] : undefined;
   let status: AlarmDataStatus = {
     isLoading: query.isLoading,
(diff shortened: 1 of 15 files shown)


packages/react-components/src/hooks/useAlarms/utils/queryUtils.ts
```diff
@@ -1,6 +1,28 @@
...
 
+/**
+ * Combine two query statuses
+ *
+ * If any one status isLoading/Refetching/Error is true then the running status is also true
+ * If any one status isSuccess is false then running status is also false
+ * Combine all errors into a running list
+ */
+// Combine two query statuses
+const combineStatuses = ({
```
jmbuss, line 12: Do we need to care about old statuses? Won't we always want to use the latest value?

packages/react-components/src/hooks/useAlarms/utils/queryUtils.ts
```diff
@@ -1,6 +1,28 @@
...
+  newStatus,
+}: {
+  oldStatus: AlarmDataStatus;
+  newStatus: AlarmDataStatus;
+}): AlarmDataStatus => ({
+  isLoading: oldStatus.isLoading || newStatus.isLoading,
+  isRefetching: oldStatus.isRefetching || newStatus.isRefetching,
+  isSuccess: oldStatus.isSuccess && newStatus.isSuccess,
+  isError: oldStatus.isError || newStatus.isError,
+  errors: [...(oldStatus.errors ?? []), ...(newStatus.errors ?? [])],
```
jmbuss, line 23: What do you think about having a single error field too which is a concatenation of all the errors into a single new error? The resource explorer variants take a single error object, so I ended up doing that concatenation. But I'm wondering if other consumers like the charts will also prefer that convenience

jmbuss (approved):

packages/react-components/src/hooks/useAlarms/utils/queryUtils.ts
hwandersman: Consider this case:
1. DescribeAsset succeeds
`{ isLoading: false, isRefetching: false, isSuccess, true, isError, false, errors: [] }`
2. GetLatest state fails
`{ isLoading: false, isRefetching: false, isSuccess, false, isError, true, errors: [<Error1>] }`
3. GetLatest type succeeds
`{ isLoading: false, isRefetching: false, isSuccess, false, isError, true, errors: [<Error1>] }`
4. GetLatest source fails
`{ isLoading: false, isRefetching: false, isSuccess, false, isError, true, errors: [<Error1>, <Error2>] }`

The new status at every step needs to resolve based on the old status to compile all of the errors and have the most accurate status.

We will still return AlarmData but it will have errors and isSuccess is false even if some queries were successful. Without caring about the old status we wouldn't have an accurate status for all of the queries.

packages/react-components/src/hooks/useAlarms/utils/queryUtils.ts
hwandersman: We should discuss what is the best practice for handling multiple errors. Do we provide like a stack trace of every error in one message? Do we just dump a list of errors as a single object?

PR #2986, opened 2024-09-13
fix(alarms): historical asset property value requests

jmbuss:
## Overview
Add configuration options to `useHistoricalAssetPropertyValues` 
* option to request most recent before start or end values.
* option to configure how many points to request

## Legal

### DIFF
--- a/packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/batchResponseProcessor.ts
+++ b/packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/batchResponseProcessor.ts
@@ -1,0 +1,80 @@
+import { BatchGetAssetPropertyValueHistoryEntry } from '@aws-sdk/client-iotsitewise';
+import {
+  BatchGetAssetPropertyValueHistory,
+  GetAssetPropertyValueHistory,
+  RequestResponse,
+} from '@iot-app-kit/core';
+import { take } from 'lodash';
+
+export class BatchResponseProcessor {
+  private entry: BatchGetAssetPropertyValueHistoryEntry;
+  private valuesTargetNumber: number;
+
+  private assetPropertyValueHistory: NonNullable<
+    RequestResponse<GetAssetPropertyValueHistory>['assetPropertyValueHistory']
+  > = [];
+
+  private errorEntries: RequestResponse<BatchGetAssetPropertyValueHistory>['errorEntries'] =
+    [];
+
+  constructor(
+    entry: BatchGetAssetPropertyValueHistoryEntry,
+    valuesTargetNumber = Infinity
+  ) {
+    this.entry = entry;
+    this.valuesTargetNumber = valuesTargetNumber;
+  }
+
+  processResponse(
+    response: RequestResponse<BatchGetAssetPropertyValueHistory>
+  ) {
+    if (this.isComplete) return;
+
+    const sucessesForEntry = (response.successEntries ?? []).filter(
+      ({ entryId }) => entryId === this.entry.entryId
+    );
+    const errorsForEntry = response.errorEntries?.filter(
+      ({ entryId }) => entryId === this.entry.entryId
+    );
+
+    const assetPropertyValues = sucessesForEntry.flatMap(
+      ({ assetPropertyValueHistory = [] }) => assetPropertyValueHistory
+    );
+
+    /**
+     * Only take the number of values determined by the target number
+     * The target number will be a running count from each paginated
+     * resposne.
+     */
+    const limitedAssetPropertyValues = take(
+      assetPropertyValues,
+      this.valuesTargetNumber
+    );
+    this.valuesTargetNumber -= assetPropertyValues.length;
+
+    if (this.entry.timeOrdering === 'ASCENDING') {
+      this.assetPropertyValueHistory.push(...limitedAssetPropertyValues);
+    } else {
+      this.assetPropertyValueHistory.unshift(
+        ...limitedAssetPropertyValues.reverse()
+      );
+    }
+
+    this.errorEntries = [
+      ...(this.errorEntries ?? []),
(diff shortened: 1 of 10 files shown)


packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/batchResponseProcessor.ts
```diff
@@ -0,0 +1,80 @@
...
+    const limitedAssetPropertyValues = take(
+      assetPropertyValues,
+      this.valuesTargetNumber
+    );
+    this.valuesTargetNumber -= assetPropertyValues.length;
+
+    if (this.entry.timeOrdering === 'ASCENDING') {
+      this.assetPropertyValueHistory.push(...limitedAssetPropertyValues);
+    } else {
+      this.assetPropertyValueHistory.unshift(
```
hwandersman, line 58: We always keep the same order no matter what?

packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/getHistoricalAssetPropertyValueRequest.ts
```diff
@@ -1,33 +1,91 @@
...
-  viewportEndDate,
-  viewportStartDate,
+  RequestResponse,
 } from '@iot-app-kit/core';
 import { HistoricalAssetPropertyValueRequest } from '../types';
+import { take } from 'lodash';
+import { mapViewport } from './mapViewport';
+import { mapTimeOrdering } from './mapTimeOrdering';
+
+const DEFAULT_MAX_RESULTS = 20000;
```
hwandersman, line 10: nit: this is already defined in the batcher file

packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/getHistoricalAssetPropertyValueRequest.ts
```diff
@@ -1,33 +1,91 @@
...
+         * resposne.
+         */
+        const limitedAssetPropertyValues = take(
+          assetPropertyValues,
+          this.valuesTargetNumber
+        );
+        this.valuesTargetNumber -= assetPropertyValues.length;
+
+        /**
+         * Always return a list in ascending order
```
hwandersman, line 77: Why always return list in ascending order?

packages/react-components/src/queries/useHistoricalAssetPropertyValues/useHistoricalAssetPropertyValues.ts
```diff
@@ -167,9 +186,11 @@ export const createHistoricalAssetPropertyValueQueryFn = (
       'Expected assetId and propertyId or propertyAlias and viewport to be defined as required by the enabled flag.'
     );
 
-    if (batchGetAssetPropertyValueHistory) {
+    const test = true;
```
hwandersman, line 189: nit: remove test code

packages/react-components/src/queries/useHistoricalAssetPropertyValues/useRefreshRate.ts
```diff
@@ -29,11 +32,13 @@ export const useRefreshRate = ({
   refreshRate,
   viewport,
   minimumRefreshRate = 5000,
```
hwandersman, line 34: Can the minimum refresh rate be smaller? We'll probably have to stress test this

packages/react-components/src/queries/useHistoricalAssetPropertyValues/types.ts
```diff
@@ -13,10 +13,15 @@ export type QueryFnClient = {
   batchGetAssetPropertyValueHistory?: BatchGetAssetPropertyValueHistory;
 };
 
+export type FetchMode =
+  | 'START_TO_END'
+  | 'MOST_RECENT_BEFORE_START'
```
hwandersman, line 18: Why do we need both most recent before start and end?

jmbuss: I wanted to include this so that it supported the same features as the existing datasource implementation. We don't need it for alarms though.

packages/react-components/src/queries/useHistoricalAssetPropertyValues/useRefreshRate.ts
jmbuss: It can be smaller, the minimum in general we support is 1s. I can make it that.

packages/react-components/src/queries/useHistoricalAssetPropertyValues/useHistoricalAssetPropertyValues.ts
jmbuss: 👍 thanks for catching that.

packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/getHistoricalAssetPropertyValueRequest.ts
jmbuss: It will be important that the points are sorted in a consistent way so that we can do operations on them quickly. For instance, if we wanted to filter points only contained within the viewport, we can do it with binary search instead of running a standard filter operation.

packages/react-components/src/queries/useHistoricalAssetPropertyValues/requestExecution/getHistoricalAssetPropertyValueRequest.ts
jmbuss: We can make a constants file, but since these are different apis, they could potentially support different max results.

hwandersman (approved):

PR #2990, opened 2024-09-16
fix(alarms): return blank alarm data when queries are loading

jmbuss:
## Overview
Bugfix for `useAlarmAssets`. If the `useDescribeAssets / useDescribeAssetModels` queries were loading for the first time, their response will be undefined. In that case, we were not returning any AlarmData objects. This means that we could not correctly access the `isLoading` flags. The hook will now return an empty AlarmData object with the query status if the `useDescribeAssets / useDescribeAssetModels` queries responses are undefined. 

## Legal

### DIFF
--- a/packages/react-components/src/hooks/useAlarms/utils/alarmDataUtils.ts
+++ b/packages/react-components/src/hooks/useAlarms/utils/alarmDataUtils.ts
@@ -27,10 +27,18 @@ export const buildFromAssetResponse = ({
   status: AlarmDataStatus;
   assetResponse?: DescribeAssetResponse;
 }): AlarmData[] => {
-  const { assetId, assetCompositeModelId, inputPropertyId } = request;
+  const { assetId, assetCompositeModelId, inputPropertyId, assetModelId } =
+    request;
 
   if (!assetResponse || !assetId) {
-    return [];
+    return [
+      {
+        assetModelId,
+        assetId,
+        compositeModelId: assetCompositeModelId,
+        status,
+      },
+    ];
   }
 
   if (assetResponse && assetId && assetResponse?.assetId !== assetId) {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/hooks/useAlarms/utils/alarmDataUtils.ts
```diff
@@ -27,10 +27,19 @@ export const buildFromAssetResponse = ({
   status: AlarmDataStatus;
   assetResponse?: DescribeAssetResponse;
 }): AlarmData[] => {
-  const { assetId, assetCompositeModelId, inputPropertyId } = request;
+  const { assetId, assetCompositeModelId, inputPropertyId, assetModelId } =
+    request;
 
   if (!assetResponse || !assetId) {
-    return [];
+    const alarmData = buildFromCompositeModel(undefined, status);
```
hwandersman, line 34: Do we need to call `buildFromCompositeModel` here? We essentially only need to add the status field.

packages/react-components/src/hooks/useAlarms/utils/alarmDataUtils.ts
```diff
@@ -27,10 +27,18 @@ export const buildFromAssetResponse = ({
...
 }): AlarmData[] => {
-  const { assetId, assetCompositeModelId, inputPropertyId } = request;
+  const { assetId, assetCompositeModelId, inputPropertyId, assetModelId } =
+    request;
 
   if (!assetResponse || !assetId) {
-    return [];
+    return [
+      {
+        assetModelId,
```
hwandersman, line 36: nit: there will never be an assetModelId in the request here cause it would have been filtered out

hwandersman (approved):

PR #2998, opened 2024-09-18
feat(alarms): use alarm state hook

jmbuss:
## Overview
Refactor alarm state values into its own hook. The hook can handle multiple contexts:
1. fetching the latest alarm state without a viewport (Resource Explorer)
2. fetching the latest alarm state for a given viewport (KPI, Status, Dial, Gauge)
3. fetching alarm state for a historical viewport (Line Chart)
4. fetching alarm state for a live viewport (Line Chart)

## Legal

### DIFF
--- a/packages/react-components/src/hooks/useAlarms/hookHelpers/useAlarmState/updateAlarmStateValues.ts
+++ b/packages/react-components/src/hooks/useAlarms/hookHelpers/useAlarmState/updateAlarmStateValues.ts
@@ -1,0 +1,80 @@
+import { AssetPropertyValue } from '@aws-sdk/client-iotsitewise';
+import { AlarmData } from '../../types';
+import { toTimestamp } from '../../../../utils/time';
+import { bisector } from 'd3-array';
+import { Viewport, timeSeriesDataFilterer } from '../../../../queries';
+import { viewportEndDate, viewportStartDate } from '@iot-app-kit/core';
+
+const assetPropertyValueTime = (assetPropertyValue: AssetPropertyValue) =>
+  toTimestamp(assetPropertyValue.timestamp);
+
+const compareAssetPropertyValues = (
+  a: AssetPropertyValue,
+  b: AssetPropertyValue
+) => {
+  return assetPropertyValueTime(a) - assetPropertyValueTime(b);
+};
+
+const assetPropertyValuesBisector = bisector(assetPropertyValueTime);
+
+const filterAssetPropertyValues = timeSeriesDataFilterer(
+  assetPropertyValuesBisector,
+  assetPropertyValueTime
+);
+
+const uniqueSortAssetPropertyValues = (
+  assetPropertyValues: AssetPropertyValue[]
+) => {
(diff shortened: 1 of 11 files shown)


packages/react-components/src/hooks/useAlarms/hookHelpers/useAlarmState/updateAlarmStateValues.ts
```diff
@@ -0,0 +1,80 @@
...
+  { data, viewport }: { data: AssetPropertyValue[]; viewport?: Viewport }
+): AlarmData => {
+  /**
+   * If there is no state property the queries will
+   * be disabled and data will be empty
+   */
+  if (!alarm.state) return alarm;
+
+  const currentData = alarm.state.data ?? [];
+  const updatedData = uniqueSortAssetPropertyValues([...currentData, ...data]);
```
hwandersman, line 70: Are we supposed to append data in every case? This will lead to longer and longer lists of data and more sorting as charts are panned. We should really only append data if the queryMode is `'LIVE'`.

packages/react-components/src/hooks/useAlarms/useAlarms.ts
```diff
@@ -57,10 +64,11 @@ function useAlarms<T>(options?: UseAlarmsOptions<T>): (T | AlarmData)[] {
    * Fetch latest asset property values for alarms with a state property.
    * Data should be available for all alarms fetched for an asset.
    */
-  const statePropertyAlarmData = useLatestAlarmPropertyValue({
+  const statePropertyAlarmData = useAlarmState({
```
hwandersman, line 67: If we still wanna do these cascading hooks we need to move this to the end. Since not all alarms fetched here are alarms we want to return, they are filtered based on the inputProperty request.

packages/react-components/src/hooks/useAlarms/hookHelpers/useAlarmState/updateAlarmStateValues.ts
jmbuss: I think it is fine to append in every case. This reduces complexity because we don't need to handle 4 explicit cases. The data will never contain any points that lie outside the viewport since we do `filterAssetPropertyValues` if the viewport exists. So the data list won't grow larger than the required amount to display in the chart as the charts are panned.

packages/react-components/src/hooks/useAlarms/useAlarms.ts
jmbuss: Yes that's a really good point. Let's address that in a separate change though? I know we are going to have to modify how we are managing the alarm data state too. we can likely combine those two updates.

packages/react-components/src/hooks/useAlarms/hookHelpers/useAlarmState/updateAlarmStateValues.ts
hwandersman: Ah right the filterer ensures the list doesn't grow, sounds good

hwandersman (approved):

PR #2999, opened 2024-09-19
feat(alarms): support alarms in resource explorer and in widget types

corteggiano:
## Overview
- useQueries in dashboard now returns both timeSeriesData and alarm queries
    - due to above change, all react-component APIs had to be updated to accept timeseries OR alarm queries
    - also needed to filter queries passed into useTimeSeriesData until UX is ready
- updated resource explorer to display alarm data (gated behind USE_ALARMS)

## Legal

### DIFF
--- a/packages/dashboard/src/components/dashboard/queryContext.ts
+++ b/packages/dashboard/src/components/dashboard/queryContext.ts
@@ -1,11 +1,41 @@
-import { DataStream, Primitive } from '@iot-app-kit/core';
 import { createContext, useContext } from 'react';
 import { useRefreshRate } from '~/customization/hooks/useRefreshRate';
+import { alarmModelQueryToSiteWiseAssetQuery } from '~/customization/widgets/utils/alarmModelQueryToAlarmQuery';
 import { assetModelQueryToSiteWiseAssetQuery } from '~/customization/widgets/utils/assetModelQueryToAssetQuery';
-import {
+
+import type {
+  DataStream,
+  Primitive,
+  TimeSeriesDataQuery,
+} from '@iot-app-kit/core';
+import type {
+  AlarmDataQuery,
+  SiteWiseDataStreamQuery,
+  SiteWiseQuery,
+  SiteWiseAlarmDataStreamQuery,
+} from '@iot-app-kit/source-iotsitewise';
+import type {
   DashboardIotSiteWiseQueries,
   IoTSiteWiseDataStreamQuery,
 } from '~/types';
+
+const createTimeSeriesDataQuery = (
+  iotSiteWiseQuery: SiteWiseQuery,
+  { assets = [], properties = [], requestSettings }: SiteWiseDataStreamQuery
+) => {
+  if (assets.length === 0 && properties.length === 0) return [];
+  return [
+    iotSiteWiseQuery.timeSeriesData({ assets, properties, requestSettings }),
+  ];
+};
+
+const createAlarmDataQuery = (
+  iotSiteWiseQuery: SiteWiseQuery,
+  { alarms = [], requestSettings }: SiteWiseAlarmDataStreamQuery
+) => {
+  if (alarms.length === 0) return [];
+  return [iotSiteWiseQuery.alarmData({ alarms, requestSettings })];
+};
 
 export const QueryContext = createContext<Partial<DashboardIotSiteWiseQueries>>(
   {}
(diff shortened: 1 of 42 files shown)


packages/dashboard/src/customization/widgets/barChart/component.tsx
```diff
@@ -47,7 +48,7 @@ const BarChartWidgetComponent: React.FC<BarChartWidget> = (widget) => {
       <BarChart
         chartSize={size}
         key={key}
-        queries={queries}
+        queries={queries.filter((q) => 'build' in q) as TimeSeriesDataQuery[]}
```
hwandersman, line 51: This goes for all of the widget query/queries fields: I see you updated the interface to accept `TimeSeriesDataQuery | AlarmDataQuery`, so why do we need to filter them here? Shouldn't we be filtering the queries in the widget component itself when it passes them to `useTimesSeriesData` and `useAlarms`?

packages/dashboard/src/components/dashboard/queryContext.ts
```diff
@@ -31,13 +39,28 @@ export const useQueries = ({
...
+  }
+
+  // if the mapped query has timeSeries queries
+  if (
+    mappedQuery.assets.length ||
+    mappedQuery.assetModels?.length ||
+    mappedQuery.properties?.length
+  ) {
+    queries.push(iotSiteWiseQuery.timeSeriesData(mappedQuery));
+  }
```
tracy-french: `assetModelQueryToSiteWiseAssetQuery` only changes the `assets` field of the query. Can we avoid using `mappedQuery.alarms`, `mappedQuery.assetModels`, and `mappedQuery.properties`? This would avoid you having to do the nullish coalescing since `alarms`, `assetModels`, and `properties` have already been defaulted as empty arrays. 

Similarly, `assetModelQueryToSiteWiseAssetQuery` should not take the whole query, where `query.assetModels` and `query.assets` may be undefined. It should take `assetModels` and `assets` as required and always return an array of asset queries, like:

```ts
export const assetModelQueryToSiteWiseAssetQuery({ 
  assetModels, 
  assets 
}: { 
  assetModels: WhateverAssetModelQueryTypeIs[]; 
  assets: WhateverAssetQueryTypeIs[]; 
}): WhateverAssetQueryTypeIs[] {}
```

If you do all this, then you don't need to use any if statements or nullish coalescing, and you can avoid pushing on the array, enabling more declarative code to be written, like:

```ts
const queries = [
  ...iotSiteWiseQuery.alarmData({ .. }),
  ...iotSiteWiseQuery.timeSeriesData({ .. }),
]
```

tracy-french (changes requested):

packages/dashboard/src/components/dashboard/queryContext.ts
```diff
@@ -1,4 +1,5 @@
-import { DataStream, Primitive } from '@iot-app-kit/core';
+import { DataStream, Primitive, TimeSeriesDataQuery } from '@iot-app-kit/core';
```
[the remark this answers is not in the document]
jmbuss: 👍  I wish the intellisense could do this automatically, or our linter...

packages/react-components/src/components/status/status.tsx
```diff
@@ -19,7 +20,7 @@ export const Status = ({
   settings,
   significantDigits,
 }: {
-  query: TimeQuery<TimeSeriesData[], TimeSeriesDataRequest>;
+  query: TimeQuery<TimeSeriesData[], TimeSeriesDataRequest> | AlarmDataQuery;
```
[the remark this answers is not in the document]
corteggiano: no they're the same! will move them to a shared type

packages/dashboard/src/components/dashboard/queryContext.ts
jmbuss: I don't know if this will be possible since `iotSiteWiseQuery.alarmData` always returns an object. If the query passed to that function is undefined, the return value will be an object with `query: undefined`. But what we could do is something like this:

```
const queries = [
  ...[...assets, ...properties].map(query => iotSiteWiseQuery.timeSeriesData(query)),
  ...[...alarms].map(query => iotSiteWiseQuery.alarmData(query)),
]
```

jmbuss: But yes, I think those assetModel / alarmAssetModel transformers could just take assetModel + assets / alarmAssetModel + alarms, and out only assets / alarms. Since we don't use assetModel / alarmAssetModel after those functions.

PR #3000, opened 2024-09-19
chore: update cloudscape to more recent version

ssjagad:
## Overview
Updating cloudscape versions to more recent versions. Header component now takes in a variant prop that allows us to keep the background black instead of the default white.

Updated snapshot testing and one unit test.

## Verifying Changes
Screenshot 2024-09-19 at 1 45 30 PM

## Legal


packages/react-components/src/components/chart/legend/table/table.spec.tsx
```diff
@@ -91,14 +91,14 @@ describe('legend table', () => {
 
     expect(table).not.toBeNull();
     expect(
-      screen.getByText(
+      screen.getAllByText(
```
corteggiano, line 94: curious, what changed to make these tests fail?

tracy-french (approved):

corteggiano (approved):

PR #3003, opened 2024-09-23
feat(bar chart alarms): add alarms to bar chart

corteggiano:
## Overview

Add alarms to the bar chart by using the synchr-charts threshold props 

https://github.com/user-attachments/assets/1b5b97d4-3a16-4550-9f61-630cd78c0347

custom alarm only:
Screenshot 2024-09-25 at 14 11 15

demo alarm only:
Screenshot 2024-09-25 at 14 11 01

zoomed in so label isnt hidden: 
image

Tests:

| Test    | Outcome |
| -------- | ------- |
| Custom alarm  | Alarm renders properly |
| Demo alarm | Alarm renders properly     |
| Custom and Demo Alarm    | Both alarms render properly    |

## Legal

### DIFF
--- a/packages/react-components/src/components/bar-chart/barChart.tsx
+++ b/packages/react-components/src/components/bar-chart/barChart.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+/* eslint-disable react-hooks/exhaustive-deps */
+import React, { useMemo } from 'react';
 import {
   StyleSettingsMap,
   Threshold,
@@ -17,12 +18,15 @@ import {
   DEFAULT_VIEWPORT,
   ECHARTS_GESTURE,
 } from '../../common/constants';
-import type {
+import { getAlarmQueries, getTimeSeriesQueries } from '../../utils/queries';
+import {
   AxisSettings,
   ChartSize,
   ComponentQuery,
 } from '../../common/chartTypes';
-import { getTimeSeriesQueries } from '../../utils/queries';
+import { useAlarms } from '../../hooks/useAlarms';
+import { convertAlarmQueryToAlarmRequest } from '../../queries/utils/convertAlarmQueryToAlarmRequest';
+import { transformAlarmsToThreshold } from '../../utils/transformAlarmsToThreshold';
 
 const HOUR_IN_MS = 1000 * 60 * 60;
 const DAY_IN_MS = HOUR_IN_MS * 24;
@@ -56,9 +60,38 @@ export const BarChart = (props: BarChartProps) => {
     ...rest
   } = props;
 
+  const { viewport, setViewport, group, lastUpdatedBy } = useViewport();
+
+  // if using echarts then echarts gesture overrides passed in viewport
+  // else explicitly passed in viewport overrides viewport group
+  const utilizedViewport =
+    (lastUpdatedBy === ECHARTS_GESTURE
+      ? viewport
+      : passedInViewport || viewport) ?? DEFAULT_VIEWPORT;
+
+  const alarmQueries = getAlarmQueries(queries);
+  const timeSeriesQueries = getTimeSeriesQueries(queries);
+
+  const mapAlarmQueriesToRequests = alarmQueries.flatMap((query) =>
+    convertAlarmQueryToAlarmRequest(query)
+  );
+
+  const transformedAlarms = useAlarms({
+    iotSiteWiseClient: alarmQueries.at(0)?.iotSiteWiseClient,
+    iotEventsClient: alarmQueries.at(0)?.iotEventsClient,
+    requests: mapAlarmQueriesToRequests,
+    viewport: utilizedViewport,
+    settings: {
+      fetchThresholds: true,
+      refreshRate: alarmQueries.at(0)?.query.requestSettings?.refreshRate,
+    },
+    transform: transformAlarmsToThreshold,
+  });
+  const filteredAlarms = transformedAlarms.filter((alarm) => !!alarm);
+
   const { dataStreams, thresholds: queryThresholds } = useTimeSeriesData({
     viewport: passedInViewport,
-    queries: getTimeSeriesQueries(queries),
+    queries: timeSeriesQueries,
     settings: {
       fetchFromStartToEnd: true,
       fetchMostRecentBeforeStart: true,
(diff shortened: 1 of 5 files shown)


packages/react-components/src/components/bar-chart/barChart.tsx
```diff
@@ -56,9 +59,33 @@ export const BarChart = (props: BarChartProps) => {
...
+
+  const alarmQueries = getAlarmQueries(queries);
+  const timeSeriesQueries = getTimeSeriesQueries(queries);
+
+  const mapAlarmQueriesToRequests = alarmQueries.flatMap((query) =>
+    convertAlarmQueryToAlarmRequest(query)
+  );
+
+  const transformedAlarms = useAlarms({
+    iotSiteWiseClient: alarmQueries[0]?.iotSiteWiseClient,
```
hwandersman, line 78: A safer way to access the query would be:
```
alarmQueries.find(Boolean)
```

This gets you the first item on the array if it exists

packages/react-components/src/components/bar-chart/barChart.tsx
```diff
@@ -56,9 +59,33 @@ export const BarChart = (props: BarChartProps) => {
...
+    iotEventsClient: alarmQueries[0]?.iotEventsClient,
+    requests: mapAlarmQueriesToRequests,
+    viewport,
+    settings: { fetchThresholds: true },
```
hwandersman, line 82: Need to set the refreshRate here as well

packages/react-components/src/components/bar-chart/barChart.tsx
```diff
@@ -56,9 +59,33 @@ export const BarChart = (props: BarChartProps) => {
     ...rest
   } = props;
 
+  const {
+    viewport: injectedViewport,
+    setViewport,
+    group,
+    lastUpdatedBy,
+  } = useViewport();
+  const viewport = passedInViewport || injectedViewport || DEFAULT_VIEWPORT;
```
hwandersman, line 68: What was the reason to update the viewport logic here?

packages/react-components/src/utils/transformAlarmThresholdToSynchroChartsThreshold.ts
```diff
@@ -0,0 +1,62 @@
...
+  thresholdValue,
+  severity,
+  labelText,
+  comparisonOperator,
+}: {
+  thresholdValue: number;
+  severity: number | undefined;
+  labelText: string;
+  comparisonOperator: COMPARISON_OPERATOR;
+}): YAnnotation => {
```
hwandersman, line 16: Any reason we use `YAnnotation` here and not `Threshold` type? That's the type on `BarChartProps` and many other widgets.

packages/react-components/src/components/bar-chart/barChart.tsx
hwandersman: Or
```
alarmQueries.at(0)
```

packages/react-components/src/utils/transformAlarmThresholdToSynchroChartsThreshold.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const transformAlarmsToSynchroChartsThreshold = (
+  alarm: AlarmData
+): YAnnotation | undefined => {
+  const { models, inputProperty, thresholds, status } = alarm;
+  const modelsFound = models && models.length !== 0;
+  const thresholdsFound = thresholds && thresholds?.length !== 0;
+
+  if (!status.isSuccess) return undefined;
```
hwandersman, line 37: Don't think we should be purposefully returning or setting `undefined` anywhere. Instead you can just wrap everything in the if statement `if (state.isSuccess)` and just not return anything otherwise.

packages/react-components/src/utils/transformAlarmThresholdToSynchroChartsThreshold.ts
corteggiano: YAnnotation is the type from synchrocharts itself, if you look [here](https://github.com/awslabs/iot-app-kit/pull/3003/files/3d4be9a199498f310f1852fd331f88672aabbc92#diff-a635ab59949bd78e56473e361f379b8f7a3f9ec85d098ad7a75c2edb221a0eacR128) you can see we actually cast the barChart thresholds to this type

packages/react-components/src/components/bar-chart/barChart.tsx
corteggiano: this is the same logic used inside useTimeSeriesData to select what viewport to use, but now that i think about it we should use the [utilizedViewport](https://github.com/awslabs/iot-app-kit/pull/3003/files/3d4be9a199498f310f1852fd331f88672aabbc92#diff-a635ab59949bd78e56473e361f379b8f7a3f9ec85d098ad7a75c2edb221a0eacR111) from this same component instead

packages/react-components/src/components/bar-chart/barChart.tsx
```diff
@@ -72,8 +99,12 @@ export const BarChart = (props: BarChartProps) => {
     },
     styles,
   });
-  const { viewport, setViewport, group, lastUpdatedBy } = useViewport();
-  const allThresholds = [...queryThresholds, ...thresholds];
+
+  const allThresholds = [
```
hwandersman, line 103: Is there any need to memoize this?

packages/react-components/src/utils/transformAlarmsToThreshold.ts
```diff
@@ -0,0 +1,61 @@
...
+      const comparisonOperator =
+        model.alarmRule?.simpleRule?.comparisonOperator;
+
+      if (thresholdValue && comparisonOperator && inputProperty) {
+        const scComparisonOperator = IoTEventsToSynchroChartsComparisonOperator[
+          comparisonOperator
+        ] as COMPARISON_OPERATOR;
+
+        return createThreshold({
+          labelText: `${inputProperty[0].name} ${COMPARATOR_MAP[scComparisonOperator]} ${thresholdValue}`,
```
hwandersman, line 53: Need to check that inputProperty has content or use `.at(0)`

hwandersman (dismissed):

ssjagad (approved):

PR #3004, opened 2024-09-24
fix(alarm thresholds): only filter threshold values past the viewport end date

corteggiano:
## Overview
Static alarm thresholds were being filtered when a viewport was provided. This change updates the logic to always prioritze static alarm threshold values. Thresholds were being filtered to use anything within viewport.

Changes: 
1. prioritize static thresholds
2. always create `mostRecentBeforeEndValueQueries` as long as queryMode !== latest
3. filter all thresholdData that is after the viewport end

### test cases + outcomes

| Alarm(s) | Viewport | Outcome |
| -------- | ------- | ------- |
| custom alarm property with a static threshold | last 10 minutes | static threshold is used |
| custom alarm property with a static threshold | none | static threshold is used |
| demo alarm with ingested threshold | last 10 minutes | will fetch and use last threshold before end of viewport |
| demo alarm with ingested threshold | none | will fetch and use last threshold before end of viewport |
| custom alarm property with a static threshold value + demo alarm with ingested threshold | last 10 minutes | will use static value for custom alarm and use last threshold before end of viewport |
| custom alarm property with a static threshold value + demo alarm with ingested threshold | none | will use static value for custom alarm and use last threshold before end of viewport |

example: fetching custom alarm with defined viewport
```
"thresholds": [
      {
        "value": {
          "doubleValue": 10
        },
        "timestamp": {
          "timeInSeconds": 1727122118
        }
      }
    ]
```

example: fetching demo data alarm with defined viewport
```
"thresholds": [
      {
        "quality": "GOOD",
        "timestamp": {
          "offsetInNanos": 0,
          "timeInSeconds": 1727121673
        },
        "value": {
          "doubleValue": 30
        }
      }
    ]
```

## Legal

### DIFF
--- a/packages/react-components/src/hooks/useAlarms/utils/updateAlarmValues.ts
+++ b/packages/react-components/src/hooks/useAlarms/utils/updateAlarmValues.ts
@@ -2,8 +2,9 @@ import { AssetPropertyValue } from '@aws-sdk/client-iotsitewise';
 import type { AlarmData } from '../types';
 import { toTimestamp } from '../../../utils/time';
 import { bisector } from 'd3-array';
-import { Viewport, timeSeriesDataFilterer } from '../../../queries';
+import { timeSeriesDataFilterer, Viewport } from '../../../queries';
 import { viewportEndDate, viewportStartDate } from '@iot-app-kit/core';
+import { alarmThresholdValueFilterer } from './alarmThresholdValueFilterer';
 
 const assetPropertyValueTime = (assetPropertyValue: AssetPropertyValue) =>
   toTimestamp(assetPropertyValue.timestamp);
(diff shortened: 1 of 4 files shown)


packages/react-components/src/hooks/useAlarms/utils/updateAlarmValues.ts
```diff
@@ -90,6 +98,18 @@ export const updateAlarmThresholdData = (
   if (!alarm.models) return alarm;
 
   const currentData = alarm.thresholds ?? [];
+
+  /**
+   * If there is a static threshold then that should be used as the alarm threshold
+   **/
+  if (
```
hwandersman, line 105: I think the root of the problem is that we are filtering out data outside the viewport when we still need to show a datapoint if its the latest before the end of the viewport. So instead of this change let's look into fixing the filtering for alarm state and threshold data.

hwandersman (approved):

PR #3009, opened 2024-09-26
fix(dashboard): state fixes for dynamic assets tab

ssjagad:
## Overview
Fix for screen jumping to top when selecting dashboard model properties. To fix, removed useState call and moved selectedAsset state up to parent QueryEditor component.

In doing so found the following bugs and fixed them as well.

- Asset Model RE keeps state of old selected asset properties after changing asset
   - fix: added clearing selected properties to useEffect of when asset changes
   
- Asset Properties table does not change selection mode based on widget type
   - fix: passed in missing 'correctSelectionMode' prop into table
  
- Asset Model Tab breaks if you add duplicate properties after edit/preview change
  - deduped properties in hook where we reduce properties to ids for later use

## Verifying Changes

https://github.com/user-attachments/assets/69206819-d696-496e-921a-ed3e73f28206

## Legal

### DIFF
--- a/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/assetModelDataStreamExplorer.tsx
+++ b/packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/assetModelDataStreamExplorer.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { type IoTSiteWise } from '@aws-sdk/client-iotsitewise';
+import type { AssetSummary, IoTSiteWise } from '@aws-sdk/client-iotsitewise';
 
 import Box from '@cloudscape-design/components/box';
 
@@ -19,7 +19,6 @@ import {
   useSelectedAsset,
 } from './useSelectedAsset';
 import { useModelBasedQuery } from './modelBasedQuery/useModelBasedQuery';
-import { getAssetModelQueryInformation } from './getAssetModelQueryInformation';
 import { useModelBasedQuerySelection } from './modelBasedQuery/useModelBasedQuerySelection';
 import { createAssetModelQuery } from './createAssetModelQuery';
 import { getPlugin } from '@iot-app-kit/core';
@@ -32,7 +31,6 @@ import {
   AssetResource,
   useGetConfigValue,
 } from '@iot-app-kit/react-components';
-import { useAssetsForAssetModel } from './assetsForAssetModelSelect/useAssetsForAssetModel/useAssetsForAssetModel';
 import { ExpandableSection } from '@cloudscape-design/components';
 import { ExpandableSectionHeading } from '../components/expandableSectionHeading';
 import { alarmSelectionLabel } from '../../helpers/alarmSelectionLabel';
@@ -45,6 +43,7 @@ export interface AssetModelDataStreamExplorerProps {
   correctSelectionMode: 'single' | 'multi';
   timeZone?: string;
   significantDigits?: number;
+  currentSelectedAsset?: AssetSummary;
 }
 
 export const AssetModelDataStreamExplorer = ({
@@ -54,6 +53,7 @@ export const AssetModelDataStreamExplorer = ({
   correctSelectionMode,
   timeZone,
   significantDigits,
+  currentSelectedAsset,
 }: AssetModelDataStreamExplorerProps) => {
   const metricsRecorder = getPlugin('metricsRecorder');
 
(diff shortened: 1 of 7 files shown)


packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/assetModelPropertiesExplorer/assetModelPropertiesExplorer.tsx
```diff
@@ -19,6 +19,7 @@ export interface AssetExplorerProps {
   selectedWidgets: DashboardWidget[];
   timeZone?: string;
   significantDigits?: number;
+  correctSelectionMode: 'single' | 'multi';
```
corteggiano, line 22: is this a public API update or just internal?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/assetModelDataStreamExplorer.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { type IoTSiteWise } from '@aws-sdk/client-iotsitewise';
+import { AssetSummary, type IoTSiteWise } from '@aws-sdk/client-iotsitewise';
```
corteggiano, line 2: should this whole line be a type import? 
`import type { AssetSummary, IoTSiteWise } from '@aws-sdk/client-iotsitewise';`

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/assetModelDataStreamExplorer.tsx
```diff
@@ -111,17 +102,8 @@ export const AssetModelDataStreamExplorer = ({
...
-  useEffect(() => {
-    if (currentSelectedAsset)
-      selectAsset([
-        {
-          ...currentSelectedAsset,
-          assetId: currentSelectedAsset?.id,
-        } as AssetResource,
-      ]);
-  }, [currentSelectedAsset, selectAsset]);
+    selectAssetModelProperties([]); // if selectedAsset changes, dont keep state from old selection in RE
```
corteggiano, line 105: should there be any checks around resetting the selected state?
is there any chance that something in the dependency array could change on accident?

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/getAssetModelQueryInformation.ts
```diff
@@ -8,9 +8,14 @@ export const getAssetModelQueryInformation = (
...
   const assetId = firstAssetModel?.assetIds?.at(0);
-  const propertyIds = firstAssetModel?.properties.map(
-    ({ propertyId }) => propertyId
-  );
+
+  const propertyIds: string[] | undefined = firstAssetModel?.properties
+    .map(({ propertyId }) => propertyId)
+    .filter((propertyId, index) => {
+      // dedupe any new property ids
+      if (propertyIds) return propertyIds.indexOf(propertyId) === index;
```
corteggiano, line 16: you can dedupe by creating a Set and then spreading that into an array:
which might be cleaner than this filter function

```
const propertyIds = firstAssetModel?.properties.map(
  ({ propertyId }) => propertyId
);

const uniquePropertyIds = [...new Set(propertyIds)]
```

packages/dashboard/src/components/queryEditor/queryEditor.tsx
```diff
@@ -18,6 +20,16 @@ export function QueryEditor({
...
   const addButtonDisabled = useIsAddButtonDisabled(selectedWidgets);
   const correctSelectionMode = getCorrectSelectionMode(selectedWidgets);
+  const { assetModelId, assetIds } = useModelBasedQuery();
+  const { assetSummaries } = useAssetsForAssetModel({
+    assetModelId,
+    iotSiteWiseClient,
+    fetchAll: true,
+  });
+
+  const currentSelectedAsset = assetSummaries.find(
```
corteggiano, line 30: will assetSummaries ever be undefined?

ssjagad: if the api errors out, or there are no models :)

packages/dashboard/src/components/queryEditor/iotSiteWiseQueryEditor/assetModelDataStreamExplorer/assetModelDataStreamExplorer.tsx
ssjagad: I dont think so. currentSelectedAsset and selectAssetModelProperties are constant functions, so the only thing that would result in re-render would be the selectedAsset itself. When a asset does change, we do want to clear whatever properties were selected from the previous asset, otherwise when the user clicks add it would add properties that don't appear selected on the screen.

packages/dashboard/src/components/queryEditor/queryEditor.tsx
ssjagad: no, but in case there is no assets retrieved, currentSelectedAsset will be undefined which has handling in the query editors.

ssjagad: unless the query errors out

corteggiano (approved):

PR #3019, opened 2024-10-01
feat(alarms): show alarms in properties tab

corteggiano:
### Deleting alarm properties
https://github.com/user-attachments/assets/5b32ff34-dab7-4ad3-b0c6-004505b633f7

### Custom and demo alarm property (bar chart)
Screenshot 2024-10-01 at 10 01 07

### Custom alarm property (KPI)
Screenshot 2024-10-01 at 10 01 01

### Demo alarm property (KPI)
Screenshot 2024-10-01 at 10 00 55

## Legal


packages/dashboard/src/util/transformAlarms.ts
```diff
@@ -0,0 +1,29 @@
...
+export const mapAlarmQueriesToRequests = (alarmQueries: AlarmQuery[]) =>
+  alarmQueries.flatMap((query) => convertAlarmQueryToAlarmRequest(query));
+
+export const transformAlarmForPropertyPanel = (alarm: AlarmData) => ({
+  propertyId: alarm.compositeModelId,
+  name: alarm.compositeModelName,
+  unit: alarm.inputProperty?.at(0)?.property.unit,
+  dataType: alarm.inputProperty?.at(0)?.property.dataType,
+  alias: undefined,
+  assetName: alarm.inputProperty?.at(0)?.property.name,
```
hwandersman, line 28: Isn't this the name of the property not the asset?

packages/dashboard/src/util/transformAlarms.ts
```diff
@@ -0,0 +1,29 @@
...
+      return results;
+    }, [] as AlarmCompositeModelRequest[]) ?? []
+  );
+};
+
+export const mapAlarmQueriesToRequests = (alarmQueries: AlarmQuery[]) =>
+  alarmQueries.flatMap((query) => convertAlarmQueryToAlarmRequest(query));
+
+export const transformAlarmForPropertyPanel = (alarm: AlarmData) => ({
+  propertyId: alarm.compositeModelId,
```
hwandersman, line 23: This isn't exactly the propertyId, how is `propertyId` used to display the property?

packages/react-components/src/index.ts
```diff
@@ -32,6 +32,9 @@ export { useTimeSeriesData } from './hooks/useTimeSeriesData';
 export { useHasFeatureFlag } from './hooks/useHasFeatureFlag';
 export { useGetConfigValue } from './store/index';
 
+export { useAlarms } from './hooks/useAlarms';
```
hwandersman, line 35: Is there any way we can avoid exporting useAlarms? Can we just use the alarm query directly and display the assetId / compositeModelId for now?

corteggiano:
putting back into draft until i can fix the implementation to not rely on useAlarms

packages/dashboard/src/util/transformAlarms.ts
corteggiano: this is more of an internal naming issue, this array of transformed alarms is a fancy way to figure out what text we want to show users!

corteggiano: we basically search through this array of transformed alarms to find one with an ID which matches the "property" we're rendering

packages/react-components/src/index.ts
corteggiano: yes! will update as a followup

packages/dashboard/src/util/transformAlarms.ts
corteggiano: bigger issue with this component is that it doesnt work with our definition of alarms so im working around its definition

currently an asset property renders its name as `property_name (asset_name)` where asset name is taken from line 28 above, so im using this same format to render alarms as `alarm_name (alarm_property_name)`!

hwandersman: Yeah that makes sense! Let's make that clearer either with comments or variable names, since that's not obvious here.

packages/dashboard/src/customization/propertiesSections/propertiesAndAlarmsSettings/styledSection.tsx
```diff
@@ -236,7 +237,60 @@ export const StyledPropertiesAlarmsSection: FC<
         })
       ) ?? [];
 
-    const components = [...modeled, ...unmodeled, ...assetModeled];
+    const alarms =
```
hwandersman, line 240: nit: there's probably a way for us to reuse the code between the section and styledSection files, but for now this is fine

hwandersman (approved):

PR #3029, opened 2024-10-03
feat(alarms): update status timeline alarm implementation

corteggiano:
## Overview
Screenshot 2024-10-03 at 09 11 59

### Adding alarms
https://github.com/user-attachments/assets/94e48035-e305-43a8-a5ae-a5115499c0f3

### Deleting alarms
https://github.com/user-attachments/assets/8ceb2c2e-f71a-43a2-bf52-f385220956d5

### Coexisting with other data streams
Screenshot 2024-10-03 at 12 20 39

## Legal

### DIFF
--- a/packages/react-components/src/components/status-timeline/alarmTransforms.ts
+++ b/packages/react-components/src/components/status-timeline/alarmTransforms.ts
@@ -1,0 +1,74 @@
+import {
+  COMPARISON_OPERATOR,
+  type DataPoint,
+  type DataStream,
+} from '@iot-app-kit/core';
+import type { AlarmData } from '../../hooks/useAlarms';
+import {
+  parseAlarmStateAssetProperty,
+  PascalCaseStateName,
+} from '../../hooks/useAlarms/transformers';
+import { CHART_ALARM_ERROR } from '../../common/constants';
+
+export const transformAlarmStateToDataStream = (
+  alarm: AlarmData
+): DataStream<PascalCaseStateName> => {
+  const data = (
+    alarm.state?.data?.map((d) => {
+      const parsedData = parseAlarmStateAssetProperty(d);
+      if (parsedData) {
+        return {
+          x: parsedData?.timestamp,
+          y: parsedData?.value.state,
+          quality: parsedData?.quality,
+        };
+      }
+    }) ?? []
+  ).filter((point) => !!point) as DataPoint<PascalCaseStateName>[];
+
+  const firstInputProperty = alarm.inputProperty?.at(0)?.property;
+
+  return {
+    id: `${alarm.assetId}---${alarm.compositeModelId}---state`,
+    data: data,
+    name: alarm.compositeModelName,
+    isLoading: alarm.status.isLoading,
+    isRefreshing: alarm.status.isRefetching,
+    assetName: firstInputProperty?.name,
+    error: alarm.status.isError ? CHART_ALARM_ERROR : undefined,
+    resolution: 0,
+  };
+};
+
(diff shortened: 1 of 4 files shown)


packages/react-components/src/hooks/useAlarms/hookHelpers/useAlarmState.ts
```diff
@@ -82,7 +82,9 @@ export const useAlarmState = ({
    * we fetch the last 10s of alarm state every 5s.
    */
   const latestQueriesViewport =
-    refreshRate !== undefined ? { duration: refreshRate * 2 } : undefined;
+    refreshRate !== undefined && refreshRate !== Infinity
```
hwandersman, line 85: Good catch!

packages/react-components/src/hooks/useAlarms/useAlarms.ts
```diff
@@ -87,6 +88,7 @@ function useAlarms<T>(options?: UseAlarmsOptions<T>): (T | AlarmData)[] {
     iotSiteWiseClient,
     alarmDataList: typePropertyAlarmData,
     alarmPropertyFieldName: 'source',
+    refreshRate: settings?.refreshRate,
```
hwandersman, line 91: We purposefully didn't set the refresh rate for fetching the alarm type and source because we only need to do that once. We should be setting the refreshRate here to `Infinity` for that reason.

jmbuss: oh, then that was a different type of bug then! because it was defaulting to 5s!

packages/react-components/src/components/status-timeline/alarmTransforms.ts
```diff
@@ -0,0 +1,72 @@
...
+    }) ?? []
+  ).filter((point) => !!point) as DataPoint<PascalCaseStateName>[];
+
+  const firstInputProperty = alarm.inputProperty?.at(0)?.property;
+
+  return {
+    id: `${alarm.assetId}---${alarm.compositeModelId}---state`,
+    data: data,
+    name: alarm.compositeModelName,
+    isLoading: alarm.status.isLoading,
```
hwandersman, line 34: How about `isError`? And surfacing the custom error that alarms failed to load? Not sure how this is handled internally in StatusTimelineBase.

packages/react-components/src/hooks/useAlarms/useAlarms.ts
jmbuss: Changed this to set refreshRate explicitly to Infinity in `useLatestAlarmPropertyValues`

packages/react-components/src/hooks/useAlarms/hookHelpers/useLatestAlarmPropertyValues.ts
```diff
@@ -47,6 +47,7 @@ export function useLatestAlarmPropertyValues({
     iotSiteWiseClient,
     requests: alarmPropertyRequests,
     enabled,
+    refreshRate: Infinity,
```
hwandersman, line 50: This is fine for now, but I would have kept the setting for refreshRate passed through, and then in useAlarms set the refreshRate to Infinity. That way this hook doesn't hardcode fetching the latest for some property just once.

packages/react-components/src/components/status-timeline/alarmTransforms.ts
corteggiano: nice catch! updated :)

hwandersman (approved):

PR #3041, opened 2024-10-16
fix(draco): modifying check to ensure globalSettings are ready for ev…

mumanity:
…aluation

## Overview
Due to how `globalSettings` are setup, we need to add a check for `getSceneObjectFuncion` to the transcoder logic to ensure that they are not executed prior to to receiving the config overrides. 
- includes fix for test

### Scene Composer

- Download the zip file

```
npm install -g httpserver
cd <Extracted Zip Directory>
httpserver
```

## Legal


packages/scene-composer/src/components/three-fiber/ModelRefComponent/__tests__/GLTFLoader.spec.ts
```diff
@@ -94,8 +94,7 @@ describe('GLTFLoader', () => {
     it('should execute without draco decoder enabled', async () => {
       const getGlobalSettingsMock = getGlobalSettings as jest.Mock;
       const dracoDecoder: DracoDecoderConfig = {
-        enable: true,
```
mumanity, line 97: The logic on this test was backwards & should have been testing with `dracoDecoder` disabled.

corteggiano (approved):

PR #3042, opened 2024-10-20
fix(dashboard): remove gap at bottom of chart in juicebox

ssjagad:
## Overview
We were seeing this gap at the bottom of the line chart in juicebox. image

Issue is due to react-adding wrapper div around base-chart element in juicebox. 
image

This is automatically done by React when there is no react fragment wrapping around passed in divs (which is occurring in the table component due to the chart + legend).  https://react.dev/reference/react/Fragment 

## Verifying Changes

This fix was tarballed and tested that it removed the gap in Juicebox.
image

## Legal


ssjagad:
Better way to do this. Closing branch.

PR #3057, opened 2024-11-21
chore: update stale packages

tracy-french:
## Overview
All over the code base, the wrong versions of dependencies are being used. This change fixes this issue, allowing the dependency versions listed in package.json files to be used correctly.

The issue was likely caused by improper installation of packages using npm workspaces. I plan on talking with the team how to install packages.

In the process of fixing this issue, which was primarily fixed by regenerating package-lock.json, the updates up many dependencies had to be handled, most notably storybook. There was a version mismatch between doc-site (storybook v7) and other packages (storybook v6) and only the lock file was holding the issue at bay. Once unlocked, transitive dependencies experienced versioning conflicts.

In handling the storybook issue, I figured we might as well upgrade all the way to storybook 8 with vite. Hope you enjoy the dev experience!

Overall, as a result of this change, we should see:
- Less bugs related to version issues, especially differing between machines.
- Significantly faster installs (2-4 times faster)
- ~80% of vulnerable packages updated (90 -> 16 on my machine)
- Faster development env using vite instead of webpack with storybook
- ~1-2 minutes removed from GH action playwright tests with vite storybook startup

## Legal


packages/scene-composer/package.json
```diff
@@ -64,11 +64,11 @@
     "_test:ui:local": "npm run playwright -- npx playwright test",
     "_test:ui:reliability:ci": "npx playwright test --repeat-each 5 --workers=1",
     "_test:ui:reliability:local": "npm run playwright -- npx playwright test --repeat-each 5 --workers=1",
-    "test:ui:update": "npm run playwright -- npx playwright test --update-snapshots",
-    "test:ui:dev": "npx playwright test --ui",
-    "storybook": "npm run copy-assets:es && start-storybook -h 0.0.0.0 -p 7006",
+    "test:ui:update2": "npm run playwright -- npx playwright test --update-snapshots",
```
hwandersman, line 67: nit: did you mean to change the command name here?

hwandersman:
Anything to add to the package READMEs about new commands, installation, storybook setup, env vars?

hwandersman (approved):

PR #3092, opened 2024-12-09
fix: packages

tracy-french:
## Overview
This change improves iot-app-kit with the following changes:
- Enable development of "protected" packages which public packages may consume, yet aren't released as a public package. The protected package is "public" as it will be in a consumer's node_modules, yet it will not be published to npm. Only iot-app-kit packages may consume public packages. It's like OOP, I think! Plenty of in code docs added.
- Live development across packages without builds. Run `npm start` to open the dashboard storybook env and that's it. Run the unit tests, lint, whatever, without building. Start watch mode on vitest for one package, change a dependency of the package, and watch the unit test runner restart. See types changes instantly between files in other packages before you even press save. Is it magic? Yes! The incantations are filled with marginalia.
- Migration from tsc to Vite for building packages.
- Improved clean and pack scripts/utils.

Also included:
- Fix for scene-composer's internationalization. IDs are now correctly injected into the dist.

## Legal


chejimmy (approved):
LGTM! This is a great utility to organize AppKit repo packages and TY for fixing some bugs 🍾 
Can you add a lil guide/README on how to make an "protected" package? It will help the next person to utilize it 🙏

PR #3100, opened 2025-01-09
feat: add support for null / nan

jmbuss:
## Overview
Update for null / nan support. The main change is to the utility function `toValue`. This function is responsible for mapping the value of the `Variant` type object. 

The new type will look like:
```
interface Variant {
    /**
     * <p>Asset property data of type string (sequence of characters).</p>
     * @public
     */
    stringValue?: string | undefined;
    /**
     * <p>Asset property data of type integer (whole number).</p>
     * @public
     */
    integerValue?: number | undefined;
    /**
     * <p>Asset property data of type double (floating point number).</p>
     * @public
     */
    doubleValue?: number | undefined;
    /**
     * <p>Asset property data of type Boolean (true or false).</p>
     * @public
     */
    booleanValue?: boolean | undefined;
    nullValue?: PropertyValueNullValue | undefined;
}
```

This change also moves the `toValue` utility function to the protected helpers package so that it can be used in source-iotsitewise and react-components since the resource-explorer does not use the same datasource as the widgets.

## Legal

### DIFF
--- a/packages/core-util/package.json
+++ b/packages/core-util/package.json
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@iot-app-kit/core": "*",
+    "@iot-app-kit/helpers": "*",
     "lodash-es": "^4.17.21",
     "pako": "^2.1.0",
     "parse-duration": "^1.0.3",
--- a/packages/data-mocked/src/iot-sitewise/handlers/batchGetAssetPropertyValue/batchGetAssetPropertyValueHandler.ts
+++ b/packages/data-mocked/src/iot-sitewise/handlers/batchGetAssetPropertyValue/batchGetAssetPropertyValueHandler.ts
@@ -67,8 +67,9 @@ export class AssetPropertyValueFactory {
       ...this.#createDefaults(),
       ...partialAssetPropertyValue,
       value: {
+        // nullValue: { valueType: 'I' },
         integerValue: faker.number.int({ min: 0, max: 100 }),
-      },
+      } as any,
     };
 
     return assetPropertyValue;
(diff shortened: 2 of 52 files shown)


packages/data-mocked/src/iot-sitewise/handlers/batchGetAssetPropertyValue/batchGetAssetPropertyValueHandler.ts
```diff
@@ -67,8 +67,9 @@ export class AssetPropertyValueFactory {
       ...this.#createDefaults(),
       ...partialAssetPropertyValue,
       value: {
+        // nullValue: { valueType: 'I' },
```
corteggiano, line 70: commented out on purpose? can it be deleted?

packages/data-mocked/src/iot-sitewise/handlers/batchGetAssetPropertyValue/batchGetAssetPropertyValueHandler.ts
```diff
@@ -81,8 +82,9 @@ export class AssetPropertyValueFactory {
       ...this.#createDefaults(),
       ...partialAssetPropertyValue,
       value: {
+        // nullValue: { valueType: 'D' },
```
corteggiano, line 85: commented out on purpose? can it be deleted?

packages/core-util/package.json
```diff
@@ -51,6 +51,7 @@
   "devDependencies": {
     "@aws-sdk/types": "^3.696.0",
     "@iot-app-kit/eslint-config": "*",
+    "@iot-app-kit/helpers": "*",
```
corteggiano, line 54: should this be a devDependency? do we need it just for local dev or for runtime as well?

corteggiano (approved):

packages/data-mocked/src/iot-sitewise/handlers/batchGetAssetPropertyValue/batchGetAssetPropertyValueHandler.ts
jmbuss: I was leaving these in for future testing purposes.

packages/core-util/package.json
jmbuss: Updated!

PR #3112, opened 2025-01-21
feat(dashboard): remove query client provider

jmbuss:
## Overview
Remove usage of the query client provider from the Dashboard components in favor of using the client singleton directly.

## Legal


chejimmy (approved):
Code LGTM! Did you consider if there are any potential impact? like breaking some use cases?

jmbuss:
> Code LGTM! Did you consider if there are any potential impact? like breaking some use cases?

Yes, if there was any issue, the react-queries in the dashboard would throw an error because they would have no client to make requests with. But with no issues, it should appear as if nothing has changed!

